Issue #5, opened 2017-03-14
Add a `--shallow` option to `list-machines`


ncoghlan:
Actually scanning a machine image for internal details can take a while, but basic info can be retrieved from the virt provider far more rapidly.

A `shallow` scan option means a CLI client can choose between the default detailed-but-slow scan and a shallow-but-fast scan that leaves out a lot of the internal details.

PR #6, opened 2017-03-14
Issue #5: add `--shallow` option to `list-machines`

ncoghlan:
Also switches the Cockpit plugin over to using the new option,
so the list refreshes in ~3 seconds rather than ~10.


shaded-enmity:
LGTM :+1:

PR #11, opened 2017-03-28
Remove vagrant ssh hack

ncoghlan:


shaded-enmity:
What do you think about generating a key pair for testing and checking that in the repo? I'm a bit worried about having to juggle with local users keys (especially in port-forwarded scenarios).

ncoghlan:
Yeah, I think that will be better, especially as it avoids a few other problems:

- user may not have an RSA keypair
- user may have a passphrase on their RSA key
- user may not have started ssh-agent
- giving sudo processes access to the user's ssh-agent

If we add a top level "test_config" directory, put a pregenerated key there, and configure the Ansible roles for both the demo and integration test VMs to load it, we eliminate a whole host of potential failure modes.

shaded-enmity:
LGTM, I'm going to create an issue about the `test_config` TLD and using a "skeleton key" in all our Ansible playbooks.

shaded-enmity:
Aha, sloppy me, didn't notice the linked issue which already mentions the skeleton key, never mind :)

PR #15, opened 2017-03-29
WIP: add playbooks for test machine provisioning

shaded-enmity:
WIP:

- [x] Install behave & additional tools needed for running the tests
- [ ] Refactor SCL installation into something more sensible
- [x] Use /etc/sudoers.d/ rathern than writing directly to /etc/sudoers
- [ ] Fix typos, add comments
- [ ] Update project directory structure as per (https://github.com/leapp-to/prototype/pull/15#issuecomment-290283501)


ansible/centos-ci/ansible/roles/vagrant/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
+# file: ansible/roles/vagrant/install.yml
+---
+
+# For some reason this task gets always skipped :/
+#- name: "Install Vagrant from SCL"
+#  include_role:
+#    name: smbambling.scl
```
ncoghlan: Is `smbambling.scl` actually on the `roles_path` when it gets skipped? (I'm currently looking at http://docs.ansible.com/ansible/playbooks_roles.html#roles as I consider how we might be able to address https://github.com/leapp-to/prototype/issues/14 )

ansible/centos-ci/ansible/roles/user/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+
+# Create a `ci` account with the following properties
+# - vagrant and libvirt group membership
+# - known ssh key in authorized_keys
+#
+# Also configure the `root` user to have a well known authorized_keys key
+
+- name: Create `ci` account
+  user:
+    name: ci
```
ncoghlan: Perhaps this should be namespaced as `leapp-ci` rather than using a generic account name?

ncoghlan:
The top-level `ansible` directory is irritating me a bit, as the fact leapp uses ansible for VM configuration and (eventually) remote access is mostly an implementation detail rather than being a meaningful high level category in the project structure.

What do you think about instead having the following arrangement:

- put the `centos-ci` directory for this PR directly at the top level
- rename the current `ansible` directory to `demo/vmdefs` (to align with `integration-tests/vmdefs`)
- move the `start_vms.sh` and `start_integrations_vms.sh` scripts into the `demo` and `integration-tests` directories (renaming the latter script to just `start_vms.sh` in the process)
- add a top level `ansible_roles` directory for https://github.com/leapp-to/prototype/issues/14 that just has the shared role definitions, and nothing else

(The `cockpit` and `nginx` directories could probably also move under the `demo` directory at that point)

ncoghlan:
Although I'm less sure about moving the `cockpit` directory, as I'd actually like to start trying to run that in the integration-tests at some point, so we know we're not breaking it when we make CLI changes in `leapp-tool.py` (or in the DBUS API once we get that far).

ansible/centos-ci/ansible/roles/vagrant/tasks/main.yml
shaded-enmity: Yup, installed it via Ansible galaxy, otherwise Ansible complains that the given role is missing and bails. (I could actually see the the first step of that role being executed, which checks for the presence of SCL GPG key).

ansible/centos-ci/ansible/roles/user/tasks/main.yml
shaded-enmity: :+1:

shaded-enmity:
Yup, agreed, added a work item for that to the OP. This may take more time, I could really use some Ansible crash course with something like "intent based matrix" mapping my intents to appropriate Ansible primitive. (I'll probably start preparing one)

ncoghlan:
From what I've been able to figure out:

- task: a specific command to run on a target system (typically a combination of a system status check and instructions on how to transition into the desired state if not already there)
- role: a sequence of tasks to ensure a particular capability is available on a given system
- playbook: mapping a set of roles to a set of machines in some combination or other

shaded-enmity:
I'm hitting a problem with `pipenv` now, it hangs while processing the lock file:

Symptoms:
```bash
$ pipenv install 
Fixing package names in Pipfile...
No package provided, installing all dependencies.
Fixing package names in Pipfile...
Pipfile found at /srv/le-app/Pipfile. Considering this to be the project home.
Installing dependencies from Pipfile.lock...
```

The process hangs at this point (I let it run for 20 minutes). Poking around the system I was able to figure out the sub-command issued by `pipenv`:

```bash
/root/.local/share/virtualenvs/le-app-xOFUapXL/bin/pip install  -r /tmp/tmpecommvcy-requirements.txt -i https://pypi.python.org/simple
```

If I execute this command manually, it complains about being unable to build certain wheels, but the command succeeds. Successive execution of `pipenv install` returns immediately.

CPU usage of both processes was minimal:
```bash
root     20069  0.4  2.2 309120 23044 pts/0    Sl+  14:32   0:00  |                       \_ /root/.local/venvs/pipenv/bin/python3 /root/.local/bin/pipenv install --ignore-hashes
root     20076  0.6  3.2 253584 32636 pts/0    S+   14:32   0:00  |                           \_ /root/.local/share/virtualenvs/le-app-xOFUapXL/bin/python3 /root/.local/share/virtualenvs/le-app-xOFUapXL/bin/pip install -r /tmp/tmpuamy4fy2-requirements.txt -i https://pypi.python.org/simple
```

Strace suggests a deadlock scenario:
```bash
[root@centos7-ci vagrant]# strace -p 20069
Process 20069 attached
wait4(20076, 

---

[root@centos7-ci vagrant]# strace -p 20076 -s 1024
Process 20076 attached
write(1, "    Skipping link https://pypi.python.org/packages/bd/bc/d754e01270adc77eb0c645078f90a654030ccd87d7433d62298ec2bf5632/pyasn1-0.1.5-py2.4.egg#md5=9801921c021171460260886117abcd30 (from https://pypi.python.org/simple/pyasn1/); unsupported archive format: .egg\n", 258
```

Since tracing the system after it has been deadlocked like this is mostly useless, I decided to strace the whole execution of `pipenv install --ignore-hashes` like this:
```bash
[root@centos7-ci le-app]# strace -v -s 1024 -ff -o pipenv -tt -y -yy pipenv install --ignore-hashes 
```

Which produced an incredible number os sub-processes:
```bash
[root@centos7-ci le-app]# ls -l pipenv.* | wc -l
984
```

What's worse, after ~20 minutes of runtime, the traces sprawled into majestic 110MB of information:
```bash
[root@centos7-ci le-app]# SUM=0 ; for file in pipenv.*; do SUM=$(($SUM + $(stat -c '%s' $file))) ; done ; echo $(($SUM / (1024 * 1024)))MB
111MB
```

To reproduce, checkout this branch and run `vagrant up` in the `centos-ci` folder, wait till it gets stuck `CTRL+C` and `vagrant ssh` to get a shell into the partially provisioned VM.

ncoghlan:
@shaded-enmity That sounds like `pipenv` and/or `pew` (the underlying environment manager) may not be reading the subprocess output as it gets produced, and hence letting the OS level pipe buffers fill up and block.

ncoghlan:
If I'm right, this may be a possible workaround:

* run `pipenv lock --requirements > requirements.txt`
* check in the resulting `requirements.txt`
* use `pip` and `virtualenv` directly in the CI, rather than `pipenv shell`

The downside being that the CI wouldn't be testing the actual workflow we suggest to developers

shaded-enmity:
@ncoghlan yup, it looks like the pipe problem, I've tried fooling around with `vendor/delegator.py` in `pipenv` and it looks like either stdout/stderr is required as `pipenv` actually checks the output. 

(On a tangential note, yesterday evening I was experimenting with the `pty` module to create a buffer-less version of `subprocess.run`, and my experiment caused a lockdown/DoS of the console subsystem in Bash on Windows)

shaded-enmity:
I'll also probably need to flesh out some parts of the Vagrant role into `bootstrap.sh` script:
1) CI gives us clean machine with a clone of this repository
2) Before running the playbook we need to install Ansible first
3) To install Ansible we need to add EPEL

Steps 2+3 are currently a part of the `vagrant` role, so I'll have to move that to the script.
Eventually the workflow should look like:

1) CI gives us clean machine with a clone of this repository
2) Run `bootstrap.sh` which will install EPEL & Ansible
3) Run the `centos-ci` playbook to provision our test env
4) Run all the tests

ncoghlan:
Note that Ansible is pip-installable - it's the sole non-development dependency currently in `Pipfile`.

We may still decide that enabling EPEL on the CI server is a good idea, it just isn't necessary for the specific task of installing Ansible.

shaded-enmity:
@ncoghlan good idea, removed the EPEL code and added simple `pip install ansible` into the bootstrap script. I've also solved the pipenv issues by ad-hoc patching 2 pipenv files during provisioning.

shaded-enmity:
The last commits changes the directory structure to the one proposed in https://github.com/leapp-to/prototype/pull/15#issuecomment-290283501 - I'll dedup the task on Monday, wrote a script to do the hard part for me already: 
https://github.com/leapp-to/prototype/blob/6eaaf098e30a0c8f333529fa943ca44821ed5f78/utils/ansible_tasks_deduplicate.py

ncoghlan:
@shaded-enmity It's probably worth moving the directory layout changes out to a separate PR and merging that first, as the way GitHub renders the directory movement diffs leaves a lot to be desired (separate entry for each file below the renamed directory)

shaded-enmity:
@ncoghlan yeah good idea, currently deduping the Ansible tasks. Will open a new PR then update this one.

PR #19, opened 2017-03-30
Partial fix for demo execution

ncoghlan:
- load SSH key into demo VMs correctly
- improve the plugin output when running the migration command


ncoghlan:
Merging to give a better starting point for the rest of #20

Issue #20, opened 2017-03-30
Problems with running Cockpit demo


ncoghlan:
I tried running the Cockpit demo for the first time since adding the integration tests, and found a couple of problems:

- the known SSH key is being loaded from the wrong path in the Ansible roles
- the migration script is assuming the migrated HTTP service in on port 80 (to match the integration tests), when the demo still expects it to be on port 9000

The former is easy to fix and I've submitted a PR for that in https://github.com/leapp-to/prototype/pull/19

For the latter, we need to decide *how* we want to fix it for the time being.

My suggestion is that, for now, we just add a `--tcp-port` option to `migrate-machine` that can be used to override the default `-p 80:80` option used by the integration tests (we'd leave the `-p 9022:22` for SSH forwarding to the macrocontainer hardcoded).

The demo plugin would then be updated to add `--tcp-port 9000:80` to configure the macrocontainer host correctly.

shaded-enmity:
Sounds good, we could perhaps support multiple instances of --tcp-port to open multiple ports, but I'm not sure if we have short-term use case for it, and from the long-term perspective the flag will probably go away at some time in favor of some config file.

ncoghlan:
#21 has been merged, so this is working again.

Issue #18, opened 2017-03-30
Eliminate name conflicts between demo and integration-testing VMs


ncoghlan:
Some names are currently re-used between the demo VMs and the integration-testing VMs, which can make things rather slow when switching back and forth between them.

One way to fix that:

- keep the existing directory names for the integration testng VM definitions
- add a `leapp-testing-` prefix to the hostnames in the Vagrant files
- adjust `_get_ip_addresses` in the integration test step definitions to expect the prefix when looking for the test VMs in the `list-machines --shallow` output

ncoghlan:
I needed this while working on https://github.com/leapp-to/prototype/issues/20 so the PR for that will also address this

ncoghlan:
The other step needed is to use `config.vm.define` to override the `default` section of the actual machine name in libvirt

ncoghlan:
#21 has been merged, so the integration test VMs no longer conflict with the demo VMs.

PR #21, opened 2017-03-31
Fix errors in demo

ncoghlan:
Fixes issue #20 by updating the Cockpit plugin to explicitly forward the PHP MyAdmin port rather than the default HTTP port.

Also fixes issues #18 by using `leapp-tests` as a disambiguator for the integration testing VM and host names.

### DIFF
--- a/integration-tests/features/environment.py
+++ b/integration-tests/features/environment.py
@@ -1,4 +1,11 @@
 import contextlib
+import subprocess
+
+def before_all(context):
+    # Some steps require sudo, so for convenience in interactive use,
+    # we ensure we prompt for elevated permissions immediately,
+    # rather than potentially halting midway through a test
+    subprocess.check_output(["sudo", "echo", "Elevated permissions needed"])
 
 def before_scenario(context, scenario):
     context.resource_manager = contextlib.ExitStack()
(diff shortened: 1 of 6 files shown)


integration-tests/features/environment.py
```diff
@@ -1,4 +1,11 @@
 import contextlib
+import subprocess
+
+def before_all(context):
+    # Some steps require sudo, so for convenience in interactive use,
+    # we ensure we prompt for elevated permissions immediately,
+    # rather than potentially halting midway through a test
+    subprocess.check_output(["sudo", "echo", "Elevated permissions needed"])
```
shaded-enmity: Wanted to suggest to `assert os.getuid() == 0`, but that wouldn't actually check the presence of the `sudo` binary.
Might be worthwhile to "centralize" the usage of sudo, and skip it completely if we're already uid == 0.

shaded-enmity:
Other than the minor nit regarding `sudo` usage LGTM.
I've filled #22 to consolidate the usage of sudo as a future task, so I'm ready to merge this one.

integration-tests/features/environment.py
ncoghlan: Alternatively, if we go the DBus option, we may be able to get away from requiring sudo on the client in the first place: https://github.com/leapp-to/prototype/issues/7

Issue #2, opened 2017-03-06
Error attempting to migrate from CentOS 7 VM to CentOS container host


ncoghlan:
Attempting to follow the instructions from the PR in #1, I get the following error from `migrate-machine`:

```
$ sudo ./leapp-tool.py migrate-machine -t rhel7-target rhel6-app-vm
[sudo] password for ncoghlan: 
! looking up "rhel6-app-vm" as source and "rhel7-target" as target
! obtaining SSH keys
Traceback (most recent call last):
  File "./leapp-tool.py", line 131, in <module>
    ssh = _get_ssh_config()
  File "./leapp-tool.py", line 17, in _get_ssh_config
    for x in check_output(['vagrant', 'ssh-config'], cwd=path).decode('utf-8').splitlines()[1:-1]]
  File "/usr/lib64/python2.7/subprocess.py", line 212, in check_output
    process = Popen(stdout=PIPE, *popenargs, **kwargs)
  File "/usr/lib64/python2.7/subprocess.py", line 390, in __init__
    errread, errwrite)
  File "/usr/lib64/python2.7/subprocess.py", line 1024, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory: '/root/leapp-to/rhel6-guest-lamp'
```

I'm guessing I'm missing a setup step somewhere.

shaded-enmity:
Ah yes, bad paths (cucos-02 ones), should be fixed now. Thanks

shaded-enmity:
Also the relative path used in the script assumes that $PWD == the repo checkout. This could be handled better I suppose, but it's a side effect from using Vagrant, and it keeps the CLI usage from becoming too verbose.

shaded-enmity:
Updated `cockpit.spawn` to reflect that: 104e751

PR #25, opened 2017-04-04
Add new SCL shared role and CI machine playbook

shaded-enmity:


ncoghlan:
LGTM!

centos-ci/bootstrap.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+if [[ $(id -u) != 0 ]]; then
+   echo "$0 needs to be run as root" ; exit 1
+fi
+
+pip install ansible==2.2.0
```
ncoghlan: Hmm, I may have been too hasty on the merge here: while this line will succeed on Fedora, isn't it going to fail on CentOS 7, since there's no "pip" in the default image?

We'll likely need to bootstrap `pip` itself first via `curl "https://bootstrap.pypa.io/get-pip.py" | python` as covered by Option #2 in https://www.liquidweb.com/kb/how-to-install-pip-on-centos-7/

shaded-enmity: Ahh, was testing it in Vagrant *after* provisioning, silly me 💩

shaded-enmity: I'll update the bootstrap script once I get access to the CI as there might be more work to do

PR #29, opened 2017-04-05
Add a test skeleton for a new DBus service

ncoghlan:
- starts with CLI performance tests, as improving listing response
  time will be the most noticeable immediate benefit of status caching
- Also renames `migration_helper` to `cli_helper`, as the former name
  doesn't make much sense when testing commands other than the migration
  command


ncoghlan:
@shaded-enmity Even though it doesn't actually add a DBus service yet, I think this is mergable, and it may be worth doing so given the `migration_helper` -> `cli_helper` change.

I do wonder whether it might be better to change the name of the feature file to be `list-machines.feature` though, as that's really what it's testing at the moment.

If I did that, I'd move the two new machine listing step definitions to their own `list_machines.py` file.

shaded-enmity:
@ncoghlan I think that `list-machines.feature` sounds better, as it's closer in scope to what we're doing, rather than testing DBus itself.

LGTM Otherwise :)

ncoghlan:
I went ahead and merged this, with https://github.com/leapp-to/prototype/tree/master/integration-tests#writing-new-test-scenarios now providing more info and links on how to write new test scenarios and test steps, as well as a couple of useful scenario tags.

PR #31, opened 2017-04-06
Add `target` role responsible for setting up the target env

shaded-enmity:
Currently we only need to create a `/opt/leapp-to` directory

### DIFF
--- a/ansible_roles/httpd/defaults/main.yml
+++ b/ansible_roles/httpd/defaults/main.yml
@@ -1,0 +1,2 @@
+doc_root: /var/www/html
+hostname: vagrantbox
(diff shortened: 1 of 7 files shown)


shaded-enmity:
Also adding the missing HTTPD role, probably forgot to cherry pick that commit during the refactor.

pcahyna:
works for me, thanks

ncoghlan:
Merged, but I'm wondering if we should add a catalog of the available roles to `ansible_roles/README.md` and reference that from `integration-tests/README.md` (I also guess the latter should list the known VM definitions).

Issue filed for the latter documentation idea at https://github.com/leapp-to/prototype/issues/34

Issue #33, opened 2017-04-07
Run LeApp in its own Python 2.7 environment


ncoghlan:
Following on from the decision in https://github.com/leapp-to/prototype/issues/30 to support running LeApp under Python 2.6, we need to adjust the integration tests to actually work that way.

This means that the Pipfile environment will be specifically for the integration tests, and the tests will use `pipsi` to create a dedicated Python 2.7 environment to run the LeApp migration tool.

- [x] Make a `src` directory, move leapp-tool.py and the leappto package into that, and add an appropriate setup.py file
- [x] Add a `bin` directory with a suitable `.gitignore` that gets used for the installed symlink that auto-activates the Python 2.7 environment
- [x] Add pipsi as a test dependency
- [x] Add a before_all step in the integration tests that runs `pipsi --bin-dir bin/ install --python $(which python2.7) src/`
- [x] Change the test helper to run `bin/leapp-tool` rather than `src/leapp-tool.py`
- [x] Change the demo setup instructions to cover generating `bin/leapp-tool`
- [x] Change the demo Cockpit plugin to run `bin/leapp-tool` rather than `src/leapp-tool.py`

Note that I'm setting the baseline version at 2.7, as I don't think we should support Python 2.6 for three reasons:

1. The `ssl` module in 2.6 is really old, and hence makes it unnecessarily difficult to do decent HTTPS (linking to the system OpenSSL helps, but still isn't ideal)
2. Upstream Python libraries are actively dropping Python 2.6 support (and have been for quite some time) so we'd likely also end up having to use older versions of other components like Twisted
3. SCLs mean we can fairly readily get access to Python 2.7 on RHEL/CentOS 6, but there's no similarly straightforward way of getting access to Python 2.6 on RHEL/CentOS 7

ncoghlan:
Work-in-progress PR: https://github.com/leapp-to/prototype/pull/35

ncoghlan:
As a side effect of working on this, I claimed the `leappto` name on PyPI: https://pypi.python.org/pypi/leappto

ncoghlan:
Closing, as #35 has been merged.

PR #35, opened 2017-04-07
Ensure Python 2.7 environment for tool exection

ncoghlan:
Addresses #33 by consistently using `pipsi` to install the CLI executable used by both the integration tests and the demo Cockpit plugin.


ncoghlan:
See the task list in #33 for items that need to be resolved before removing the WIP marker.

shaded-enmity:
We might need to change the default path from `/opt/leapp` to `/opt/leapp/prototype` since I have no idea how to mimic the functionality of `git clone REPO TARGET` using the JJB SCM syntax: https://docs.openstack.org/infra/jenkins-job-builder/scm.html#scm.git (perhaps `sparse-checkout`) - I can set `basedir` (https://github.com/leapp-to/prototype/pull/37/files#diff-d682dde4dbcdc306c0bfe624ddf79236R12) but then it's going to be like `/opt/leapp/protoype`.

ncoghlan:
@shaded-enmity The integration tests don't rely on `/opt/leapp` existing, only the Cockpit demo plugin. The tests do everything relative to the repo root, which they find based on `__file__` in `environment.py`.

If/when we add a preliminary integration test for the plugin, we should start out by doing the same thing we advise developers to do in the README: symlink their source checkout into the appropriate places.

bin/.gitignore
```diff
@@ -0,0 +1 @@
+leapp-tool
```
shaded-enmity: Ahh, just noticed the gitignore in the root dir.

PR #38, opened 2017-04-07
Correctly skip tagged scenarios

ncoghlan:
The autoskipping logic for scenarios tagged with `@skip` wasn't
quite right, this fixes it.


ncoghlan:
This now gives the expected behaviour for the scenario that's marked with both `@wip` and `@skip`:

    behave # Tagged scenario is skipped
    behave --tags=skip # Tagged scenario is executed
    behave --tags=wip # Tagged scenario is executed
    behave --tags=wip --tags=skip # Tagged scenario is executed

PR #39, opened 2017-04-07
Remove pipenv patches since delegator.py is now fixed

shaded-enmity:
While the PRs have already been merged we're waiting for new release in PyPI, hence the WIP flag.


shaded-enmity:
@ncoghlan Good point, will regenerate as soon as a new release hits PyPI.

shaded-enmity:
dotests

ncoghlan:
Looking at https://github.com/kennethreitz/pipenv/commits/master the version we're going to need will be either 3.5.7 or 3.6.0 (the problem was fixed shortly after 3.5.6 was released, and there hasn't been a new release since then).

Since the fix has been merged upstream, that means the CI Ansible playbook could be updated to specify `'pipenv>3.5.6'` in this PR, and it should start passing once the release has been made.

shaded-enmity:
Jenkins Test

Issue #28, opened 2017-04-05
Always reprovision test VMs even without "ensure_fresh=yes"


ncoghlan:
At the moment, test VMs only have their Ansible playbooks run when `ensure_fresh` is set to `yes` in the VM definition step.

This causes two problems:

- `ensure_fresh=no` can not only miss `Vagrantfile` changes, but also changes to the Ansible roles and playbooks
- `ensure_fresh=yes` is really slow, since it completely destroys and recreates the VM from scratch, but is currently the only way to specify that a test should re-run the Ansible playbook for a VM definition

My proposed resolution: update the tests to always pass the `--provision` flag to `vagrant up`.

That way any changes to the Ansible files will be picked up automatically on the next test run (when running the tests locally), and in CI it can be used to provide greater isolation between features by allowing the Ansible playbooks to be used to reset certain aspects of the VM to a known state

shaded-enmity:
This will also help us to make sure that all our Ansible tasks are idempotent since during re-provision all non-idempotent tasks will error.

ncoghlan:
PR: https://github.com/leapp-to/prototype/pull/40

pcahyna:
Should start_vms.sh be changed in the same way?

ncoghlan:
@pcahyna Indeed it should, and that also reminds me we were going to move those inside their respective directories (`demo` and `integration-tests`).

ncoghlan:
PR to update the helper scripts: https://github.com/leapp-to/prototype/pull/44

I found while doing that that they'd bitrotted somewhat relative to the way the integration tests worked, so it's a general update, rather than just adding the `--provision` option.

ncoghlan:
#44 is merged, so we're always reprovisioning the VMs by default now.

PR #40, opened 2017-04-10
Issue #28: always reprovision test VMs

ncoghlan:
The Ansible playbooks for the test VMs are now always run for each
VM provisioning step, providing the following benefits:

- playbook changes will automatically be picked up on the next test
  run, even when re-using previously created VMs
- playbooks can ensure VMs are reverted to a known state prior to
  use, without having to completely destroy and recreate the VM
- VMs can now readily be reprovisioned within a test scenario


shaded-enmity:
LGTM, merging.

PR #41, opened 2017-04-10
Make the IP addresses of the demo VMs configurable

pcahyna:
I have often the problem of being in a 10.0.0.0 network. I then can not run the demo, vagrant complains that: `Error while activating network: Call to virNetworkCreate failed: internal error: Network is already in use by interface enp0s31f6.`

This patch allows one to customize the addresses of the source and target VMs using environment variables SRCIPADDR and DSTIPADDR. Care was taken to keep the defaults in case the variables are unset or null.

Should be probably also done for:
- start_integration_vms.sh and vagrantfiles under integration-tests/vmdefs
- centos7-guest (@shaded-enmity, now it gets 10.0.0.11, isn't it a typo? I would suppose the source VM should get 10.0.0.10.)

### DIFF
--- a/demo/vmdefs/centos6-guest/Vagrantfile
+++ b/demo/vmdefs/centos6-guest/Vagrantfile
@@ -1,5 +1,7 @@
 VAGRANTFILE_API_VERSION = "2"
 
+ipaddr = ENV.fetch('SRC_IP_ADDR', '10.0.0.10')
+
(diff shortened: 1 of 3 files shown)


pcahyna:
Done. (N.B. I rebased the branch and squashed the changes into one commit.)

shaded-enmity:
LGTM, thanks!

PR #42, opened 2017-04-10
playbooks that we can use to redeploy le-app-01 and le-app-02

vmindru:
playbooks that we can use to redeploy le-app-01 and le-app-02


shaded-enmity:
Thanks for this @vmindru, I think that better fit for this would be the housekeeping repo, can you please open the PR there?

vmindru:
ack

PR #44, opened 2017-04-11
Improvements to VM helper scripts

ncoghlan:
- move `start_vms.sh` scripts into their respective directories
- make the scripts auto-adjust to new VM definitions
- make the scripts independent of the current working directory
- assume passwordless Vagrant access to match the integration tests


ncoghlan:
@pcahyna This PR currently partially reverts https://github.com/leapp-to/prototype/pull/41 (due to the changes to the VM startup scripts), so you'll likely want to take a look and confirm it still works for you.

We may need to change the environment variables to be a single `LEAPP_VM_SUBNET` prefix setting, rather than specifying the exact IP addresses of any of the individual machines.

ncoghlan:
On closer inspection, this should still work with the previous environment variables, as the updates in this PR *also* remove the use of `sudo` from the VM startup scripts - if you need sudo to run vagrant, you now need to wrap it around the script invocation.

That means if you have the environment variables defined and passwordless vagrant access configured, Vagrant will see them, without any custom handling in the launch scripts.

ncoghlan:
@shaded-enmity pointed out that given this change, we should remove the currently unused centos7-guest from the demo directory.

pcahyna:
Hi. I suspect that after your last merge (e0b3ab623a82ecbbcd71c484aff190619fe5adce) there are two `start_vms.sh` scripts in the repository.

Otherwise I agree with what you said: the change to `start_vms.sh` was there to work around sudo not preserving the environment, and without sudo it is no longer required. I also think `LEAPP_VM_SUBNET` is a good idea.

What was `centos7-guest` good for and why is it unneeded now?

integration-tests/README.md
```diff
@@ -150,3 +150,13 @@ useful attributes for use in step implementations:
...
+
+## Debugging the test VMs
+
+From the `integration-tests` directory, an instance of each of the integration
+test VMs can be started by running:
+
+    $ start_vms.sh
+
+This script iterates over all the subdirectories of `integration-tests/vmdefs`
+and runs `vagrant up --provision`.
```
shaded-enmity: We might want to add something along the lines "to get SSH access to test VMs change directory to it's respective folder and run `sudo vagrant ssh`.

ncoghlan:
I filed https://github.com/leapp-to/prototype/issues/45 to cover switching to the subnet env vars. For this PR, the fact the startup scripts don't use `sudo` any more turned out to be sufficient to keep the current demo-only env vars working.

Good idea to note explicitly how to ssh into a Vagrant machine rather than assuming readers are already familiar with that, so I'll make that change.

integration-tests/README.md
ncoghlan: Done!

shaded-enmity:
LGTM, merging!

PR #50, opened 2017-04-13
Initial Cockpit plugin integration tests

ncoghlan:
- [x] checks Cockpit is available on host
- [x] sets up a user with required permissions and configuration
- [x] sets the user's password appropriately
- [x] uses [splinter](https://splinter.readthedocs.io/en/latest/) to log in to http://localhost:9090 (Cockpit) and go to the named menu item
- [x] checks the expected VMs are listed there within the given time frame

Deferred

* run with a headless browser (e.g. PhantomJS) and the latest version of selenium rather than Firefox with selenium 2.x (this is a bit fiddly since it requires extra dependencies that can't be automatically installed via `pip`)


ncoghlan:
Splinter uses its Firefox driver by default, and that should be good enough for initial testing, but we're likely going to want to use something else in CI longer term.

ncoghlan:
Initial implementation of test case pushed. It's currently failing, and I'm pretty sure the problem is that the generated test user doesn't have read access to the checked out plugin - the symlink trick we have in the README assumes that the user cloning the git repo is using the same account to log into Cockpit.

I think the simplest near term fix would be to recursively copy the plugin directory over into the generated test user's home directory during the setup step, rather than symlinking it.

integration-tests/features/cockpit-demo.feature
```diff
@@ -0,0 +1,14 @@
...
+@skip
+Scenario: Initial page visit
+  Given Cockpit is installed on the testing host
+    And the demonstration user exists
+    And the demonstration plugin is installed
+    And the local virtual machines:
+         | name       | definition          | ensure_fresh |
+         | vm1        | centos6-guest-httpd | no           |
+         | vm2        | centos7-target      | no           |
+  When the demonstration user visits the Kernel Flexibility page
```
shaded-enmity: s/Kernel Flexibility/Le-App/

ncoghlan:
I figured out the permissions problem, and rather than being solely due to the symlink back into the home directory of the user running the tests, it looks like it's also due to using the `tempfile` module to create the temporary user's home directory: https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp

So I'll adjust that code to just get the base temp directory from the tempfile module, and use the randomly generated account name as the directory within that.

ncoghlan:
Jenkins: ok to test

@shaded-enmity This test scenario works now, but I've left it tagged as both `skip` and `wip` due to the current implicit Firefox dependency. Before we enable it in CI, it will need:

- the ability to run headless, rather than using Firefox
- clear documentation in the README on installing the required dependencies
- updates to the CentOS CI playbook to install those dependencies

shaded-enmity:
Glancing over LGTM, will take a closer look in the afternoon.

PR #53, opened 2017-04-13
Move the scm block to its own section

bstinsonmhk:
This should also fix the brace-escaping issue that is currently going on in CentOS CI


shaded-enmity:
@bstinsonmhk thanks! can we override the trigger phrase parameter in the future? if not, what is the magic phrase we can use?

PR #62, opened 2017-04-19
Add TCP port details to the demo instructions.

pcahyna:
- without a "--tcp-port 9000:9000" argument to leapp-tool the target VM will not show the PHP admin page
- document that the PHP admin page is on port 9000, this is not obvious.


shaded-enmity:
LGTM thanks!

PR #64, opened 2017-04-20
Make ui user friendly

mfranczy:
This PR contains:
- Simplify the information displayed in Cockpit
- Remove the "reprovision" button
- Centralized migration button

### DIFF
--- a/cockpit/leapp.html
+++ b/cockpit/leapp.html
@@ -16,31 +16,29 @@
                 </div>
                 <div class="spinner spinner-lg" id="loader"></div>
             </div>
-            <div class="source-machines" style="float: left; width: 49%;">
+            <div class="source-machines" style="float: left; min-width: 45%;">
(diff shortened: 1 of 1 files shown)


mfranczy:
How UI looks now:
screenshot from 2017-04-20 14-11-03

@shaded-enmity gave me the idea to keep hostname and ip in both tables instead of "Each source/target machine in the list contains just simple ID the admin will understand - either a hostname (if available) or IP". I agree it's better and readable for admin.

shaded-enmity:
Looks amazing @mfranczy ! Would it be possible to vertically align the `>>` symbol with the table headers, I think it'll look better.

mfranczy:
yes, you're right, it looks better
screenshot from 2017-04-20 15-07-04

shaded-enmity:
Great, thanks! 🎉 Merging

PR #65, opened 2017-04-21
First cut of JBPMS/Drools in CentOS 6 Box

shaded-enmity:
- [x] Get rid of `install.sh` and use Ansible

**DEFERRED**

- [ ] Integration tests
  - [ ] Add new VMDef
  - [ ] Add new scenarios/steps
- [ ] Documentation


shaded-enmity:
Removing WIP here as there are two cards (impl & intg. tests) and the impl part here is done,

ncoghlan:
I went ahead and merged this - we can adjust as necessary when the integration tests are written.

PR #66, opened 2017-04-21
Port discovery feature

mfranczy:


src/setup.py
```diff
@@ -48,6 +48,7 @@
     install_requires=[
         "enum34",
```
mfranczy: I think we should freeze modules before summit

shaded-enmity: Yup, I think we should have a `Pipfile` + lockfile for `src/` as well.

shaded-enmity:
Jenkins Test please!

shaded-enmity:
The test failed due to `nmap` not being installed, can you add it to the list of packages in https://github.com/leapp-to/prototype/blob/master/centos-ci/ansible/roles/vagrant/tasks/main.yml#L18 ? Thanks

mfranczy:
I have added nmap, this time should works.

mfranczy:
this is how UI looks now:
screenshot from 2017-04-25 09-04-40
should I change something or it's ok?

vinzenz:
@mfranczy well the screenshot kind of already shows a problem we need to handle, the SSH mapping needs to be skipped

mfranczy:
@vinzenz that was only example to show how it looks. I will change that in next PR from https://trello.com/c/0kJ1S3we/33-automatic-port-mapping

shaded-enmity:
Jenkins Test please!

shaded-enmity:
Jenkins Test please!

shaded-enmity:
LGTM, merging! 🎉

src/setup.py
ncoghlan: If we're going to do that, it would probably make sense to give leapp-tool its own repo, distinct from the one for the integration tests and demo - `pipenv` is fairly opinionated about the one repo = one environment model, and at the moment we're not really abiding by that (hence the need for `pipsi` in the integration tests).

PR #71, opened 2017-04-25
Add README describing our CI setup

shaded-enmity:


centos-ci/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Architecture
+
+When Jenkins is triggered via a webhook it uses a Duffy key located in `~/duffy.key` to get a node from the CICO
+cluster on which the test will run. Duffy is CentOS CI middleware for resource management inside the cluster.
+The lifetime (read, debuggability) of a failed test run is 12 hours.
+
+## Configuration
+
+All configuration regarding the CI can be found in the `centos-ci` directory where this readme is located.
+Secrets and SSH keys can be found in the the housekeeping repository.
```
mfranczy: can you add a link to housekeeping repository?

ncoghlan: Typo: "the the"

centos-ci/README.md
```diff
@@ -0,0 +1,37 @@
+# LeApp CI
+
+LeApp CI is hosted on the [CentOS CI infrastructure](http://ci.centos.org/) using Jenkins and since
+our tests are virtualization heavy we have a dedicated slave node (slave03) for our tests.
+
+## Architecture
+
+When Jenkins is triggered via a webhook it uses a Duffy key located in `~/duffy.key` to get a node from the CICO
+cluster on which the test will run. Duffy is CentOS CI middleware for resource management inside the cluster.
+The lifetime (read, debuggability) of a failed test run is 12 hours.
```
ncoghlan: Any chance we could get that extended to 18 or 24 hours? 12 hours is cutting it fine for the "start a CI run at the end of the day, check it in the morning".

OTOH, I guess it encourages good test logging for post-mortem debugging, rather than expecting to be able to connect to the failed instance.

centos-ci/README.md
```diff
@@ -0,0 +1,37 @@
...
+Jenkins Job Builder file is a declarative way (YAML) of describing Jenkins Job, this file also describes who
+can start test runs and what trigger phrase to use for PRs not by people from the admin list.
+
+### Adding yourself to the list of people who can trigger a test run
+
+Add your GitHub handle to the `admin-list` list under `jobs.triggers.github-pull-request` and open a PR with the change.
+
+### The trigger phrase
+
+We're currently using `*Jenkins Test*` with asterisks being wildcards, when any user in the `admin-list` comments on a PR
```
ncoghlan: Could we make the command phrase a bit more obviously a command or directive, rather than an ordinary noun phrase? I'm thinking something like: "Jenkins: ok to test"

centos-ci/README.md
shaded-enmity: I'll ask on the CI mailing list, perhaps the timeout can be tweaked via the JJB file, it's good to have this documented. I think it would be better to stick with 12 hours and adopt good logging hygiene like you mention - after all, with longer timeout we might end up blocking *more* bare metal nodes and I'm afraid the CI guys wouldn't like us for that :)

centos-ci/README.md
shaded-enmity: Sounds good, updated the docs + the JJB definition. 

Also made me realize that I didn't document how/when the changes to the JJB files actually take effect in the CI, so added another section explaining that.

PR #73, opened 2017-04-25
Using a spec file for specifying required dependencies to run the demo

vinzenz:

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,11 @@ The migration prototype currently handles exactly two cases:
 
 ### Setting up to run the prototype demonstration
 
+Install all the requirements of this demo via:
+
+    $ sudo dnf dnf-plugins-core
+    $ sudo dnf builddep ./demo/prototype-deps.spec
+
 If the integration tests haven't been run, first install the testing
 instance of the CLI:
 
@@ -27,11 +32,6 @@ instance of the CLI:
 
 The integration tests do this automatically, so this step can be skipped if
 those have already been run.
-
-GCC is needed to compile vagrant plugins 
-
-    $ sudo yum install qemu libguestfs-tools-c libvirt libvirt-devel ruby-devel gcc qemu-kvm nmap
-
 
 The prototype requires Vagrant with some relevant plugins:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -20,6 +20,10 @@ The migration prototype currently handles exactly two cases:
 
 ### Setting up to run the prototype demonstration
 
+Install all the requirements of this demo via:
+
+    $ sudo yum-buildep ./demo/prototype-deps.spec
```
vinzenz: Should be:

    sudo dnf builddep ./demo/prototype-deps.spec

Because pipsi isn't even in epel so this can't even work on CentOS/RHEL right now, but that we can fix with some switches

vmindru:
had a quick talk with @vinzenz  we agreed to keep the spec file around for the general public.

demo/prototype-deps.spec
```diff
@@ -0,0 +1,26 @@
...
+BuildRequires: libguestfs-tools-c
+BuildRequires: libvirt-client
+BuildRequires: libvirt-devel
+BuildRequires: nmap
+BuildRequires: pipsi
+BuildRequires: python2-devel
+BuildRequires: redhat-rpm-config
+BuildRequires: rubygem-hitimes
+BuildRequires: rubygem-nio4r
+BuildRequires: vagrant-libvirt
```
shaded-enmity: To make this useful for the CI bootstrap script we need the following packages as well:

- epel-release
- openssl-devel
- python2-pip

Would be good to also make the `centos-ci/bootstrap.sh` use this new file.

vinzenz: This is not really a thing for the bootstrap - I added openssl-devel and python2-pip anyway
but bootstrap needs other things. Anyway I will try to create a copr repo and make a proper prototype rpm in a follow up and then we can see what to do about it - That should make things easier in general.

Anyway for this PR it's not planned to replace the boostrap for the CI just yet - Its meant as a convenience for setting up new dev / demo environments but won't do all the tasks

vmindru:
@vinzenz  are you still going to merge this? Please rebase if yes.

vinzenz:
not sure just yet - Putting it for now on WIP anyway

PR #77, opened 2017-04-26
Fix default IP in the drools Vagrant file

shaded-enmity:
The previous setting of '10.0.0.10' was clashing with the
definition of centos6-guest IP


vmindru:
+1 merge this

vmindru:
actually , don't we have to change also     config.vm.box = "centos/6" ?

vmindru:
ok let's merge this

PR #86, opened 2017-04-28
Specify username for remote access

ncoghlan:
The CLI is currently hardcoded to connect to remote systems as the
`vagrant` user. This PR adds an explicit `-u`/`--user` option to specify
a username for remote access, and otherwise defaults to the current
user.


ncoghlan:
Initial WIP PR is just the preparatory refactoring to give folks a chance to see the proposed restructure, the actual change is still to come :)

ncoghlan:
Jenkins test

ncoghlan:
I think the main open question here is how to implement the default behaviour. The current PR does it by explicitly defaulting to `getpass.getuser()`, but the alternative the would to instead default to `None`, and completely omit the `User` entry from the generated SSH config in that case.

vmindru:
Let's stick with None until we figure out a bullet proof way of "guessing" 

VM

shaded-enmity:
I like the `None` default option better -- both settings essentially reflect the way SSH works, if you don't explicitly specify user name the user name of invoking user is used, but the `None` way achieves that with less code.

ncoghlan:
I updated the PR to let SSH handle setting the default identity. The commit message reflects the fact that this also puts us in a better position to default to using ssh-agent - all we need to do is similarly make the other SSH configuration overrides optional.

Handling things that way also means we can skimp on the test coverage a bit and only cover the ssh-agent based flow in the Cockpit plugin integration tests (since we should have the latter merged before we enable the ssh-agent support)

ncoghlan:
Putting this back to WIP, as I expect it to conflict with #97 and I'd prefer to merge that one before merging this one.

ncoghlan:
Merging based on the prior reviews that requested letting `ssh` handle the default behaviour (which the patch now does), and were otherwise positive.

PR #89, opened 2017-04-28
Fix/drools

shaded-enmity:


ansible_roles/drools/tasks/main.yml
```diff
@@ -61,8 +61,17 @@
...
+  find:
+    path: '{{ jboss_home  }}'
+    file_type: directory
+    patterns: '*'
+  register: jboss_home_files
+
+
 - name: Move WildFly checkout from version specific directory
   shell: 'mv /tmp/wildfly/wildfly-{{ wildfly_version }}/* {{ jboss_home }}'
+  when: jboss_home_files.matched|int == 0
```
vmindru: nice

ansible_roles/drools/files/drools
```diff
@@ -31,6 +32,10 @@ case "$1" in
     if is_running; then
         echo "Already started"
     else
+        rm -rf /opt/jboss/wildfly/standalone/deployments/drools-wb.war.*
+        sudo -u $user touch /opt/jboss/wildfly/standalone/deployments/drools-wb.war.dodeploy
+        ps -eo pid,cmd a | grep 'standalon[e]' | cut -d' ' -f2 | xargs kill -9
```
mfranczy: maybe it's better to use `pgrep 'standalone[e]' | xargs kill -9` here

shaded-enmity: Much better, actually! Give me a minute :)

shaded-enmity: Updated, needed `-f` and since pgrep is clever enough to filter itself out we can drop the `[e]` as well :)

mfranczy: yes you're right :)

PR #91, opened 2017-04-28
Vagrant ssh scan

vinzenz:
This PR introduces a new fallback mechanism to use vagrant SSH  to scan the machine for OS info, ips and hostname

### DIFF
--- a/src/leappto/providers/libvirt_provider.py
+++ b/src/leappto/providers/libvirt_provider.py
@@ -1,10 +1,17 @@
-import libvirt
-from leappto import AbstractMachineProvider, MachineType, Machine, Disk, Package, OperatingSystem, Installation
-from xml.etree import ElementTree as ET
-from subprocess import check_output
-import socket
 import errno
 import json
+import libvirt
+import os
+import shlex
+import socket
+from io import BytesIO
+from subprocess import check_output
+from xml.etree import ElementTree as ET
+
+from paramiko import AutoAddPolicy, SSHClient, SSHConfig
+
+from leappto import AbstractMachineProvider, MachineType, Machine, Disk, \
+        Package, OperatingSystem, Installation
 
 class LibvirtMachineProvider(AbstractMachineProvider):
     def __init__(self, shallow_scan=True):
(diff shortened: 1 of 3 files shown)


shaded-enmity:
Jenkins Test

src/leappto/providers/libvirt_provider.py
```diff
@@ -125,9 +133,41 @@ def __inspect_os(domain_name):
...
+            domain_name = check_output(cmd, shell=True).strip()
+            cmd = """python -c 'import platform; print chr(0xa).join(platform.linux_distribution()[:2])'"""
+            output = __vagrant_ssh_checkoutput(domain_name, cmd)
+            distro, version = output.strip().replace('\r', '').split('\n', 1)
+            cmd = """python -c 'import socket; print socket.getfqdn()'"""
+            output = __vagrant_ssh_checkoutput(domain_name, cmd)
+            hostname = output.strip()
+            cmd = r"ip -4 -o addr list | grep -E 'e(th|ns)' | sed 's/.*inet \([0-9\.]\+\)\/.*$/\1/g'"
+            output = __vagrant_ssh_checkoutput(domain_name, cmd)
+            ips = [i.strip() for i in list(output.split('\n')) if i.strip()]
```
shaded-enmity: Nitpick: `list()` shouldn't be needed here

vinzenz: Right, that's a left over from some previous iteration will fix that

vinzenz: Done

shaded-enmity:
Jenkins Test

src/leappto/providers/libvirt_provider.py
```diff
@@ -92,16 +93,23 @@ def __get_value(elem):
             else:
                 return ''
 
+        def __virt_inspector_supports_shallow():
+            """
+            :return: True if virt-inspector supports --no-applications and --no-icon
+            """
+            inspector_version = check_output(['virt-inspector', '-V'])
+            major, minor, _ = inspector_version.split(' ')[1].split('.', 2)
+            return int(minor) >= 34 or int(major) > 1
```
ncoghlan: Due to the way Python defines tuple comparison, this can be shortened to:

    return (int(major), int(minor)) >= (1, 34)

src/leappto/providers/libvirt_provider.py
```diff
@@ -125,9 +133,41 @@ def __inspect_os(domain_name):
...
+        def __vagrant_ssh_checkoutput(domain_name, command):
+            with open(os.devnull, 'w') as devnull:
+                cmd = ["vagrant", "ssh", domain_name, "-c", command]
+                return check_output(cmd, stderr=devnull)
+
+        def __get_ssh_info(domain_name):
+            domain_name, _ = domain_name.split('_', 1)
+            cmd = 'vagrant global-status | grep {} | cut -d\\  -f1'.format(domain_name)
+            domain_name = check_output(cmd, shell=True).strip()
+            cmd = """python -c 'import platform; print chr(0xa).join(platform.linux_distribution()[:2])'"""
```
ncoghlan: Ah, interesting. As an added bonus, using Python level checks like this will generalise nicely to arbitrary remote system access (including via Ansible, if/when we go that way).

vinzenz: Applied in the next update

src/leappto/providers/libvirt_provider.py
vinzenz: Applied in the next update

ncoghlan:
Jenkins test

ncoghlan:
@vinzenz Something that occurs to me with the provisioner inventory based approach: are we sure that file only included currently running VMs? If it may include suspended and shut down VMs, we'll need to find a way to filter those out.

vinzenz:
@ncoghlan It just reads all the info, but we're still using the active domains from libvirt as source, also you can get the information from the machine index file, it has a field with the state. When it's running, it's up :-)

PR #94, opened 2017-05-02
Automatic port mapping

mfranczy:


cockpit/leapp.html
```diff
@@ -118,7 +118,8 @@
                 </div>
             </div>
         </div>
-        <button translatable="yes" id="migrate-button" class="btn btn-default btn-primary migrate-app pull-right" disabled>Migrate</button>
+        <button translatable="yes" id="migrate-button" class="btn btn-default btn-primary migrate-app pull-right"
+        disabled>Migrate</button>
```
vinzenz: Any reason for this line break?

cockpit/leapp.html
```diff
@@ -139,7 +140,18 @@
             list_local_vms();
         });
 
+        function set_lock_cmd() {
+            $("input[name=source-machine], input[name=target-machine]").attr("disabled", true);
+            $("#migrate-button").attr("disabled", true);
```
vinzenz: one could also use a new class 'ui-lockable' and use that as indicator, I guess that's easier, no?

cockpit/leapp.html
```diff
@@ -267,29 +294,142 @@
...
                 ! $("input[name=source-machine]:checked").length) {
                 return;
             }
-            $("#target-ports ul li").remove();
+
+            var ip_addr = $("input[name=target-machine]:checked").data('ip');
+
+            var checked_ports_cnt = $("#panel-body input.source-ports:checked").length;
+            var min = null;
+            var max = null;
```
vinzenz: can we call them min_port and max_port? min/max feels wrong especially since you use them as parameters in add_ports_range as well (Kind of confusing)

mfranczy: yes, good point

cockpit/leapp.html
mfranczy: +1

cockpit/leapp.html
mfranczy: because one line have more than 120 characters

cockpit/leapp.html
```diff
@@ -267,29 +292,142 @@
...
+            if (typeof(data.ports.tcp) !== undefined && data.ports.tcp) {
+                var busy_ports = data.ports.tcp  
+            } else {
+                var busy_ports = {};
+            };
+
+            var mapped_ports = map_ports(ports_scan_range, busy_ports, selected_ports);
+
+            $("#target-ports ul li").remove();
+            for (target_port in mapped_ports) {
```
vinzenz: I think in javascript one would write this:
```js
mapped_ports.forEach(function(target_port) {
    $("#target-ports ul").append(`
        <li>
            Port:
            <input type="text" size="5" name="${mapped_ports[target_port]}" value="${target_port}"
            class="target-ports" disabled />
        </li>
    `);
});
```

mfranczy: mapped_ports is an object, it's not an array, when it comes to code above, should be:
```
Object.keys(mapped_ports).forEach...
```
so I think it's better to use `in`

vinzenz: Right, my bad, overlooked the fact that this is a map

mfranczy:
I have to change one thing in CLI, give a possibility to set random port by docker for target.  It should works like that: if you want to expose port 9000 then check 9000-9025 ports on target host, select one from this range if free, if not, docker should select something for you.

I think it's pointless to scan again with shifted range (for example 9026-9051) if ports from previous range are busy. We have issue to change nmap to netstat and after that I will have information about all used ports on target, so algorithm will be different.

shaded-enmity:
+1 on doing a narrow-phase scan first, then letting Docker select the port. Only thing to keep in mind is that the port chosen by Docker will be only visible after the container has started (so at the end of the migration)

mfranczy:
jenkins: ok to test

shaded-enmity:
LGTM, one quick question - how does the person interacting with the Cockpit UI figure out which random port was chosen by Docker?

mfranczy:
for the demo it should be rare when docker is choosing random ports, so it's like you don't have this information in UI, although I will eliminate this problem in next PR because I will discover all used ports, not only some range.

PR #95, opened 2017-05-03
ci-bootstrap: Don't fail on things already installed

vinzenz:


shaded-enmity:
I don't like the `|| :` stuff because while it can solve the problem of already installed packages it also *masks* every other form of error. At least in the Ansible roles we should first save the installed/not-installed status in a register and add `when:` conditional. 
As for `yum` I think this is OK, since return codes of `yum` are just grotesquely inconsistent.

centos-ci/ansible/roles/vagrant/tasks/main.yml
```diff
@@ -29,26 +29,26 @@
     - nmap
 
 - name: Run VMs under root
-  lineinfile: 
+  lineinfile:
```
vinzenz: The only thing I did here was to remove the trailing space

ncoghlan:
Like @shaded-enmity, I don't like the fact that the approach here hides all `yum` errors, not just "package already installed" ones.

Perhaps we could just pass the `-t` switch to yum, so it doesn't treat "already installed" as an error?

ncoghlan:
Also, it would be good to separate the functional change here from the formatting change that strips the trailing spaces from the various files.

PR #97, opened 2017-05-03
Allow leapp-tool to be run as non-root user

ncoghlan:
Most leapp-tool operations should be executable
without local root privileges, so drop the
requirement to always run it as root.

Instead, "sudo" is used for the commands that
need to be run as root for interoperability
with the Vagrant based tests and demo plugin.

### DIFF
--- a/integration-tests/features/environment.py
+++ b/integration-tests/features/environment.py
@@ -189,7 +189,7 @@ class ClientHelper(object):
 
     @staticmethod
     def _run_leapp(*args):
-        cmd = ["sudo", _LEAPP_TOOL]
+        cmd = [_LEAPP_TOOL]
         cmd.extend(args)
         return _run_command(cmd, work_dir=str(_LEAPP_BIN_DIR), ignore_errors=False)
 
--- a/integration-tests/features/environment.py
+++ b/integration-tests/features/environment.py
@@ -1,5 +1,6 @@
 import contextlib
 import json
+import os
 import pathlib
 import shutil
 import subprocess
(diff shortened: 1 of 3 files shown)


ncoghlan:
OK, I'm pretty sure this is going to fail due to the git checking out the predefined testing key as world readable by default. I'm testing an update locally that fixes the permissions on that file prior to running the test suite.

This is also why the Cockpit plugin still runs leapp-tool as root - otherwise it would fail due to the overly lax permissions on the pre-configured key file.

vinzenz:
@ncoghlan I think your theory is a bit flawed at this point, because we currently need it to be root to be able to get all the running domains from libvirt.

Also you need to be able to access the root home folder to get the global machine state from vagrant if we rely on vagrant.

So until then we should create an epic and redesign everything we're doing right now and fix all the bits and pieces that require root and it doesn't seem even obvious.

Until then I would not merge this even if it seems like the CI works

ncoghlan:
The proposed change moves the `sudo` calls inside leapp-tool - as a result `virt-inspector` and `virt-tar-out` both still run as root, since the integration tests failed when that wasn't the case. (I don't just rely on the CI for testing, I test things locally as part of the change development process)

The test suite doesn't run vagrant as root, it runs it as the current user, so those files should be accessible regardless.

That said, as noted above, those kinds of concerns are why this PR *doesn't* change the way that the Cockpit plugin runs leapp-tool - it only changes the test suite.

For the plugin (at least initially), we'll probably end up going with the option of passing through the current user's ssh-agent settings as part of the environment, and getting the appropriate value for `--user` from the cockpit user status.

However, long term, there's zero reason for leapp-tool to have root access to the local system - this takes the first step towards that by moving the privilege escalation to the operations that currently need it.

ncoghlan:
Would restoring an early exit that just allows appropriate group membership as an alternative to actually being root address your concern?

This patch isn't intended to make leapp-tool work for entirely unprivileged users - it's intended to let most of the process (most importanly, the SSH calls) run as the user *if* they're a member of the `vagrant` and `libvirt` groups, while relying on `sudo` access for the parts that fail when run without first switching to running as root.

ncoghlan:
No worries - your concerns prompted to broaden the early exit check rather than dropping it entirely, and I think that's a more user friendly way to go (since it spots the configuration error early, and describes the two possible ways to resolve it, rather than expecting users to reverse engineer cryptic permissions errors)

PR #99, opened 2017-05-05
Tidy up CI testing output

ncoghlan:
- don't emit ANSI colour control codes in logs
- emit JUnit compatible XML test reports

### DIFF
--- a/integration-tests/.gitignore
+++ b/integration-tests/.gitignore
@@ -1,2 +1,2 @@
-# Symlink to virtualenv
-.venv
+# behave testing reports
+reports/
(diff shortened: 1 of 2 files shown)


integration-tests/.gitignore
```diff
@@ -1,2 +1,2 @@
-# Symlink to virtualenv
-.venv
```
vinzenz: Is there a reason to remove this from the .gitignore?

ncoghlan: Yeah, it's in the wrong place. The real one is one level up in the base `.gitignore`. I just hadn't noticed it was still here until I went to add the exclusions for the directory with the XML reports.

ncoghlan:
I went ahead and merged this, as the `behave` output is otherwise close to unreadable, and generating the Junit-compatible XML reports means we just need to work out how to get Jenkins to read and display them.

PR #102, opened 2017-05-05
Leapp packaging

vinzenz:


ncoghlan:
Jenkins test

leapp.spec
```diff
@@ -0,0 +1,121 @@
+Name:       leapp
+Version:    0.0.1
+Release:    3
+Summary:    leapp tool rpm
+
+Group:      Unspecified
+License:    GPLv2+
```
ncoghlan: We should resolve https://github.com/leapp-to/prototype/issues/36 before merging this so we don't inadvertently indicate conflicting licensing terms anywhere.

vinzenz: Yeah I know, but I had to put something in - I noticed later that you put there LGPL  

Anyway the COPYING file says it more clearly - It's TBD

leapp.spec
```diff
@@ -0,0 +1,121 @@
...
+%else
+BuildRequires:   python2-setuptools
+%endif
+
+BuildRequires:   libvirt-python
+BuildRequires:   libvirt-devel
+BuildRequires:   python-enum34
+BuildRequires:   python2
+BuildRequires:   python2-nmap
+BuildRequires:   python2-paramiko
```
vinzenz: Oh right it might not be necessary anymore because I am messing with the setup.py during the build to NOT require something

vinzenz: Done

src/leappto/providers/libvirt_provider.py
```diff
@@ -151,11 +151,6 @@ def __inspect_os(domain_name):
 
             return (hostname, Installation(OperatingSystem(distro, '{}.{}'.format(major, minor)), packages))
 
-        def __vagrant_ssh_checkoutput(domain_name, command):
```
vinzenz: no it doesn't belong in there ;-) I just haven't moved it out of here just yet but will do

cockpit/leapp.html
```diff
@@ -141,10 +141,10 @@
 
         function call_leapp(cmd_args) {
             return cockpit.spawn(
-                ["/opt/leapp/bin/leapp-tool"].concat(cmd_args),
+                ["/usr/bin/leapp-tool"].concat(cmd_args),
```
ncoghlan: The symlink creation instructions in the README (that cover running the demo from a checkout rather than an installed package) will need to be updated for this change, as will the integration tests (the plugin integration test isn't run by default, which is why the CI isn't complaining about it, but `behave --wip` will run it locally).

vinzenz: you're right and we'll have to eventually solve this - Probably through environment variables as you suggested - I have created issue #114 for it

vmindru:
can someone please revert this change, this has not been updated properly. 

when performing this type of changes we have to 

1) update docs
2) update CI 
3) update CD 
4) mail the mailing list ( maybe i guess this step can be optional )

ncoghlan:
@vmindru Ah, you're right, I forgot about the regression to existing setups. Sorry about that.

@vinzenz Could you submit a follow-up PR switching the plugin HTML back to using the non-FHS location and add a corresponding symlink to the package?

That seems like a better interim fix than full reversion.

PR #110, opened 2017-05-10
#103: Allow use of default SSH credentials in CLI

### LINKED ISSUE #103: Allow use of default SSH credentials in CLI
The CLI currently requires that a specific `--identity` be passed in for use with SSH, and aborts any subcommands requiring SSH if that argument is missing.

To allow the use of default SSH keys and ssh-agent, we want to make that setting optional, and delegate authentication to SSH if no identity file is given.

Actually making this change is straightforward (default the identity file to `None` and skip setting it in that case), but covering it in the integration tests would be a bit more involved.

Based on the work already done for the WIP Cockpit plugin integration test, one way to do that would be to:

- move the user definition helpers out of the plugin testing steps file and into the shared test context
- create a suitably configured test user and add the predefined testing key as their `~/.ssh/id_rsa/` key
- run the affected commands as that user and show that they still work in an end-to-end scenario

However, I'm wondering if it really makes sense to test this as an integration test - we can trust that end-to-end SSH with default credentials works, we just want to ensure that our CLI is configuring *SSH* as expected when no identity file is given.

A potentially more sensible alternative would be to start building out a CLI level unit testing capability - that could then test the CLI arguments -> SSH configuration aspect directly, and avoid running the full integration tests twice with different authentication methods.

ncoghlan:
When no --identity file is specified, the CLI now falls back
on the default SSH credentials (typically provided by ssh-agent)

Closes #103

### DIFF
--- a/integration-tests/README.md
+++ b/integration-tests/README.md
@@ -80,6 +80,8 @@ The following features are currently defined:
   subcommand when migrating stateless applications running under Apache `httpd`
 * `list-machines.feature`: expected behaviour of the `list-machines` subcommand
 * `port-inspect.feature`: expected behaviour of the `port-inspect` subcommand
+* `remote-authentication.feature`: end-to-end testing of available remote
+  authentication options
 
 If a new test scenario doesn't align with any of the existing features, then
 an appropriate new feature should also be defined.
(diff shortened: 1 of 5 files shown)


ncoghlan:
While working on this PR, I found a latent defect where destroy-containers wasn't actually setting a return code: https://github.com/leapp-to/prototype/pull/110/commits/4e56c3cb51cf3c825b95e417e76ee4b27264e269

ncoghlan:
As suspected, there's no ssh-agent running by default in the CI setup, so this is going to require some additional adjustments to temporarily run one in the affected scenario.

vinzenz:
LGTM

PR #109, opened 2017-05-10
merging all vagrant scripts into vmctl.sh single script

vmindru:
vagrant start is 4x faster than vagrant start --provision

PR #118, opened 2017-05-12
Fix ansible roles path broken in #109

vinzenz:


vmindru:
+1 good catch ...  wonder how our CD passed.

vmindru:
@vinzenz  can you also change https://github.com/leapp-to/prototype/blob/master/demo/vmctl.sh#L13 stop to halt and then we can merge this.

PR #119, opened 2017-05-15
Manual port mapping

mfranczy:
with two steps ports scan


integration-tests/features/leapp_testing/__init__.py
```diff
@@ -185,7 +187,8 @@ def redeploy_as_macrocontainer(self, source_vm, target_vm):
 
     def check_response_time(self, cmd_args, time_limit, *,
                             complete_user=False,
-                            complete_identity=False):
+                            complete_identity=False,
+                            as_sudo=False):
```
ncoghlan: #120 also makes changes to the parameters here, so these PRs are going to conflict

mfranczy:
ok, thanks

PR #120, opened 2017-05-16
Support --ask-pass option in CLI

ncoghlan:
This uses getpass.getpass() to interactively retrieve a password
from an interactive prompt, and then sshpass to pass it along
to ssh via a custom file descriptor.

The associated test case similarly uses sshpass to supply the
test password to the LeApp CLI.

To avoid problems with the default per-tty credential caching
in sudo, the test case only runs when the tests are being
run as root (as is the case in CI).

### DIFF
--- a/ansible_roles/sshd-password-auth/handlers/main.yml
+++ b/ansible_roles/sshd-password-auth/handlers/main.yml
@@ -1,0 +1,2 @@
+- name: Restart sshd
+  service: name=sshd state=restarted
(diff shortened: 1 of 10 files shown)


ncoghlan:
`sshpass` is a new dependency to handle securely passing the provided password to ssh, but isn't available by default on RHEL/CentOS. Required updates to handle that:

- raise a more helpful error if `--ask-pass` is used, but `sshpass` is unavailable
- update the CI configuration to ensure `sshpass` is available
- skip the test scenario if sshpass is unavailable (but don't add that until after it's working in CI)

ncoghlan:
On reflection, I think it makes more sense to let the test scenario fail if `sshpass` is unavailable, so I think this PR is complete now.

PR #121, opened 2017-05-16
Update readme with temporary user bug "fix"

JacobCZ:


README.md
```diff
@@ -67,7 +67,10 @@ Then, from the base of the local clone, run:
 
     $ sudo bin/leapp-tool migrate-machine \
            --identity integration-tests/config/leappto_testing_key \
-           --tcp-port 9000:9000 -t centos7-target centos6-app-vm
+           --tcp-port 9000:9000 --user=vagrant -t centos7-target centos6-app-vm
+           
+*Due to a "bug", you have to specify the "vagrant" username manually, otherwise the tool won'
+t be able to ssh into the machines.*
```
ncoghlan: This isn't a bug in the CLI, it's just a plain docs bug where we forgot to update this example after switching to letting ssh use its standard defaults if nothing was specified on the command line.

Since the default user in the test and demo VMs goes with the default testing key, I'd suggest placing the `--user` and `--identity` options on the same line, and either dropping the explanatory paragraph entirely, or else rewording it to explain that the default username comes from the fact we're using local Vagrant boxes, while the default public key file is loaded in by the Ansible playbooks.

ncoghlan:
OK, I've done some additional digging into the behaviour here (see https://github.com/leapp-to/prototype/issues/46#issuecomment-301983845 ), and it turns out there's been a regression where the `--user vagrant` setting was lost from the Cockpit plugin.

So in addition to the original docs bug where we missed updating this command, there's also a regression to be fixed in the Cockpit plugin (to add the hardcoded user setting back in).

vmindru:
@JacobCZ  are you going to implement this change suggested by Nick ?

JacobCZ:
@vmindru yes, working on it atm

PR #132, opened 2017-05-24
TRELLO#104 Inspect source machine only using ssh

vinzenz:
This PR removes usage of virt-inspector and ovirt-guest-agent which will also drop all remnants of the guest agent usage (e.g. channel creation and ansible roles)


src/leappto/providers/libvirt_provider.py
```diff
@@ -211,67 +137,14 @@ def __get_ssh_info(domain_name):
             cmd = "/sbin/ip -4 -o addr list | grep -E 'e(th|ns)' | sed 's/.*inet \\([0-9\\.]\\+\\)\\/.*$/\\1/g'\n"
             _, output, stderr = client.exec_command(cmd)
             ips = [i.strip() for i in output.read().split('\n') if i.strip()]
```
ncoghlan: This looks potentially problematic to me, as I'd expect it to give the local IPs as seen from the target machine, which aren't necessarily going to be the same as the machine's externally accessible IP address(es) due to intervening NAT bridges.

I'd be more confident in the correctness if we were relying on the Vagrant SSH config to the get IP address details for the source and target. (Suggesting we should probably rename this provide as `vagrant_provider.py`, since the direct `libvirt` dependencies are gone now, but the Vagrant dependency has been introduced along the way)

ncoghlan:
Merging, since there aren't any further changes needed specifically for this - we already know we need to tidy up & refactor the backend interaction in general, and this is just aimed at handling one piece of that (removing the `virt-inspect` and oVirt guest agent query options, since we don't need them any more)

PR #137, opened 2017-05-25
Custom target support in CLI and UI

vinzenz:
This adds a custom target to the cockpit UI and ensures that the command line also allows to use it on scan

Custom target entry


mfranczy:
you have bug (probably) with toggle, how to replay bug:
1. click Custom Target 
2. click centos7-target
3. click Custom Target
as result you will have hidden text input, but in that case should be visible

src/leappto/cli.py
```diff
@@ -248,14 +248,15 @@ def fix_systemd(self):
             machine_src = _find_machine(machines, source)
             machine_dst = _find_machine(machines, target)
 
-            if not machine_dst or not machine_src:
-                print("Machines are not ready:")
+            # We assume the target to be an IP or FQDN if not a machine name
+            ip = machine_dst.ip[0] if machine_dst else target
+
+            if not machine_src:
+                print("Machine are not ready:")
```
mfranczy: probably should be "Machine is not ready"

vinzenz: yes you're right - That was an oversight when removing the s ;-)

vinzenz: Done

vinzenz:
> you have bug (probably) with toggle, how to replay bug:
> 1. click Custom Target 
> 2. click centos7-target
> 3. click Custom Target
> as result you will have hidden text input, but in that case should be visible
Fixed as well

cockpit/leapp.html
```diff
@@ -226,12 +226,34 @@ <h3 class="mapped_ports">Mapped ports - Source &rarr; Target</h3>
...
+                    $("#custom_target").hide();
+                }
+            });
+            $("#custom_target")
+                .toggle()
+                .focusout(get_target_ports)
+                .change(function() {
+                    $("#machine-radio-custom-target").val($("#custom_target").val());
+                    $("#machine-radio-custom-target").data('ip', $("#custom_target").val());
+            });
```
mfranczy: one more thing, it would be good to add ```$("#machine-radio-custom-target").unbind('click');``` after
```$(".radio-target-machine").on('click..``` to prevent situation like scanning target ports without filled address

vinzenz: well the unfilled IP is handled this code just copies the value from the input to the value and the data-ip attribute of the  radio (for consistency with the other entries)

mfranczy: bad comment, it's better to check if address is empty like in 422 line

PR #138, opened 2017-05-25
Restore ability to run tests as non-root user

ncoghlan:
Due to recent changes in the way source & target VM lookup
is handled, the Cockpit plugin integration test now requires
that the VMs it uses be started by the root user (otherwise
it fails to find them).

While this isn't a problem for CI (where the tests run as
root), it causes problems when running the tests as a
normal user in a local `pipenv shell` session.

To fix that, the follow changes have been made:

- test scenarios can now indicate they need their VMs
  started as root, and if the tests are running as a
  normal user, the VMs will be destroyed and recreated
  by root
- test scenarios can now be marked with `@root_recommended`
  to indicate that they're really slow or otherwise
  problematic when run as a non-root user
- the Cockpit plugin integration test has been updated to
  use both of these features

### DIFF
--- a/integration-tests/features/steps/common.py
+++ b/integration-tests/features/steps/common.py
@@ -10,8 +10,9 @@ from hamcrest import assert_that, equal_to, not_none, greater_than
 # Local VM management
 ##############################
 
+@given("the local virtual machines as {user}")
 @given("the local virtual machines")
-def create_local_machines(context):
+def create_local_machines(context, user=None):
     """Accepts a table of local virtual machine definitions, consisting of:
 
     * name: the name to be used to refer to the VM in later test steps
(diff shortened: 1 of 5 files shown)


integration-tests/features/steps/common.py
```diff
@@ -22,14 +23,22 @@ def create_local_machines(context):
     even when *ensure_fresh* is set to 'no'. For faster test execution, relying
     on Ansible to restore the VM to a known state is recommended, rather than
     requiring a full VM create/destroy cycle.
+
+    When the tests are run as a non-root user, specifying 'as root' implies
+    'ensure_fresh=yes', even if it is otherwise set to 'no'
     """
+    as_root = (user == "root")
+    if user is not None and not as_root:
```
vinzenz: I might misunderstand this condition but it seems to be wrong
As root can only be true when user is not none
So if user is not none it should be enough according to the error message 

Anyway I think that there is something mixed up here

ncoghlan: I'll change it to a preceding `if user not in (None, "root"):` check so that it's clearer what's going on.

PR #145, opened 2017-06-01
Cli default port map

GAZDOWN:
Implemented CLI default port mapping


vmindru:
@GAZDOWN   i think you have to rebase

src/leappto/cli.py
```diff
@@ -114,6 +124,96 @@ def _find_machine(ms, name):
...
+
+        ## add user ports which was not discovered
+        for protocol in user_mapped_ports:
+            for port in user_mapped_ports[protocol]:
+                if not (protocol in source_ports and port in source_ports[protocol]):
+                    ## Add dummy port to sources
+                    source_ports[protocol][port] = None 
+
+        ## remap ports
+        for protocol in source_ports:
```
vinzenz:
```python
for protocol, port in user_mapped_ports.items(): 
    if port in user_mapped_ports:
....
```

vinzenz:
Jenkins: ok to test

GAZDOWN:
Jenkins: ok to test

shaded-enmity:
Jenkins: ok to test

GAZDOWN:
The test failed because port 80 was already used, so the tool mapped the new one to port 81 and test didn't expect that...

src/leappto/cli.py
```diff
@@ -122,6 +140,223 @@ def _find_machine(ms, name):
...
+
+
+        def __init_structure(self):
+            if not self.PROTO_TCP in self:
+                self[self.PROTO_TCP] = OrderedDict()
+        
+            if not self.PROTO_UDP in self:
+                self[self.PROTO_UDP] = OrderedDict()
+        
+        def __is_int(self, number):
```
mfranczy: you don't need that, check ```isinstance```

src/leappto/cli.py
```diff
@@ -122,6 +140,223 @@ def _find_machine(ms, name):
...
+            #    port_map = params[0]
+
+            if port_map and isinstance(port_map, OrderedDict):
+                self.__from_dict(port_map)
+            elif port_map:
+                raise AttributeError("Unknown port_map type")
+
+
+        def __init_structure(self):
+            if not self.PROTO_TCP in self:
```
mfranczy: I think it is unnecessary, you can just write ```self.tcp = OrderedDict()```  the same thing with udp

src/leappto/cli.py
```diff
@@ -122,6 +140,223 @@ def _find_machine(ms, name):
...
+                    continue
+
+                ports[proto][port] = port_scanner[ip][proto][port]
+
+        return ports
+
+
+    def _port_remap(source_ports, target_ports = PortMap(), user_mapped_ports = PortMap(), user_excluded_ports = PortMap()):
+        """
+        param source_ports      - ports found by the tool on source machine
```
shaded-enmity: Nice docstrings, but please tweak them so that they're machine readable i.e.:
`:param source_ports: description`

Rather than the current style.

src/leappto/cli.py
```diff
@@ -122,6 +140,219 @@ def _find_machine(ms, name):
...
+                self.__from_dict(port_map)
+            elif port_map:
+                raise AttributeError("Unknown port_map type")
+
+
+        def __init_structure(self):
+            self[self.PROTO_TCP] = OrderedDict()
+            self[self.PROTO_UDP] = OrderedDict()
+        
+        def __is_int(self, number):
```
GAZDOWN: This was removed

src/leappto/cli.py
```diff
@@ -122,6 +140,219 @@ def _find_machine(ms, name):
...
+            self.__init_structure()
+
+            if port_map and isinstance(port_map, OrderedDict):
+                self.__from_dict(port_map)
+            elif port_map:
+                raise AttributeError("Unknown port_map type")
+
+
+        def __init_structure(self):
+            self[self.PROTO_TCP] = OrderedDict()
```
GAZDOWN: Basically we do not, unless we want to keep order for user output. Which we actually want to for port inspect.

PR #147, opened 2017-06-01
Remove list-machines command

JacobCZ:
Fixes https://trello.com/c/AlY9vpyt/123-remove-the-list-machines-command-from-cli

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ __pycache__/
 playbook.retry
 dist/
 *.egg-info/
+.idea/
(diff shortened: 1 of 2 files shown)


ncoghlan:
Jenkins: test

Note that the tests are going to fail, as all the tests for this command will also need to be removed, as will the usage of it from the Cockpit plugin.

The documentation for this subcommand also needs to be removed: https://github.com/leapp-to/prototype/blob/master/docs/source/leapp-tool.rst#list-machines

ncoghlan:
Note that #201 removes the `list-machines` test cases (as a change to the test infrastructure broke them, and it didn't make sense to spend any time fixing them)

PR #148, opened 2017-06-01
Remove rest of vagrant refs

JacobCZ:
https://trello.com/c/dNR62PfZ/130-task%3A-remove-all-vagrant-references-from-leapp-%5Bhigh%5D

### DIFF
--- a/src/leappto/cli.py
+++ b/src/leappto/cli.py
@@ -23,7 +23,7 @@ except AttributeError:
     _set_inheritable = None
 
 # Checking for required permissions
-_REQUIRED_GROUPS = ["vagrant", "libvirt"]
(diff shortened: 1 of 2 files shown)


src/leappto/cli.py
```diff
@@ -49,12 +49,10 @@ def _make_argument_parser():
     ap.add_argument('-v', '--version', action='version', version=VERSION, help='display version information')
     parser = ap.add_subparsers(help='sub-command', dest='action')
 
-    list_cmd = parser.add_parser('list-machines', help='list running virtual machines and some information')
+
```
ncoghlan: This change should be part of the `list-machines` removal patch.

PR #149, opened 2017-06-01
Rsync source fs

mfranczy:


integration-tests/features/steps/common.py
```diff
@@ -47,21 +47,24 @@ def create_local_machines(context, user=None):
...
     context.redeployment_target = target_vm
-    result = context.cli_helper.redeploy_as_macrocontainer(source_vm, target_vm)
+    result = context.cli_helper.redeploy_as_macrocontainer(source_vm, target_vm, migration_opt)
     assert_that(result.local_vm_count, greater_than(1), "At least 2 local VMs")
     assert_that(result.source_ip, not_none(), "Valid source IP")
     assert_that(result.target_ip, not_none(), "Valid target IP")
     context.redeployment_source_ip = result.source_ip
     context.redeployment_target_ip = result.target_ip
 
+@when("{source_vm} is redeployed to {target_vm} as a macrocontainer and {migration_opt} is used for fs migration")
```
ncoghlan: `behave` allows a single function to be used to handle multiple step clauses , so you can apply this decorator directly to the original function definition above rather than needing to define a second function.

integration-tests/features/httpd-stateless.feature
```diff
@@ -7,3 +7,12 @@ Scenario: HTTP default server page
          | target     | centos7-target      | no           |
     When app-source is redeployed to target as a macrocontainer
     Then the HTTP 403 response on port 80 should match within 90 seconds
+
+@wip
```
ncoghlan: I believe this is complete now, so the `@wip` shouldn't need to be checked in.

integration-tests/features/steps/common.py
mfranczy: you're right, thanks for advice. Now it should be ok

integration-tests/features/httpd-stateless.feature
mfranczy: true, fixed.

mfranczy:
Jenkins: ok to test

src/leappto/cli.py
```diff
@@ -135,27 +142,64 @@ def _set_ssh_config(username, identity, use_sshpass=False):
...
-        def __init__(self, target, ssh_settings, disk, forwarded_ports=None):
+
+        _SSH_CTL_PATH = '{}/.ssh/ctl'.format(os.environ['HOME'])
+        _SSH_CONTROL_PATH = '-o ControlPath="{}/%L-%r@%h:%p"'.format(_SSH_CTL_PATH)
+
+        def __init__(self, source, target, ssh_settings, disk, forwarded_ports=None, rsync_cp_backend=False):
+            self.source = source
             self.target = target
-            self.use_sshpass, self.target_cfg = ssh_settings
+            self.use_sshpass, self.ssh_cfg = ssh_settings
```
shaded-enmity: I'm a bit on the fence by this rename - I know that the current implementation of `MigrationContext` is not exactly right, but the thing is that we're moving away from the Libvirt/Vagrant workflow towards arbitrary source/target workflow and in that sense I think it would be best to introduce a pair of variables (`self.target_cfg` and `self.source_cfg`) - each of them facilitating access to the respective machine, since each machine can have different SSH config/keys, and this change would be also well aligned with the below addition of `{source,target}_addr` properties.

mfranczy: ok no problem, I will change that

src/leappto/cli.py
```diff
@@ -135,27 +142,64 @@ def _set_ssh_config(username, identity, use_sshpass=False):
...
             if self.use_sshpass:
                 return self._sshpass(ssh_cmd, **kwargs)
             return Popen(ssh_cmd, **kwargs).wait()
 
+        def _open_permanent_ssh_conn(self, addr):
+            if not os.path.exists(self._SSH_CTL_PATH):
+                try:
+                    os.makedirs(self._SSH_CTL_PATH)
+                except OSError as exc:
+                    if exc.errno != 17:
```
shaded-enmity: One more nitpick: ```import errno``` and then use ```errno.EEXIST``` instead of the Magic Number :)

mfranczy: true! Shame on me.

src/leappto/cli.py
mfranczy: done, I think we have to refactor (or split) that class, I didn't want to do this in that PR.

src/leappto/cli.py
mfranczy: done

src/leappto/cli.py
```diff
@@ -135,27 +143,77 @@ def _set_ssh_config(username, identity, use_sshpass=False):
...
             self.disk = disk
+            self.rsync_cp_backend = rsync_cp_backend
             if forwarded_ports is None:
                 forwarded_ports = [(80, 80)]  # Default to forwarding plain HTTP
             else:
                 forwarded_ports = list(forwarded_ports)
             self.forwarded_ports = forwarded_ports
 
+        def __get_machine_opt_by_context(self, machine_context):
+            return (getattr(self, '{}_{}'.format(machine_context, opt)) for opt in ['addr', 'cfg', 'use_sshpass'])
```
mfranczy: I don't like that, but I didn't want to change a lot of functions and it was better to use class attr, we need to refactor that class.

mfranczy:
Jenkins: ok to test

ncoghlan:
Merging, as I'm pretty sure the Ansible-based container cleanup PR will conflict and I didn't want to break this one.

PR #150, opened 2017-06-02
Default target is localhost and source by address changes

vinzenz:


cockpit/leapp.html
```diff
@@ -8,155 +8,220 @@
...
+                    data-target="#credentials-dialog" class="alert-link" id="credentials-item">here</a>.
+                    <strong>RSync has to be installed on the source machine.</strong>
                 </div>
-                <div id="target-credentials" class="wd50-left border-left-dashed">
-                    <h3>Credentials to target host</h3>
+            </div>
+        </div>
+        <div id="analyze-stage" class="row section border-top-dashed">
+            <div class="col-md-6 col-sm-12">
+                <h3>Externally visible TCP ports</h3>
```
vmindru: I would just call this Discovered TCP ports

vinzenz: It's based on a mockup I received - So I just took the wording from there

ncoghlan:
@vinzenz I'm thinking that as part of this it will make sense to mark the current Cockpit plugin testing with `@skip`, and add a new testing scenario that:

1. Launches the target VM as usual
2. Runs a new `given` step that installs Cockpit
3. Runs a new `given` step that installs the Le-App plugin from the RPM built on the CI host
4. Triggers a source-to-local app import through the plugin running on the target VM

The existing test case would then only be restored once we add back the ability to do remote migrations through Cockpit (assuming we end up doing that at all - if it's in doubt, we could just delete that test scenario and the associated steps)

ncoghlan:
The old integration testing has been skipped (and could potentially just be deleted entirely for now), but it would be desirable to have a *new* integration test that checks importing an app through Cockpit running on the target system works as expected.

vinzenz:
@ncoghlan While I agree that this is necessary it's blocking too much progress right now to wait with this PR to move forward - We need to get these changes in to be able scaling our work better, right now a lot of people start to bottlenecks on my branch because they need to do UI changes, or need to get the changes in to remove CLI sub commands etc

I will start a new PR for the integration tests but this one really should get in as is

vmindru:
+1 let's merge this asis and start working on CI right after

vmindru:
oh just rebase it lol :)

vinzenz:
@vmindru Done - Rebased against master

ncoghlan:
@vinzenz Thanks for the clarification. In that case, +1 from me, too :)

PR #151, opened 2017-06-02
Split out target system cleanup

ncoghlan:
- moves mutable state to /var/lib/leapp
- clears it in the Ansible `target` role
- ensures Docker is installed before that role runs
- no longer implicitly deletes a container on migration
- no test case until early target validity checks are added


ncoghlan:
(I suspect this may conflict with some of the other pending PRs - if so, I'd suggest focusing on them for now, and I'll update this accordingly next week)

ncoghlan:
Ah, I forgot that `destroy-containers` was used for the authentication integration tests. I'll guess I'll have to fix it to be compatible with the updates, rather than just dropping it.

I'm not clear on what's going wrong with the rsync command, as that *was* working immediately after the merge - the refactoring to get virt-tar-out support working again must have broken it.

mfranczy:
I have checked your branch and rsync works fine, probably test needs more than 90 seconds to pass, but I didn't have this problem, maybe connection in your test was slower or something.

PR #153, opened 2017-06-06
Build RPM in pre-merge CI

ncoghlan:
Also fixes current tito build errors

### DIFF
--- a/centos-ci/bootstrap.sh
+++ b/centos-ci/bootstrap.sh
@@ -12,6 +12,12 @@ yum install -y python2-pip gcc redhat-rpm-config openssl-devel python-devel
 # Interim EPEL-based approach to enable testing of --ask-pass option
 yum install -y sshpass
 
+# Check the RPM can be built successfully
+yum install -y tito
+yum-builddep -y leapp.spec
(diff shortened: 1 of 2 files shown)


centos-ci/bootstrap.sh
```diff
@@ -12,6 +12,11 @@ yum install -y python2-pip gcc redhat-rpm-config openssl-devel python-devel
 # Interim EPEL-based approach to enable testing of --ask-pass option
 yum install -y sshpass
 
+# Check the RPM can be built successfully
+yum install tito
```
vinzenz: missing -y this will not work - AFAICT

ncoghlan:
The CI stalled because of another missing `-y` option, so I took the opportunity to clean up and rebase the PR while fixing that.

ncoghlan:
It turns out the cryptic "no job control message" almost certainly means RPM on CentOS 7 isn't recognising the `%py2_*` macros as valid, and hence the `%` is getting passed through to the shell.

ncoghlan:
The RPM build is working now, but there was a bug in the operand precedence for the build command that caused the bootstrap script to always exit immediately, even when the build succeeded.

I've fixed that, and also tidied up the commit history again.

PR #155, opened 2017-06-06
remove default forwarded ports

mfranczy:


mfranczy:
Jenkins: ok to test

src/leappto/cli.py
```diff
@@ -163,11 +163,7 @@ def __init__(self, target, target_ssh_cfg, disk, source=None, source_ssh_cfg=Non
             self._cached_ssh_password = None
             self.disk = disk
             self.rsync_cp_backend = rsync_cp_backend
-            if forwarded_ports is None:
-                forwarded_ports = [(80, 80)]  # Default to forwarding plain HTTP
-            else:
-                forwarded_ports = list(forwarded_ports)
-            self.forwarded_ports = forwarded_ports
+            self.forwarded_ports = list(forwarded_ports) if forwarded_ports is not None else list()
```
vinzenz: self.forwarded_ports = list(forwarded_ports or list())

vinzenz: instead of this is else here - Put forwarded_ports to default to () which is an empty tuple (Not a list because a list is a reference and not a value and it could remember it's content)

then you just convert the self.forwarded_ports to a list on this line

```python
self.forwarded_ports = list(forwarded_ports)
```
otherwise use the shorthand of this:
```python
self.forwarded_ports = list(forwarded_ports or ())
```

mfranczy: done

mfranczy:
Jenkins: ok to test

PR #160, opened 2017-06-08
Cli custom source

vinzenz:


src/leappto/cli.py
```diff
@@ -7,7 +8,12 @@
 from pwd import getpwuid
 from subprocess import Popen, PIPE
 from collections import OrderedDict
-from leappto.providers.libvirt_provider import LibvirtMachineProvider, LibvirtMachine
+from leappto import Machine
+from leappto.driver import LocalDriver
+from leappto.driver.ssh import LocalDriver, SSHDriver
```
ncoghlan: `LocalDriver` currently gets imported twice - I'm guessing the first one is the intended one.

vinzenz: right ,thanks

vinzenz: Fixed that

src/leappto/cli.py
```diff
@@ -494,8 +527,8 @@ def _rsync():
                                  .format(rsync_dir, ' '.join(self.target_cfg), self.target_addr)
                     Popen(shlex.split(target_cmd)).wait()
 
-                # temporary, after task with different names for containers should be removed
-                shutil.rmtree(rsync_dir)
+                    # temporary, after task with different names for containers should be removed
+                    shutil.rmtree(rsync_dir)
```
mfranczy: Is it ok? I mean now source fs dir will be removed only when target is different than localhost. I wanted to avoid situation where you have mixed files from more than one machine. I know common files are being replaced, but you can have some extra files from different source, for example mysql conf if it's running on specific machine etc.

shaded-enmity: I think this should be solved by picking unique name for the container (and hence the directory in which the files reside). But I'll wait for @vinzenz to clarify.

shaded-enmity:
LGTM, once @mfranczy's question is answered I'm ready to merge this.

src/leappto/cli.py
mfranczy: @shaded-enmity I agree, after task with unique names for containers ```shutil.rmtree``` should be removed and then we will be able to use power of rsync (I mean "quick check" algorithm in case we want to run migrate-machine again for the same source). Although it's not done yet, that's why I am asking if it's ok to change that line.

mfranczy: ok I am so stupid, of course this line must be changed (or moved before ```os.makedirs```), otherwise it's impossible to mount deleted directories for localhost... so I am ok with merge..

vinzenz: Yeah the files would have been gone otherwise :-)

vinzenz:
Since it's now all approved I will merge it now so others can start rebase their changes on top of it

PR #163, opened 2017-06-09
WIP: Check target command

### LINKED ISSUE #156: Add a `check-target` subcommand & use it in `migrate-machine`
Before starting an app migration, we want to be able to check if the target is suitable. We also want to be able to display a list of unavailable names based on containers and storage directories already present on the target.

The support that, we're going to add a new `check-target` subcommand that connects to the target and reports a combined list of names based on:

- the output of `docker ps -a`
- the contents of `/var/lib/leapp/macrocontainers/`

It will also ensure that it has the ability to escalate privileges on the target (as if it can't do that, any subsequent migration attempt won't actually work)

The `migrate-machine` command will also be updated to implicitly run `check-target` before it does anything else (since if that fails, or the desired container name is unavailable on the target, there's no point in continuing)

ncoghlan:
WIP as this depends on @vinzenz's refactoring to support the local machine as a target, and shouldn't be merged until after that has landed.

Closes #156


ncoghlan:
@vinzenz merged the CLI refactoring, so rather than trying to rebase this PR, I'm just going to cherry-pick the subcommand implementation into a fresh branch.

PR #164, opened 2017-06-12
check-target subcommand

ncoghlan:
- reports names already claimed on target system
- implicitly checks for privileged access to target

Additional changes:

- auth integration tests now use check-target
- new test setup step to claim names on target systems
- new test step to rerun the previous migration command

### DIFF
--- a/integration-tests/features/check-target.feature
+++ b/integration-tests/features/check-target.feature
@@ -1,0 +1,16 @@
+Feature: Checking usability of target virtual machine 
+
+Scenario: Checking for already used names on target VM
+   Given the local virtual machines:
+         | name       | definition          | ensure_fresh |
+         | target     | centos7-target      | no           |
+     And the following names claimed on target:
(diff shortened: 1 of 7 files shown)


ncoghlan:
Currently WIP as I want to tweak it to use @vinzenz's new remote access driver infrastructure, rather than the original SSH helper methods (which is what it currently uses)

ncoghlan:
I was originally planning to include a call to check target in migrate-machine as part of this PR, but it turns implementing that requires a significant refactoring of the way port scanning works (essentially, in order to allow the commands to be run in the desired sequence, the port scanning needs to be moved down to being a method on the migration context, since checking the target requires that the migration context already exist.

ncoghlan:
Since this is working now, and #176 should build on it, I'd like to postpone refactoring the implementation to use the new remote access driver capabilities: https://github.com/leapp-to/prototype/issues/177

Issue #165, opened 2017-06-12
nmap isn't called with elevated privileges as needed


ncoghlan:
Running the tests as a non-root user currently fails some tests with the following error:

    nmap.nmap.PortScannerError: u'You requested a scan type which requires root privileges.\nQUITTING!\n'

ncoghlan:
After looking into this, the problem is actually on the tests side - `migrate-machine` always needs to be run as root at the moment.

I'll include a fix for this in #167

vinzenz:
@ncoghlan Also port-inspect should be run as root

ncoghlan:
@vinzenz You're right, trying those locally, they're currently failing as a non-root user, and culprit is that the tests are only escalating privileges for `port-detect --shallow`, rather than escalating unconditionally.

ncoghlan:
Rather than continuing the current approach of scattering `as_sudo=True` through the test cases, I'm instead making the CLI helper aware of which commands inherently require root privileges, and having it implicitly set `as_sudo` in those cases.

PR #167, opened 2017-06-12
Avoid setting "ensure fresh" in the integration tests

### LINKED ISSUE #166: Integration tests are currently unusably slow
Several instances of `ensure_fresh=yes` have been added to the integration test suite, rendering it unusably slow for local development, since it's completely creating and destroying the VMs with Vagrant, rather than just pausing and resuming them, and the re-running the Ansible provisioning playbook.

Fixes needed:

- [x] Update the integration testing and code review docs to say *not* to check in tests with `ensure_fresh=yes`. It's a debugging tool for finding missing Ansible tasks and scenario-specific `given` steps, not a way we actually want test scenarios to be run (it costs 3-5 minutes every time, which is more than we want to pay)
- [x] Set `ensure_fresh=no` in `default-port-detection.feature`
- [x] Set `ensure_fresh=no` in `https-stateless.feature`

ncoghlan:
This attempts to speed up the test suite again be removing the
current usage of the "ensure fresh" option when setting up the
test VMs.

It also updates the integration test development and code
review documentation accordingly.

Fixes #166 and #165

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ __pycache__/
 playbook.retry
 dist/
 *.egg-info/
+docs/build/
--- a/Pipfile
+++ b/Pipfile
@@ -3,18 +3,20 @@ url = "https://pypi.python.org/simple"
 verify_ssl = true
 
 [dev-packages]
+# Docs build dependencies
+Sphinx = "*"
+sphinx_rtd_theme = "*"
+recommonmark = "*"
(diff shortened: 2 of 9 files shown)


ncoghlan:
The one actual failure reported is the second machine migration test, which suggests that the current cleanup of previously migrated containers isn't closing down the open HTTP listening socket properly.

PR #171, opened 2017-06-12
Change output if Source/Target machine is not accessible via SSH

artmello:
- Display error message in case of a SSH Connection exception is raised;
- Check if target machine is accessible;
- Do not  show traceback in case of a SSH Connection exception.


vinzenz:
Jenkins: ok to test

PR #176, opened 2017-06-14
[Proposal] Derive container name from source hostname

artmello:
Based on a TODO comment present on source code, this PR proposes:

- Instead of creating a container and a macro container directory with name "container" use "container_<source_hostname>" to make it easier to know from which machine such container was created;
- Enable user to define a specific container name during migration using --container-name or -n options.

### DIFF
--- a/src/leappto/cli.py
+++ b/src/leappto/cli.py
@@ -487,12 +487,16 @@ def main():
         def _ssh_sudo(self, cmd, **kwargs):
             return self._ssh("sudo bash -c '{}'".format(cmd), **kwargs)
 
+        def _get_container_name(self):
+            name_list = ["container", self.source.hostname]
+            return '_'.join(name_list)
+
         def _get_container_dir(self):
-            # TODO: Derive container name from source host name
-            container_name = "container"
+            container_name = self._get_container_name()
             return os.path.join(MACROCONTAINER_STORAGE_DIR, container_name)
 
         def copy(self):
+            container_name = self._get_container_name()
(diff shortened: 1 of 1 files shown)


ncoghlan:
Once https://github.com/leapp-to/prototype/pull/164 lands (thus switching the remote authentication tests to using `check-target` as their underlying subcommand), I'm hoping we can just delete the `destroy-containers` subcommand, and leave that entirely to the Ansible cleanup scripts.

artmello:
Right, once #164 lands the changes related with destroy-containers subcommand will be unnecessary. If the container name proposal part get an approval I will rebase this PR.

artmello:
Code is merged with changes added by #164. Should be good for another review

ncoghlan:
Jenkins: ok to test

src/leappto/cli.py
```diff
@@ -507,12 +509,19 @@ def _ssh_sudo_out(self, cmd, **kwargs):
...
+            return '_'.join(name_list)
+
         def _get_container_dir(self):
-            # TODO: Derive container name from source host name
-            container_name = "container"
+            container_name = self._get_container_name()
             return os.path.join(MACROCONTAINER_STORAGE_DIR, container_name)
 
         def copy(self):
+            container_name = self._get_container_name()
```
mfranczy: Please remove lines 531-532. Since we have different names for containers, removing container dir is no needed.

artmello: done, thanks for reviewing

PR #179, opened 2017-06-14
Removal of trailing spaces in Python code

vinzenz:


shaded-enmity:
Makes me wonder if we want to run `pep8` tool or something like that after/before `behave` tests.

vinzenz:
I actually would go so far and go for flake8 checks - Maybe even using tox (as suggested by @mfranczy)
and we might want to consider unit tests as well

vmindru:
+1 flake8

vmindru:
yeah and +1 for tox, I have a positive experience using this within Ansible project integrates well with VIM.

PR #180, opened 2017-06-14
Check target access early in migrate-machine

ncoghlan:
- migrate-machine port remapping moved into MigrationContext
- target access checked before calculating port mapping
- port mapping helpers moved to CLI module top level
- related test steps enabled

### DIFF
--- a/src/leappto/cli.py
+++ b/src/leappto/cli.py
@@ -159,6 +159,7 @@ def main():
 
     ap = _make_argument_parser()
 
+    # TODO: Move these helper functions into the leappto library
     def _find_machine(ms, name, shallow=True, user='root'):
         for machine in ms:
             if machine.hostname == name:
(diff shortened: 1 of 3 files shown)


src/leappto/cli.py
```diff
@@ -676,42 +519,42 @@ def print_migrate_info(text):
...
+            # target, and the desired container name is available
+            check_result, claimed_names = mc.check_target()
+            if check_result != 0:
+                print("! Checking target access failed")
+                sys.exit(check_result)
+
+            container_name = mc.get_target_container_name()
+            if container_name in claimed_names:
+                print("! Container name {} is not available".format(container_name))
+                sys.exit(-10)
```
shaded-enmity: I wonder if we should start using errno's as exit codes (plenty of tools do) and rather return `errno.EEXIST` here since `-10` feels rather arbitrary.

shaded-enmity: Also, we should probably unify the usage of `sys.exit` and the builtin `exit`.

shaded-enmity:
Left 2 comments inline, however I don't consider them merge blockers, just something we might wanna address in the future so I'm going to file 2 issues about this.

Issue #190, opened 2017-06-16
 ls: cannot access '/var/lib/leapp/macrocontainers/': No such file or directory; with freshly installed leapptool


GAZDOWN:
All in default, Access cockpit -> Import App -> Click find app.

Linux leapp-f25.mgazdik.lan 4.10.16-200.fc25.x86_64 #1 SMP Mon May 15 15:19:52 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

[root@leapp-f25]~/prototype# rpm -qa | grep leapp
leapp-cockpit-0.0.1-32.git.587.c162f68None.noarch
python2-leapp-0.0.1-32.git.587.c162f68None.noarch
leapp-tool-0.0.1-32.git.587.c162f68None.noarch

2017-06-16T08:52:49.641Z - /usr/bin/leapp-tool port-inspect 127.0.0.1 --shallow
2017-06-16T08:52:49.642Z - Scanning ports on 127.0.0.1
2017-06-16T08:52:49.942Z - /usr/bin/leapp-tool check-target 127.0.0.1
2017-06-16T08:52:50.186Z - > ls: cannot access '/var/lib/leapp/macrocontainers/': No such file or directory

trace:
subprocess.py(119):     while True:
subprocess.py(120):         try:
subprocess.py(121):             return func(*args)
subprocess.py(1084):                 if pid == self.pid:
subprocess.py(1085):                     self._handle_exitstatus(sts)
 --- modulename: subprocess, funcname: _handle_exitstatus
subprocess.py(1032):             if _WIFSIGNALED(sts):
subprocess.py(1034):             elif _WIFEXITED(sts):
subprocess.py(1035):                 self.returncode = _WEXITSTATUS(sts)
subprocess.py(1071):             while self.returncode is None:
subprocess.py(1086):             return self.returncode
subprocess.py(1119):             return (stdout, stderr)
cli.py(303):             if err_output:
cli.py(304):                 sys.stderr.write(err_output + b"\n")
ls: cannot access '/var/lib/leapp/macrocontainers/': No such file or directory

cli.py(305):             return child.returncode, output
 --- modulename: subprocess, funcname: __del__
subprocess.py(438):         if not self._child_created:
subprocess.py(442):         self._internal_poll(_deadstate=_maxint)
 --- modulename: subprocess, funcname: _internal_poll
subprocess.py(1050):             if self.returncode is None:
subprocess.py(1065):             return self.returncode
[...]

PR #193, opened 2017-06-16
Check-target do not fail if storage dir is missing on target machine

artmello:
While listing containers on target machine do not fail if missing macro containers storage dir, as seem on #190


vinzenz:
Jenkins: ok to test

PR #194, opened 2017-06-16
Agent Socket Forwarding

vinzenz:


vinzenz:
@ncoghlan The integration test part is some kind of proposal for now - But this needs other changes which are actually facilitating the new UI - Which I tried to to mix into this PR

ncoghlan:
Aye, there's a follow-up task to actually enable the Cockpit integration test again, but that's a much bigger change that involves adapting it from the current VM->VM migration flow into a VM->host application import test.

PR #196, opened 2017-06-19
Enable check-target to show availability of docker/rsync on target machine

artmello:


artmello:
As mentioned on the description, I am changing the output from "check-target" to also display status of target services (i.e. docker, rsync) . I took a very naive approach of only printing result as text output. This does have an impact on integrated tests and, probably, on UI since "check-target" is not returning a list of containers names any more.

I was thinking on two different solutions:
- Outputting data in a more structured format (i.e json) and so callers would be able to parser output properly, or;
- Adding options to check-target like: -containers_name, -docker, -rsync. And based on this input returned the desired output (i.e return code, list of containers).

Also, only after proposing this I checked the Trello card mentioning that maybe @ncoghlan was already working on a solution for this problem. If that is the case, please, just close this PR.

ncoghlan:
I think adding a `--json` option will be the way to go for implementing this in a backwards compatible way, and then at a later point we can look at making `--json` the default.

src/leappto/cli.py
```diff
@@ -384,7 +384,16 @@ def _virt_tar_out():
                 return _rsync()
             return _virt_tar_out()
 
-        def check_target(self):
+        def check_service_status(self, name, check_cmd):
+            rc = self._ssh_sudo(check_cmd)
+            print("! Check {} availability at {}: {}".format(
```
shaded-enmity: This looks a bit confusing, the parameters to format are vertically close to `print`, but still not quite there with the first argument, I'd suggest changing the code so that the argument to `format` are aligned with `format` in pep8 manner:

```python
msg = "! Check {} availability at {}: {}"
print(msg.format(name, self.target.hostname,
                 "success" if not rc else "error"))
```

artmello:
I proposed using a --status/-s together with check-target instead of --json since data returned will not be the same if option is not provided. Without --status check-target will only return claimed names on target machine to keep compatibility.

src/leappto/cli.py
```diff
@@ -628,8 +634,26 @@ def print_migrate_info(text):
             None
         )
 
-        return_code, claimed_names = mc.check_target()
-        for name in sorted(claimed_names):
+        check_result = dict()
```
shaded-enmity: `check_result = {}` should be more idiomatic

src/leappto/cli.py
```diff
@@ -628,8 +634,26 @@ def print_migrate_info(text):
...
+        return_code, check_result['containers'] = mc.check_target_containers()
+
+        if parsed.status:
+            check_commands = dict()
+            check_commands['docker'] = 'docker info'
+            check_commands['rsync'] = 'rsync --version'
+
+            for service in check_commands.keys():
```
shaded-enmity: I'd suggest `for service, command in check_commands.items()` to save the indirection below.

artmello:
@vinzenz / @ncoghlan I dont see to be able to remove the wip tag, but  I finished the suggested changes on this PR. It should be ready for another review round

ncoghlan:
The `--status` approach looks good to me. Since this isn't urgent, it should probably be updated to use the new SSH driver infrastructure though, rather than the old helper methods.

The `check-target` feature tests should also be updated to cover the new capability.

ncoghlan:
Contradicting my earlier refactoring comment: since we're going to need to refactor `check-target` later anyway, it probably makes more sense to keep it self-consistent for now, and then refactor the whole thing to use the remote access driver infrastructure later.

This does need additional test cases, though.

ncoghlan:
I added the WIP label back, as this still needs test cases.

artmello:
@ncoghlan, new test added.

integration-tests/features/check-target.feature
```diff
@@ -14,3 +14,18 @@ Scenario: Checking for already used names on target VM
...
+     And the following names claimed on target:
+         | name       | kind                |
+         | app1       | idle macrocontainer |
+         | app2       | idle macrocontainer |
+         | container1 | idle container      |
+         | container2 | idle container      |
+         | storage1   | macrocontainer dir  |
+         | storage2   | macrocontainer dir  |
+    Then checking target services status should take less than 10 seconds
+     And all status reported are correct
```
ncoghlan: This might be clearer for readers if we spelled out the expectations as distinct steps:

```
 And docker should be reported as "ok"
 And rsync should be reported as "ok"
 And all claimed names should be reported exactly once
```

To allow the existing definition for the last step to be re-used, `run_check_target_status` could do `context._reported_names = context._target_status['containers']`

src/leappto/cli.py
```diff
@@ -780,8 +785,26 @@ def print_migrate_info(text):
             None
         )
 
-        return_code, claimed_names = mc.check_target()
-        for name in sorted(claimed_names):
+        check_result = {}
+        return_code, check_result['containers'] = mc.check_target_containers()
+
+        if parsed.status:
+            check_commands = dict()
```
ncoghlan: I'd prefer to keep all the logic here inside the MigrationContext, so I'd suggest moving the current `check_target` to `check_target_containers` (as you've already done), but then re-using the `check_target` method name for the full status check.

The CLI logic would then look like:

```
if parsed.status:
    return_code, target_status = mc.check_target()
    print(dumps(target_status))
    sys.exit(return_code)

return_code, claimed_names = mc.check_target_containers()
# Existing code
```

src/leappto/cli.py
```diff
@@ -703,7 +708,7 @@ def print_migrate_info(text):
         if not parsed.print_port_map:
             # If we're doing an actual migration, check we have access to the
             # target, and the desired container name is available
-            check_result, claimed_names = mc.check_target()
+            check_result, claimed_names = mc.check_target_containers()
```
ncoghlan: Given the suggested refactoring below, this should use the new version of `mc.check_target()` so that it also ensures that the required commands are available.

src/leappto/cli.py
```diff
@@ -780,8 +785,26 @@ def print_migrate_info(text):
...
         )
 
-        return_code, claimed_names = mc.check_target()
-        for name in sorted(claimed_names):
+        check_result = {}
+        return_code, check_result['containers'] = mc.check_target_containers()
+
+        if parsed.status:
+            check_commands = dict()
+            check_commands['docker'] = 'docker info'
```
ncoghlan: I'm not sure this `docker` check is quite sufficient - due to the volume mount, we not only need the docker client to be installed, we also need the docker daemon to be running directly on the target system, rather than the client connecting to a daemon hosted elsewhere.

artmello:
@ncoghlan I think most of your suggestions are addressed. Only missing the improvement of docker check. Any suggestion of what a properly test would be? Do you think something like a `pgrep -f docker` should be enough?

ncoghlan:
I filed a separate issue for the docker check improvements: https://github.com/leapp-to/leapp/issues/260

ncoghlan:
Looks good to me, so merged! :)

PR #198, opened 2017-06-20
Selective container removal and forced creation

ncoghlan:
- adds `migrate-machine --force-create`
- replaces `destroy-containers` with `destroy-container <name>`

Also updates the integration tests and integrating testing
infrastructure accordingly.

### DIFF
--- a/integration-tests/features/steps/destroy_containers.py
+++ b/integration-tests/features/steps/destroy_containers.py
@@ -4,7 +4,7 @@ from hamcrest import assert_that, equal_to, not_, is_in
 
 
 @then("destroying {container} on {target} should take less than {time_limit:g} seconds")
-def check_destroy_containers(context, container, target, time_limit):
+def check_destroy_container(context, container, target, time_limit):
     """Destroy named container on named VM"""
     claimed_names = getattr(context, "_claimed_names")
     if not claimed_names:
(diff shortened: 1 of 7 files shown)


integration-tests/features/steps/destroy_containers.py
```diff
@@ -1,12 +1,29 @@
 """Steps to test the "destroy-containers" subcommand"""
 from behave import then
+from hamcrest import assert_that, equal_to, not_, is_in
 
 
-@then("destroying existing macrocontainers on {target} should take less than {time_limit:g} seconds")
-def check_destroy_containers(context, target, time_limit):
-    """check if exisiting containers have been destroyed"""
+@then("destroying {container} on {target} should take less than {time_limit:g} seconds")
+def check_destroy_containers(context, container, target, time_limit):
```
vinzenz: This should be singular now

src/leappto/cli.py
```diff
@@ -387,14 +392,17 @@ def _virt_tar_out():
...
             if rc:
                 return rc, []
             names = containers.splitlines()
             ls_storage_directories = 'ls -1 "{}"'.format(storage_dir)
-            rc, dirs = self._ssh_sudo_out(ls_storage_directories)
-            if not rc:
-                names.extend(dirs.splitlines())
+            rc, dirs = self._ssh_sudo_out(ls_storage_directories,
+                                          machine_context=self.TARGET)
+            if rc:
```
vinzenz: Why did you revert this? 
It needs to stay:
```python
if not rc:
    names.extend(dirs.splitlines())
```

because otherwise you fail when the directory is missing on the target which is valid, the target directory is enough to be created on the migrate-machine call itself

PR #201, opened 2017-06-21
Test app migration to the host running LeApp

ncoghlan:
- adds Docker to the CI Ansible roles
- adds an integration test for app import to localhost
- always uses `--force-create` for the app import test
  so we don't need a separate clean up step when
  repeatedly running the tests


ncoghlan:
Note: this currently leaves the migrated container lying around if you run the tests locally rather than on an ephemeral CI system. Perhaps the import step should register a scenario cleanup command to destroy it?

ncoghlan:
Hmm, the last test failure is because I now implicitly add "localhost" to the list of registered machines in the VM helper, which confused some of the logic in the "list-machines" tests.

Since list-machines is going away anyway, maybe we should just delete those tests and the associated step definitions?

vinzenz:
@ncoghlan yes we definitely should drop those list-machine tests

But you should sync with @jmikovic he is working on the drop of the command

ncoghlan:
Jenkins: OK to test

Re-testing, since the failure seems to be an odd nmap-related one. #202 (which changes local port scans to use psutil rather than nmap) may also turn out to be relevant here.

ncoghlan:
@vinzenz @mfranczy So this passed on the re-test without any changes, but that just suggests the nmap failure is intermittent. Should I go ahead and merge it as-is, or should we wait for the use-psutil-on-localhost change in #202 first?

ncoghlan:
I'm going to go ahead with merging this as-is, since I want to push the follow up PR with the Cockpit plugin tests for review. If it turns out to be flaky, then we can focus on getting #202 merged.

PR #202, opened 2017-06-21
Bugfix: port scan

mfranczy:


src/leappto/cli.py
```diff
@@ -773,28 +772,40 @@ def set_port(self, protocol, source, target = None):
...
+                if port_scanner[ip][proto][port]['state'] in ('open', 'filtered'):
+                    port_list.set_port(proto, port, port_scanner[ip][proto][port])
+        return port_list
 
-            ports.set_port(proto, port, port_scanner[ip][proto][port])
+    def _net_util(port_list):
+        sconns = psutil.net_connections(kind=port_list.PROTO_TCP)
+        for sconn in sconns:
+            addr, port = sconn.laddr
+            if not port_list.has_port(port_list.PROTO_TCP, port):
```
mfranczy: Port mapping algorithm doesn't support ports with various ips, that's why I check if port is open on the local machine (I don't care where it's assigned, docker will bind port to ::: as default, so have to check everything). I will create a card which will cover above feature.

shaded-enmity:
Looks good, although I'm not really sure what bug is this fixing, could you perhaps add small description in the commit message explaining what changed and why?

mfranczy:
Sure I will add also as a commit message. Short explanation, earlier we were scanning localhost by nmap and we had ip from LocalMachine object. It was like: nmap -sS 10.43.12.171 as result we have ports which are assigned to ip 10.43.12.171 but we need info about all used ports. It was better just to check all by psutil.net_connections than nmap localhost with some options.

I need information about all used ports because docker is binding them to ```:::``` as default (don't know if we want to change that, probably topic for grooming)

ncoghlan:
https://ci.centos.org/job/leapp-integration-tests/354/console shows a test failure for #201 that looks like it's related to running `nmap` against the local host.

mfranczy:
please don't merge, I would like to talk again with @shaded-enmity about my last commit

PR #203, opened 2017-06-21
ui: Add custom port mapping form

GAZDOWN:


cockpit/leapp.html
```diff
@@ -241,6 +296,61 @@ <h4 class="text-center">Application names already in use:</h4>
             $(".analyze").fadeIn();
         });
 
+        $("#map-new-port-show").click(function(){
+            $("#map-new-port-form").find("input[type=text]").val("")
```
mfranczy: please use semicolons, like everywhere in the code, probably we should use strict mode. Maybe it's a good idea to assign #map-new-port-form to some var?

cockpit/leapp.html
```diff
@@ -241,6 +296,61 @@ <h4 class="text-center">Application names already in use:</h4>
...
+            $("#map-new-port-add-form").removeClass("hidden")
+
+            $("#add-port-src").focus()
+        });
+
+        $("#map-new-port-form").submit(function(event){
+            event.preventDefault()
+            $("#map-new-port-form").find("input[type=text]").parent().removeClass("has-error")
+
+            src_element = $("#add-port-src")
```
mfranczy: it's a common practice to add $ before var name in that case, it's useful because it informs you that you keep jQuery selection, so would be great if you could add this, btw why you don't use ```var``` here?

vinzenz:
Already merged

PR #205, opened 2017-06-22
Restore Cockpit plugin test

ncoghlan:
- updated for the new app import workflow
- updated to use ssh-agent for authentication

(WIP as the ssh-agent part isn't working yet)

### DIFF
--- a/integration-tests/features/steps/cockpit_demo.py
+++ b/integration-tests/features/steps/cockpit_demo.py
@@ -36,33 +36,27 @@ def visit_demo_page(context, menu_item):
     session.login()
     session.open_plugin(menu_item)
 
-@then("the local VMs should be listed within {time_limit:g} seconds")
-def check_demo_machine_listing(context, time_limit):
-    """Checks contents and response time for the demo plugin"""
+@when("enters {source_vm}'s IP address as the import source")
+def step_impl(context, source_vm):
     session = context.demo_session
-    expected_machines = context.vm_helper.machines
-    expected_hosts = sorted(host for name, host in expected_machines.items())
-    missing = session.list_missing_machines(expected_hosts, time_limit)
-    assert_that(missing, equal_to([]), "Machines missing from listing")
+    source_ip = context.vm_helper.get_ip_address(source_vm)
+    context.demo_user.register_host_key(source_ip)
+    session.enter_source_ip(source_ip)
+    session.wait_for_active_import_button(10)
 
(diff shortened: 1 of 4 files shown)


ncoghlan:
When running this locally, the failure mode I get is that Cockpit doesn't actually load the demonstration user's keys, and I assume that's why LeApp subsequently can't find a key to use.

integration-tests/features/steps/cockpit_demo.py
```diff
@@ -36,33 +36,27 @@ def visit_demo_page(context, menu_item):
...
-@when("the demonstration user redeploys {source_vm} to {target_vm}")
-def redeploy_vm_via_plugin(context, source_vm, target_vm):
+@when('clicks the "Import" button')
+def import_app_via_plugin(context):
     session = context.demo_session
-    gethost = context.vm_helper.get_hostname
-    session.start_vm_redeployment(gethost(source_vm), gethost(target_vm))
-    session.wait_for_redeployment_to_start(10)
+    session.start_app_import()
+    input("Hit enter to continue testing")
```
ncoghlan: This is for debugging, since it pauses the test while the failed command is displayed in the browser.

vinzenz: and how is this supposed to be cancelled on jenkins?

ncoghlan:
@vinzenz I've put a `sys.stdin.isatty()` guard in front of the debugging helper now, so even if I forget to remove it from a WIP PR, it won't block in Jenkins.

ncoghlan:
The code to set up the test user's SSH keys had a permissions bug that meant Cockpit initially couldn't read the user's SSH key directory, and some of the checks for a successful command needed additional updates to account for the new UI.

ncoghlan:
CI failure indicates that the UI is reporting something went wrong immediately on starting the application, so I'm going to tweak the progress display such that the tests can find and print all the progress messages after something goes wrong.

ncoghlan:
Rebased and squashed, so we can readily revert this if we decide we need to do so.

ncoghlan:
Jenkins: ok to test

The test failure looks to be an unrelated transient one that I'd only seen once before. Now filed as an issue here: https://github.com/leapp-to/prototype/issues/219

ncoghlan:
Jenkins: ok to test

Another instance of #219

ncoghlan:
Latest run has hit #219 again, so I'm going to go ahead and merge this anyway. @mfranczy is looking into the problem with nmap.

PR #123, opened 2017-05-17
Enable Cockpit plugin integration testing

ncoghlan:
To help avoid further regressions along the lines of #122, this enables
execution of the Cockpit plugin integration tests during CI

PR #206, opened 2017-06-22
cli: remove list machines cmd

jmikovic:
[#123](https://trello.com/c/AlY9vpyt/123-remove-the-list-machines-command-from-cli)
list-machine cmd no longer exists and libvirt is not used


vinzenz:
Jenkins: ok to test

src/leappto/cli.py
```diff
@@ -614,10 +598,7 @@ def print_migrate_info(text):
     elif parsed.action == 'check-target':
         target = parsed.target
 
-        lmp = LibvirtMachineProvider()
-        machines = lmp.get_machines()
-
-        machine_dst = _find_machine(machines, target)
+        machine_dst = _inspect_machine(target)
```
vinzenz: Seeing this I think we need to have the custom user support here - That however would be a different PR

src/leappto/cli.py
```diff
@@ -637,10 +618,7 @@ def print_migrate_info(text):
     elif parsed.action == 'destroy-container':
         target = parsed.target
 
-        lmp = LibvirtMachineProvider()
-        machines = lmp.get_machines()
-
-        machine_dst = _find_machine(machines, target)
+        machine_dst = _inspect_machine(target)
```
vinzenz: Same here for the custom user support

GAZDOWN:
I am also missing the documentation update. We should have it removed form leapp-tool page.

vinzenz:
Taken over by #267

PR #210, opened 2017-06-22
WIP: TRELLO#152 - make sure container runs after docker restart

vmindru:
WIP: TRELLO#152 - make sure container run's after docker restart

### DIFF
--- a/integration-tests/features/steps/common.py
+++ b/integration-tests/features/steps/common.py
@@ -3,14 +3,15 @@
 All steps in this file should only assume access to the LeApp CLI client,
 to allow testing of behaviour without the DBus service running.
 """
-from behave import given, when, then, step
-from hamcrest import assert_that, equal_to, is_not, not_none, greater_than
+from behave import given, when, then
+from hamcrest import assert_that, equal_to, is_not, not_none
 
 import shutil
 
 ##############################
 # Local VM management
 ##############################
+
 
 @given("the local virtual machines as {user}")
 @given("the local virtual machines")
@@ -61,6 +62,7 @@ def check_docker_is_installed(context):
 
 @when("{source_vm} is migrated to {target_vm} as a macrocontainer")
 @when("{source_vm} is migrated to {target_vm} as a macrocontainer and {migration_opt} is used for fs migration")
+@when("{source_vm} is migrated to {target_vm} as a macrocontainer named {container_name}")
 def migrate_vm_as_macrocontainer(context, source_vm, target_vm, migration_opt=None):
     """Uses leapp-tool.py to migrate the given source VM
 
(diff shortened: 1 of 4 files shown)


integration-tests/features/restart-container.feature
```diff
@@ -0,0 +1,11 @@
+Feature: container shold be running after restart
```
ncoghlan: I don't think this is really an independent feature, but rather a new scenario in the existing httpd-stateless migration tests.

integration-tests/features/restart-container.feature
```diff
@@ -0,0 +1,11 @@
...
+
+Scenario: Restart docker and check that the container is running 
+   Given the local virtual machines:
+         | name       | definition          | ensure_fresh |
+         | target     | centos7-target      | no           |
+         | app-source | centos6-source      | no           |
+   When app-source is migrated to target as a macrocontainer named container-app
+   And  docker is restarted
+   Then cotainer named container-app should be restarted and running
```
ncoghlan: Typo in step text: cotainer -> container

integration-tests/features/steps/common.py
```diff
@@ -71,9 +73,10 @@ def migrate_vm_as_macrocontainer(context, source_vm, target_vm, migration_opt=No
     context.migration_target = target_vm
     context.migration_migration_opt = migration_opt
     migrate_app = context.cli_helper.migrate_as_macrocontainer
-    output = migrate_app(source_vm, target_vm, migration_opt)
+    output = migrate_app(source_vm, target_vm, migration_opt)  # nopep8
```
ncoghlan: What's `pycodestyle` complaining about here? If it's something we think is noise, I prefer to deal with that by providing suitable config files that turn off particular warnings globally, rather than by selectively turning off all warnings for particular lines.

integration-tests/features/steps/restart-container.py
```diff
@@ -0,0 +1,13 @@
+"""Steps to test the "scan-ports" subcommand"""
```
ncoghlan: Docstring will need updating.

integration-tests/features/steps/common.py
vmindru: the warning is here about variable not beeing used, i also prefer global adjustments of errors, in this case, it has to be per line.

integration-tests/features/restart-container.feature
vmindru: I was actually wondering the same let me check again http-stateless

integration-tests/features/steps/common.py
ncoghlan: In that case, it probably makes more sense to just drop the variable assignment from this line (thus avoiding the warning), and reword the following comment a bit:

    # TODO: Assert specifics regarding command output returned by the migrate_app() call

vmindru: ack

vinzenz:
Replacing #274 with this PR

PR #213, opened 2017-06-23
cli: By default do not check host keys as a workaround

vinzenz:


ncoghlan:
Follow-up issue filed to fix this properly: https://github.com/leapp-to/prototype/issues/214

PR #217, opened 2017-06-23
remove unnecessary PortsException - do not raise if hostname is not i…

mfranczy:
…n scan.all_hosts, check only error value


vmindru:
+1

Issue #214, opened 2017-06-23
Security: check remote SSH host keys by default


ncoghlan:
We currently skip checking remote host keys by default because we don't have any way to handle adding them through either the GUI or the CLI.

That's just a short term workaround: the desired flow is to check them by default, and having a way of dealing with not yet known hosts in both the GUI and the CLI.

PR #224, opened 2017-06-26
CLI: Strict host key checking

vinzenz:
This PR is meant to address #214

### DIFF
--- a/src/leappto/cli.py
+++ b/src/leappto/cli.py
@@ -54,6 +54,7 @@ def _user_has_required_permissions():
             return False
     return True
 
+
 # Parsing CLI arguments
 def _add_identity_options(cli_cmd, context=''):
     if context:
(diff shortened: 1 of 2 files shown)


src/leappto/cli.py
```diff
@@ -66,6 +67,9 @@ def _add_identity_options(cli_cmd, context=''):
         cli_cmd.add_argument('--identity', default=None, help='Path to private SSH key')
         cli_cmd.add_argument('--ask-pass', '-k', action='store_true', help='Ask for SSH password')
         cli_cmd.add_argument('--user', '-u', default=None, help='Connect as this user')
+    cli_cmd.add_argument('--unsafe-disable-host-check', default=False, action='store_true',
```
vmindru: this is really annoying to type, just make it no host check, on the other side we don't really need this we can use SSH config like other software does.

PR #227, opened 2017-06-26
bugfix - do not resolve localhost to ip

mfranczy:


mfranczy:
@ncoghlan I had a wrong condition in port_scan code, let's see if this fix will also fix your problems with int-tests

PR #228, opened 2017-06-27
Ui show error message

GAZDOWN:


vinzenz:
Please rebase this against the latest master and squash it - The failure is caused by something that is fixed in master right now

cockpit/leapp.html
```diff
@@ -332,6 +337,11 @@ <h4 class="text-center">Application names already in use:</h4>
         */
         var new_leapp_call = function(command, options) {
             command = [LEAPP_CONF["tool-path"]].concat(command);
+            return new_generic_call(command, options);
+        }
+
+        var new_generic_call = function(command, options) {
+            //command = [LEAPP_CONF["tool-path"]].concat(command);
```
mfranczy: please remove this line, we shouldn't keep code which is not used, after that LGTM

PR #229, opened 2017-06-27
Update Python level CI dependencies

### LINKED ISSUE #220: Avoid pointlessly installing Ansible (et al) twice in CI
To ensure that we always get a consistent version of our dependencies, we're currently using `pipenv` to manage the Python virtual environment in our CI setup.

Unfortunately, this isn't really the ideal configuration, since it means that we end up installing a bunch of dependencies (including ansible itself) twice: once inside the virtual environment, and once outside it. And because the system level packages don't populate the Python wheel cache, the second installation ends up being quite slow.

The original concept below turned out to be based on a flawed assumption (since we're using the Python 3.5 SCL to run the test suite, we don't have access to the system site packages anyway), but that turned out not to matter: the duplicated components were ones the test suite only accesses via their CLI anyway, so we were able to just drop them from `Pipfile` and make it a requirement that they be set up externally.

---
Original concept:

Since we tightly control the CI execution environment, it would make sense to use the `--system-site-packages` option to eliminate the duplication, but that isn't currently supported in `pipenv` (for good reasons): https://github.com/kennethreitz/pipenv/issues/419

So while I still think `pipenv` is useful to create a completely self-contained local development environment, I'm starting to think that for CI purposes we're likely to be better off doing something like the following:

- Update `Pipfile` to pin the versions for components that we're expecting to get from the EPEL/RHEL/CentOS repos
- Update the CI Ansible roles to skip calling `pipenv install`, and instead call `pipenv lock --requirements` (alternatively: we could stop checking in `Pipfile.lock` and instead check in `integration-tests/requirements.txt`. If we went that way, we wouldn't need `pipenv` in the CI environment at all, and using it locally would also become entirely optional)
- Change the CI Ansible roles to set up the venv natively with `python -m venv --system-site-packages` rather than `pipenv`

ncoghlan:
- don't use `pipenv` in CI
- make Ansible solely a system level dependency
- explicitly make pipenv and pew optional
- expect pipsi to be installed externally
- document how to handle lock file updates
- document how to configure a test environment without pipenv
- regenerate lock files

Fixes #220


ncoghlan:
See #220 for more context on this draft PR: https://github.com/leapp-to/prototype/issues/220#issuecomment-311290346

ncoghlan:
The CI run actually failed - the changes to `bootstrap.sh` must have broken the return code propagation from behave.

integration-tests/features/environment.py
```diff
@@ -47,6 +47,7 @@ def _skip_test_group(context, test_group):
     return skip_group
 
 def before_all(context):
+    1/0 # Force behave invocation failure to check CI error reporting
```
vinzenz: um what? How can this succeed then?

ncoghlan: An eariler run was reported as green in GitHub, even though the tests had actually failed. So this is here temporarily to check that `behave` returning a non-zero exit status actually leads to the CI being reported as failing.

centos-ci/bootstrap.sh
```diff
@@ -51,5 +51,5 @@ export DISPLAY=:19
 
 
 echo "$(print_date): running integration tests"
-time scl enable rh-python35 sclo-vagrant1 -- sh -xc "cd ../.. && pipenv shell 'cd integration-tests && behave --no-color --junit; exit \\\$?'"
+time scl enable rh-python35 sclo-vagrant1 -- sh -xc 'cd ../../integration-tests && behave --no-color --junit; exit \\\$?'
```
shaded-enmity: Since there's no nested expansion now `exit $?` should be the way to go.

ncoghlan: That's what I thought too, but when I had that, the version with a forced divide-by-zero error still reported `exit 0`

shaded-enmity: Yup, seeing your last commit there must be something funny going on since:

```bash
$ sh -xc 'true ; exit $?'
+ true
+ echo 0
0
$ sh -xc 'false ; echo $?'                                                                                         
+ false
+ echo 1
1
```

I suspect that SCL is doing some black magic under the hood:
```bash
$ scl enable rh-python35 -- sh -xc 'false ; exit $?'
+ false
+ exit 0
$ scl enable rh-python35 -- sh -xc 'false ; exit \$?'
+ false
+ exit 1
```
Weird stuff :)

ncoghlan: Aye, if it's SCL related, that would explain why I couldn't reproduce it locally. I've tried to force another failure based on your experiment above (skipping the full test run this time), so if that fails as expected, I'll be able to remove the forced failure and this will be ready to go.

ncoghlan:
Assuming the next CI run passes, I'd suggest merging this via Squash & Merge, rather than just "Merge" :)

PR #230, opened 2017-06-27
Remove the fix_upstart function, turns out that it was not needed

pcahyna:
and actually there has been no service named mysqld, only mysql55-mysqld, so
the mysqld part never did anything.


vinzenz:
Jenkins: ok to test

PR #235, opened 2017-06-27
CLI: Added support for exclude paths

GAZDOWN:
I've added --exclude-path option for migrate-machine command. I am not sure, whether we should keep the default list of excluded paths as it was before: 
  
  '/dev/*', '/proc/*', '/sys/*', '/tmp/*', '/run/*', '/mnt/*', '/media/*', '/lost+found/*'

Or rather go this way:

  '/dev/*', '/proc/*', '/sys/*', '/tmp/*', '/run/*',  '/lost+found/*'

Because there might be a situation that user will have some service data in /mnt or /media:

P.S. Virt-tar-out exclusion was not implemented since we are planning to drop it anyways.

### DIFF
--- a/src/leappto/cli.py
+++ b/src/leappto/cli.py
@@ -118,7 +118,7 @@ def _make_argument_parser():
     def _path_spec(arg):
         path = os.path.normpath(arg)
         if not os.path.isabs(path):
-            raise TypeError("Path '{}' is not absoulute or valid.".format(str(path)))
+            raise TypeError("Path '{}' is not absoulute or valid.".format(str(arg)))
 
         return path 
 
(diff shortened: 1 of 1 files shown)


vmindru:
i go for excluding "'/dev/', '/proc/', '/sys/', '/tmp/', '/run/', '/mnt/', '/media/', '/lost+found/'"

vmindru:
--exclude-path should overwrite completely the default list to give full freedom e.g if one want's to exclude --/usr/ use will have to pass --excude-path  "'/dev/', '/proc/', '/sys/', '/tmp/', '/run/', '/mnt/', '/media/', '/lost+found/,'/usr/''"

src/leappto/cli.py
```diff
@@ -118,7 +118,7 @@ def _port_spec(arg):
     def _path_spec(arg):
         path = os.path.normpath(arg)
         if not os.path.isabs(path):
-            raise TypeError("Path '{}' is not absoulute or valid.".format(str(path)))
+            raise TypeError("Path '{}' is not absoulute or valid.".format(str(arg)))
```
vmindru: ValueError

PR #237, opened 2017-06-27
add missing deps to leapp.spec 

vmindru:
right now leapp won't run on a fresh install, it misses docker during setup and python  libs

### DIFF
--- a/leapp.spec
+++ b/leapp.spec
@@ -46,16 +46,19 @@ Requires:   libvirt-python
 Requires:   nmap
 Requires:   sshpass
 Requires:   python-enum34
+Requires:   docker 
 %if 0%{?el7}
 Requires:   python
 Requires:   python-psutil
 Requires:   python-nmap
 Requires:   python-paramiko
+Requires:   python-setuptools
 %else
 Requires:   python2
 Requires:   python2-nmap
 Requires:   python2-paramiko
 Requires:   python2-psutil
+Requires:   python2-setuptools
 %endif
 
 %description -n python2-%{name}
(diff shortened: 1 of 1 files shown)


leapp.spec
```diff
@@ -46,16 +46,20 @@ Requires:   libvirt-python
...
 Requires:   sshpass
 Requires:   python-enum34
+Requires:   docker 
+Requires:   rsync
 %if 0%{?el7}
 Requires:   python
 Requires:   python-psutil
 Requires:   python-nmap
 Requires:   python-paramiko
+Requires:   python-setuptools
```
vinzenz: This is maximal a BuildRequires not a requires

leapp.spec
```diff
@@ -46,16 +46,20 @@ Requires:   libvirt-python
 Requires:   nmap
 Requires:   sshpass
 Requires:   python-enum34
+Requires:   docker 
+Requires:   rsync
```
vinzenz: We did agree on that we do not depend on docker or rsync and don't want to install it.

We simply require target to have them - It's not a real dependency for the tool - Since it even supports doing things remotely

leapp.spec
vmindru: it's not it fails at runtime

leapp.spec
vmindru: untill we have a proper failure message that docker has to be installed let's have it as required. regarding rsync i would prefer to have it required it's the only supported method to operate the product now.

shaded-enmity: We do depend on `rsync` for the CLI to work (might be one more reason for the P2P rsync use case, but there migth be another things to solve in that regard)

vmindru:
@vinzenz can we give this a second round?

PR #246, opened 2017-06-29
Remove references to LeApp-To in the docs (replace by LeApp).

pcahyna:
Leave only "LeApp-To Organization on Github" as the organization is indeed still called leapp-to.


integration-tests/features/steps/remote_authentication.py
```diff
@@ -10,7 +10,7 @@ def skip_unless_running_as_root(context):
         context.scenario.skip("This scenario can only be run as root")
 
 def _make_auth_test_command(context, target):
-    """Create a leapp-to command to test remote authentication"""
+    """Create a leapp command to test remote authentication"""
```
vinzenz: in this particular case it could be leapp-tool since the command line tool is called leapp-tool

pcahyna: done.

Issue #154, opened 2017-06-06
Proposal: modular, dataflow-based architecture for the LeApp CLI


pcahyna:
# Synopsis
It is proposed to create a more modular architecture for the LeApp CLI, using dataflow-based programming concepts and frameworks. Modules in this architecture are called Actors. This infrastructure is also intended to serve as the next generation of the preupgrade-assistant tool, whose current modular architecture is too restricted. Actors will specify what types of data they require and provide on their input and output ports respectively, and a dependency solver will produce a dependency graph to satisfy those requirements, i.e. connect output ports to appropriate input ports. We considered two frameworks for flow-based programming: WOW:-P and Selinon. The former has the advantage of being simpler and apparently better expressing the dependency information that we need. The advantage of Selinon on the other hand is that it is designed to cope with non-idempotent actors. The actual implementation of actors will need to be possible to be done in shell, with a Python wrapper to present the proper interface for the (Python-based) framework. It is furthermore proposed to execute those shell parts using Ansible (as Ansible modules) and let them communicate using JSON, as it is the standard for Ansible modules.
# Goals
-  create a modular architecture for LeApp, so that the information about various migration strategies and application-specific logic is not kept in a single source file but in independent "modules" or "plugins", which can eventually be maintained by different people, even by experts on a particular application who are not part of the core LeApp team.
-  share this infrastructure with preupgrade-assistant, to serve as a basis of a new generation of preupgrade-assistant. Current preupgrade-assistant has a modular architecture and lots of modules have been already written, but the infrastructure is limited. In particular, there are no dependencies between modules and no way to pass information between modules, so if several modules need the same information, the code which produces it can not be a module itself, as it has no way to pass this information to them.
# Terminology
As "module" is a very overloaded word, we will use the term "actors" for the components in the proposed framework. (This is a term borrowed from one of the infrastructures considered - WOW:-P - in similar infrastructures this concept may be also referred to as "tasks".) It should be still kept in mind that they correspond to modules in the current version of preupgrade-assistant. (We will need to reserve the word "module" for Ansible modules later.) Actors communicate using interfaces called "ports", they accept inputs on input ports and provide outputs on output ports. The list of input and output port are a part of the definition of an actor.
# Requirements
-  Dependencies between actors must be possible to be expressed. (I.e. one must be able to express that one actor should be run before the other.)
-  Actors must be able to pass data to other actors.
[...]

PR #250, opened 2017-06-29
WIP - check-target command modularization

artmello:
Follow up from Proposal #154 and work done from @pcahyna , this PR is still WIP and will probably keep that way for a while.

Based on the idea of creating a more modular architecture for the Leapp CLI using dataflow based programming, this PR changes check-target command. Using WOW:P, framework for flow-based programming, to enable check-target to run shell scripts tests as actors, respecting the dependency between them, as defined in a provided YAML file.

Architecture proposed here is still a POC, so it can and probably will change a lot based on feedback and as requests shows up. Idea of proposing it against master earlier is to keep track of comments and any provided feedback.


vinzenz:
@artmello Please consider checking for the RPM dependencies in the leapp.spec file for all the python dependencies - Otherwise this will not work because we need to have this installed by RPM without pip usage

artmello:
wowp module it is not packaged yet (for Fedora at least). FWIW here is a copr with a rpm for it: https://copr.fedorainfracloud.org/coprs/amello/python-wowp/

PR #251, opened 2017-06-29
Prevent fsck from running.

pcahyna:
Sometimes I've encountered that fsck wants to run during container startup. This of course does not work and fsck asks interactively what to do, blocking the whole boot process.
This change prevents it by creating the /fastboot file in the target container which causes fsck to be
skipped. It is created outside the container and r/o mounted there to prevent
the container from removing it. At the same time create also /.nolvm, which
prevents LVM configuration.
It would be preferable to get rid of the whole /etc/rc.sysinit script, whose majority does not make sense in a container environment, but this I found too invasive.

### DIFF
--- a/src/leappto/cli.py
+++ b/src/leappto/cli.py
@@ -592,14 +592,18 @@ def main():
                 'rm -rf {1}/{0}'.format(container_name, storage_dir)
             )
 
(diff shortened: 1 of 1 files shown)


GAZDOWN:
Jenkins: ok to test

PR #253, opened 2017-06-30
Test Leapp RPM in CI

### LINKED ISSUE #245: Test the `leapp` RPM in CI
Currently the CI tests use `pipsi` and symlinks to run both the Cockpit plugin and `leapp-tool` itself directly from the git clone.

We want to change that so it tests the RPM-installed version in CI (while still supporting running the tests directly against the git clone locally), so this is my proposal for how to achieve that:

- [x] Add a `LEAPP_TEST_RPM` environment variable check to the `before_all` hook in the integration tests and note the result on the test context
- [x] Skip the `pipsi` invocation in the setup if that environment variable is set
- [x] Skip the creation of the user level Cockpit plugin symlinks if that environment variable is set
- [x] Update the integration testing Vagrant/Ansible configuration so it's possible to log in to those VMs as root using the default testing key (not just as vagrant)
- [x] Install the RPM after building it in `centos-ci/bootstrap.sh` (already implemented previously)
- [x] Run the remote-authentication feature tests without setting `LEAPP_TEST_RPM` in `centos-ci/bootstrap.sh` as a smoke test prior to running the full tests against the RPM
- [x] Set `LEAPP_TEST_RPM=1` when running the full tests in `centos-ci/bootstrap.sh`
- [x] Update the integration testing documentation accordingly

ncoghlan:
Closes #245

### DIFF
--- a/docs/source/integration-tests.rst
+++ b/docs/source/integration-tests.rst
@@ -216,6 +216,11 @@ not only between test scenarios, but also between different test *runs*, saving
 around 3-5 minutes of test execution time for each VM destruction and recreation
 cycle avoided.
 
+By default, VMs are still *halted* after a test run, even with ``ensure_fresh=no``.
+To further speed up test execution, set ``LEAPP_TEST_KEEPS_VMS=1`` in the
+environment: this will leave the VMs running, saving additional time when you
+expect to be using them again in a subsequent test run.
+
 For development and test debugging purposes, the ``ensure_fresh` setting can be
 changed to ``yes``. This means that instead of just re-running the Ansible
 provisioning playbook when a suitable VM instance already exists and halting
--- a/docs/source/integration-tests.rst
+++ b/docs/source/integration-tests.rst
@@ -1,8 +1,8 @@
 Integration tests
 =================
 
-Kernel modernization integration tests
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Legacy Application import integration tests
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 This directory contains integration tests that check the expected results
 of kernel modernization attempts given different starting scenarios.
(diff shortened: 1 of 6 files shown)


ncoghlan:
Flagged as WIP as at least docs updates are still needed

ncoghlan:
Removed the WIP marker as I've covered all the points identified in #245.

That doesn't mean there won't be further changes based on CI results and review comments, though :)

docs/source/integration-tests.rst
```diff
@@ -1,8 +1,8 @@
 Integration tests
 =================
 
-Kernel modernization integration tests
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Legacy Application import integration tests
```
vinzenz: Thanks I noticed that already the other day and forgot to post a patch later

ncoghlan:
Ah, just noticed that this is going to use the wrong `leapp-tool` for most of the testing, since the tests are currently still hardcoded to use the pipsi-installed version.

I need to move the calculation of the path to `leapp-tool` into the ClientHelper constructor.

integration-tests/features/steps/cockpit_demo.py
```diff
@@ -18,16 +18,21 @@ def create_demonstration_user(context):
...
-    assert_that(os.path.exists(user_plugin_dir), "User plugin not installed")
-    # Check the rest of the repo is linked where the plugin will find it
-    leapp_dir_link = pathlib.Path("/opt/leapp")
-    desired_leapp_target = context.BASE_REPO_DIR
-    _ensure_expected_link(leapp_dir_link, desired_leapp_target)
+    """Ensures demo plugin is available to demo user"""
+    if context.TESTING_RPM:
+        # Plugin should be installed globally
+        assert_that(os.path.exists("/usr/bin/leapp-tool"), "LeApp CLI not installed")
+        assert_that(os.path.exists("/var/lib/cockpit/leapp"), "LeApp plugin not installed")
```
vinzenz: /usr/share/cockpit/leapp

ncoghlan:
The approach I went with to resolve the tool path issue was to update ClientHelper to accept it as a constructor parameter, and the pipsi install helper to return the destination path it installs to.

ncoghlan:
While that passed, I'm not happy with the extra CI runtime introduced by the second behave call - I'm going to push an update that leaves the affected test VM running between the two invocations.

PR #261, opened 2017-07-05
Enable systemd based source machine migration

GAZDOWN:
Image is now based on centos/systemd in order to save some work with removing *.wanted folders content. 
To be able to run it, the container_manage_cgroup selinux boolean has to be enabled (P.S. even for upstart based distros, if we want to access cgroups)
"--cap-add", "SYS_ADMIN" parameter also added since this will be required on docker 1.13 (tested), but it works without it on current fedora docker 1.12

### DIFF
--- a/src/README.rst
+++ b/src/README.rst
@@ -23,3 +23,9 @@ Installation
 Installation is currently only supported as part of
 the LeApp Cockpit plugin demo or the LeApp integration
 tests.
+
+SELinux
+-------
+Please enable *container_manage_cgroup* boolean in order to manage cgroups within container.
+
+  semanage boolean -m -1 container_manage_cgroup
--- a/src/leappto/cli.py
+++ b/src/leappto/cli.py
@@ -643,17 +643,6 @@ def main():
             with open(os.path.join(container_dir, 'etc', 'init', 'rcS-emergency.conf'), 'w') as f:
                 f.write('exit 0\n')
 
-        def fix_systemd(self):
-            # systemd cleans /var/log/ and mariadb & httpd can't handle that, might be a systemd bug
-            fixer = 'bash -c "echo ! waiting ; ' + \
-                    'sleep 2 ; ' + \
-                    'mkdir -p /var/log/{httpd,mariadb} && ' + \
-                    'chown mysql:mysql /var/log/mariadb && ' + \
-                    'systemctl enable httpd mariadb ; ' + \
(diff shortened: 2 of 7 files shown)


src/leappto/cli.py
```diff
@@ -756,11 +745,24 @@ def print_migrate_info(text):
...
+                "--cap-add", "SYS_ADMIN",
+                # Implemented in default command
+                #"-v", "/sys/fs/cgroup:/sys/fs/cgroup:ro",
+                "--tmpfs", "/run", 
+                "--tmpfs", "/tmp",
+                "-e", "container=docker"
+                
+            ]
+
+            result = mc.start_container('centos/systemd',
```
ncoghlan: This should have a comment noting that `centos/systemd` doesn't currently offer version labels, so for now we're just relying on "CentOS 7 is the latest version".

src/leappto/cli.py
```diff
@@ -756,11 +745,24 @@ def print_migrate_info(text):
...
+                                        #'/usr/lib/systemd/systemd --system',
+                                        '/sbin/init',
```
ncoghlan: If the commented out command isn't needed any more, it should be removed entirely. However, it may still be worth commenting that `/sbin/init` kicks everything off correctly, so it's being called on purpose rather than as a holdover from CentOS 6.

src/README.rst
```diff
@@ -22,4 +22,10 @@ Installation
...
 the LeApp Cockpit plugin demo or the LeApp integration
-tests.
\ No newline at end of file
+tests.
+
+SELinux
+-------
+Please enable *container_manage_cgroup* boolean in order to manage cgroups within container.
+
+  semanage boolean -m -1 container_manage_cgroup
```
ncoghlan: Should we add that to the `target` Ansible role used in the test suite?

src/leappto/cli.py
```diff
@@ -756,8 +756,10 @@ def print_migrate_info(text):
                 
             ]
 
-            result = mc.start_container('centos/systemd',
-                                        #'/usr/lib/systemd/systemd --system',
+            # centos/systemd doesn't have any other tag than latest
+            # but it is derived from centos:7 image
+            result = mc.start_container('centos/systemd:latest',
+                                        # This is not relict from CentOS6, it will execute systemd properly
```
ncoghlan: Typo: "not relict" -> "not a relic"

PR #263, opened 2017-07-06
Set default target for destroy-container and check-target

GAZDOWN:
Set default target for:
  destroy-container
  check-target

And unified it with migrate-machine, so the target can be changed using -t option.

### DIFF
--- a/docs/source/leapp-tool.rst
+++ b/docs/source/leapp-tool.rst
@@ -53,23 +53,25 @@ migrate-machine
 destroy-containers
 ^^^^^^^^^^^^^^^^^^
     **usage:**
-        leapp-tool destroy-containers [-h] [--identity IDENTITY] [--ask-pass]
-                                      [--user USER]
-                                      target
+        leapp-tool destroy-container [-h] [-t TARGET] [--identity IDENTITY]
+                                    [--ask-pass] [--user USER]
+                                    container
+                                      
     
     positional arguments:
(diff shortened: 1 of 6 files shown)


src/leappto/cli.py
```diff
@@ -169,10 +169,10 @@ def _path_spec(arg):
     _add_identity_options(migrate_cmd, context='source')
     _add_identity_options(migrate_cmd, context='target')
 
-    check_target_cmd.add_argument('target', help='target VM name')
+    check_target_cmd.add_argument('-t', '--target', default='localhost', help='Target running docker')
```
ncoghlan: I'd suggest "Target container host" as the option description (while we currently require docker specifically, I assume we'll eventually support any OCI compliant container management daemon)

docs/source/leapp-tool.rst
```diff
@@ -53,23 +53,25 @@ migrate-machine
 destroy-containers
```
ncoghlan: Hmm, we should probably delete the `list-machines` docs above, and add `check-target` docs.

GAZDOWN: Will do, the list-machines command removal is actually still hanging on something (PR 147) and do not include docs either. I was actually thinking of removing that in the very same PR, but I would rather keep it separated. And about the check-target, I am planning to add it also in different PR. I correct the typos and continue with merging the changes.

PR #267, opened 2017-07-10
Completely remove libvirt support from leapp

vinzenz:


integration-tests/features/httpd-stateless.feature
```diff
@@ -1,14 +1,5 @@
 Feature: Redeploy stateless HTTP services
 
-Scenario: Migrate between remote hosts using virt-tar-out
-   Given the local virtual machines:
-         | name       | definition          | ensure_fresh |
-         | app-source | centos6-guest-httpd | no           |
-         | target     | centos7-target      | no           |
-    When app-source is migrated to target as a macrocontainer
```
ncoghlan: This simpler step definition can be reclaimed to refer to the rsync-based migration

src/leappto/cli.py
```diff
@@ -51,7 +51,7 @@
     _set_inheritable = None
 
 # Checking for required permissions
-_REQUIRED_GROUPS = ["vagrant", "libvirt"]
+_REQUIRED_GROUPS = ["vagrant"]
```
ncoghlan: We shouldn't need `vagrant` any more either, should we?

ncoghlan:
Test failures appear to be due to the switch to using IP addresses for target identification confusing the remote authentication tests for the `--ask-pass` and `--identity` options.

src/leappto/cli.py
vinzenz: @ncoghlan Yes I want to get rid of that too, I am just trying to verify it before ;-)

PR #274, opened 2017-07-24
TRELLO#152 - make sure container runs after docker restart

vinzenz:


vmindru:
10x a LOT Vinny for taking care of this :)

PR #275, opened 2017-07-25
Add instructions on how to set up a dev environment

bertinatto:


bertinatto:
Updated PR to move content to `docs/source/contributing.rst` instead.

docs/source/contributing.rst
```diff
@@ -115,3 +115,105 @@ Integration tests
 * It should run correctly when used in Firefox on RHEL/CentOS 7 (this
   requirement is not currently checked in the CI environment, but can be
   tested locally by running `behave --wip`)
+
+Setting up a development environment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This guide assumes that you are running Fedora. Make sure to translate the commands if you are running a different distribution.
```
mfranczy: +1 for updated doc. I would like to change only one thing, as a user I expect from doc to explicitly say what should I do (let's treat user as a junior), so in my opinion we should remove this ```Make sure to translate the commands if you are running a different distribution.``` and later we should add instruction for different distributions. That's only my opinion :)

vinzenz: @mfranczy Agreed - I think however in this particular case it's just good enough

PR #277, opened 2017-08-04
Used seccomp instad of sys_admin cap for systemd

GAZDOWN:


shaded-enmity:
Please document somewhere why this is necessary.

PR #280, opened 2017-08-25
Initial snactor integration

vinzenz:


src/leappto/actor_cli.py
```diff
@@ -0,0 +1,154 @@
...
+    else:
+        container_port = int(container_port)
+    return str(host_port), container_port
+
+
+def to_port_map(items):
+    port_map = {}
+    for source, target in items:
+        port_map[source] = port_map.get(source, []) + [target]
+    return {k: list(set(v)) for k, v in port_map.items()}
```
vmindru: key, value looks better imho

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,154 @@
...
+    if context:
+        cli_cmd.add_argument('--' + context + '-identity', default=None,
+                             help='Path to private SSH key for the ' + context + ' machine')
+        cli_cmd.add_argument('--' + context + '-user', default="root", help='Connect as this user to the ' + context)
+    else:
+        cli_cmd.add_argument('--identity', default=None, help='Path to private SSH key')
+        cli_cmd.add_argument('--user', '-u', default="root", help='Connect as this user')
+
+
+def _port_spec(arg):
```
vmindru: why not define this under +def make_argument_parser(): ?

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,154 @@
...
+def make_base_string(s):
+    return {"value": s}
+
+
+def make_base_bool(s):
+    return {"value": s}
+
+
+def start_agent_if_not_available():
+    if 'SSH_AUTH_SOCK' in os.environ:
```
vmindru: not sure we want to limit only to var value here, i would rather try to access that agent with current SSH_AUTH_SOCK and SSH_AGENT_PID other wise i think this check is not very efficent

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,154 @@
...
+        "excluded_tcp_port_list": {"tcp": map(lambda x: int(x[0]), arguments.excluded_tcp_ports)},
+        "excluded_paths": {"value": arguments.excluded_paths},
+        "start_container": make_base_bool(not arguments.disable_start),
+        "target_user_name": make_base_string(arguments.target_user),
+        "source_user_name": make_base_string(arguments.source_user),
+    }
+
+    if not registry.get_actor('migrate-machine').execute(data):
+        pprint(data)
+        exit(-1)
```
vmindru: don't we want some message on this exit?

vmindru: i mean more informative that just that preety print

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,154 @@
...
+
+def main():
+    ap = make_argument_parser()
+    argcomplete.autocomplete(ap)
+    parsed = ap.parse_args()
+
+    for identity in ('identity', 'target_identity', 'source_identity'):
+        if identity in parsed and getattr(parsed, identity):
+            start_agent_if_not_available()
+            subprocess.call(['ssh-add', getattr(parsed, identity)])
```
vmindru: don't we want a try: here?

vmindru:
also not sure if start_agent_if_not_available base_string and base_bool should be prefixed as private maybe?  I guess no blockers here just questions.

vmindru:
oh and btw i miss you folks :)

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,154 @@
...
+from argparse import ArgumentParser
+from pprint import pprint
+from subprocess import PIPE
+
+import argcomplete
+import snactor.executors
+
+from snactor import loader
+from snactor import registry
+
```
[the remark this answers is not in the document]
vinzenz: Well this was actually done automatically by pycharm and there is some reasoning behind this - First section is to import standard libraries, then 3rd party and at the end app/library modules / packages and each section is separated by new line

src/leappto/actor_cli.py
vinzenz: I don't know why it should be under it - There's nothing that requires it to be under it - Just because you can create them locally it doesn't have to be - And there's no reason to have it within the argument parser

src/leappto/actor_cli.py
vinzenz: well k, v is a very common naming scheme in python, especially in a dict comprehension context and it leaves little space for ambiguity.

src/leappto/actor_cli.py
vinzenz: well you might have a point here - However this is already a curtesy option to do that, and to be quite honest I don't even would want to support that in the end - I will drop the identity file support - I never really liked it anyway because we don't even know if it is password protected or not so rather let the user deal with the 'not being able to access the remote machine' scenario

src/leappto/actor_cli.py
vinzenz: obsolete with the removal of this

src/leappto/actor_cli.py
vinzenz: Since this is just WIP this has not been finished yet - That's what needs to be properly done

vinzenz:
Jenkins: ok to test

vinzenz:
Jenkins: ok to test

integration-tests/features/steps/fsfreeze.py
```diff
@@ -50,13 +50,12 @@ def test_migrate_fsfreeze(context, source, target, not_frozen=None):
 
         _replace_fsfreeze_bin(context, source)
```
vmindru: this does not look like a cool idea to directly alter a binary on the source server, though maybe i got something wrong

leapp.spec
```diff
@@ -9,8 +11,9 @@ URL:        https://github.com/leapp-to/leapp
 # git clone https://github.com/leapp-to/leapp
 # tito build --tgz --tag=%{version}
 Source0:    %{name}-%{version}.tar.gz
-BuildArch:  noarch
+BuildArch:  x86_64
 
+BuildRequires:   golang
```
vmindru: golang :) hipsterish ) :+1:

leapp.spec
```diff
@@ -270,7 +307,7 @@ popd
 - Changed error type and variable name (mgazdik@redhat.com)
 - --exclude-path overrides default list (mgazdik@redhat.com)
 - CLI: Added support for exclude paths (mgazdik@redhat.com)
-- Add excĺuded paths - initial commit (mgazdik@redhat.com)
+- Add exc??uded paths - initial commit (mgazdik@redhat.com)
```
vmindru: exc??uded looks wrong

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,265 @@
...
+    return path
+
+
+def _make_base_object(s):
+    return {"value": s}
+
+
+def _migrate_machine_arguments(parser):
+    migrate_cmd = parser.add_parser('migrate-machine', help='migrate source VM to a target container host')
```
vmindru: s/VM/machine/g

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,265 @@
...
+    migrate_cmd.add_argument("-p", "--print-port-map", default=False, help='List suggested port mapping on target host',
```
vmindru: s/list/print/g

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,265 @@
...
+    migrate_cmd.add_argument('--container-name', '-n', default=None,
+                             help='Name of new container created on target host')
+    migrate_cmd.add_argument(
+        '--force-create',
+        action='store_true',
+        help='force creation of new target container, even if one already exists'
+    )
+    migrate_cmd.add_argument('--disable-start', dest='disable_start', default=False,
+                             help='Migrated container will not be started immediately', action="store_true")
+    migrate_cmd.add_argument('--target-user', default="root", help='Connect as this user to the target via ssh')
```
vmindru: s/help='*"/help="target ssh user"/g

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,265 @@
...
+                             help='Default port mapping detected by leapp toll will be ignored', action="store_true")
+    migrate_cmd.add_argument('--container-name', '-n', default=None,
+                             help='Name of new container created on target host')
+    migrate_cmd.add_argument(
+        '--force-create',
+        action='store_true',
+        help='force creation of new target container, even if one already exists'
+    )
+    migrate_cmd.add_argument('--disable-start', dest='disable_start', default=False,
+                             help='Migrated container will not be started immediately', action="store_true")
```
vmindru: help="Do not start container after migration"

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,265 @@
...
+    migrate_cmd.add_argument('--target-user', default="root", help='Connect as this user to the target via ssh')
+    migrate_cmd.add_argument('--source-user', default="root", help='Connect as this user to the source via ssh')
+    migrate_cmd.add_argument('--debug', default=False, action="store_true", help='Turn on debug logging on stderr')
+
+
+def _migrate_machine(arguments):
+    default_excluded_paths = ['/dev/*', '/proc/*', '/sys/*', '/tmp/*', '/run/*', '/mnt/*', '/media/*', '/lost+found/*']
```
vmindru: i would also add /boot/ here

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,265 @@
...
+    registry.register_environment_variable('LEAPP_ACTOR_STDOUT_SOCK', name)
+    # This might be wanted to be a bit more dynamic but well it's good enough for now
+    registry.register_environment_variable('LEAPP_ACTOR_OUTPUT', '/usr/bin/actor-stdout')
+
+    env = os.environ.copy()
+    env["LEAPP_ACTOR_STDOUT_SOCK"] = name
+    p = subprocess.Popen(["actor-stdout", "server"], env=env)
+
+    yield
+    if p.poll():
```
vmindru: shouldn't  we have a try caluse here rather than that if?

https://docs.python.org/2/library/contextlib.html

At the point where the generator yields, the block nested in the with statement is executed. The generator is then resumed after the block is exited. If an unhandled exception occurs in the block, it is reraised inside the generator at the point where the yield occurred. Thus, you can use a try…except…finally statement to trap the error (if any), or ensure that some cleanup takes place. If an exception is trapped merely in order to log it or to perform some action (rather than to suppress it entirely), the generator must reraise that exception. Otherwise the generator context manager will indicate to the with statement that the exception has been handled, and execution will resume with the statement immediately following the with statement.

src/leappto/actor_cli.py
```diff
@@ -0,0 +1,265 @@
...
+    os.unlink(name)
+    os.rmdir(directory)
+
+
+def _check_target_arguments(parser):
+    check_target_cmd = parser.add_parser('check-target', help='check for claimed names on target container host')
+    check_target_cmd.add_argument('-t', '--target', default='localhost', help='Target container host')
+    check_target_cmd.add_argument("-s", "--status", default=False, help='Check for services status on target machine',
+                                  action="store_true")
+    check_target_cmd.add_argument('--user', default="root", help='Connect as this user to the target via ssh')
```
vmindru: similar to above  target ssh user

vmindru:
this is a really LONG PR, i would merge this and start a new one  =)

src/leappto/actor_cli.py
vinzenz: boot is usually not a tempfs or remote

src/leappto/actor_cli.py
vinzenz: good point

src/leappto/actor_cli.py
vmindru: it's not but why would you move this to a container?

vinzenz:
Jenkins: ok to test

PR #281, opened 2017-09-05
add 2 servers wordpress demo

jmikovic:


demo/vmdefs/available/centos6-wordpress/Vagrantfile
```diff
@@ -0,0 +1,32 @@
...
+                :mount_options => ["dmode=777", "fmode=666"]
+  
+            machine.vm.provider :libvirt do |libvirt|
+                libvirt.memory = 1024
+                libvirt.cpus = 1
+            end
+ 
+            machine.vm.provision "ansible" do |ansible|
+                ansible.limit = "all"
+                ansible.playbook = "playbook#{machine_id}.yml"
```
mfranczy: I guess you can have one playbook file, by host you can select what should be executed. Am I wrong?

demo/vmdefs/available/centos6-wordpress/Vagrantfile
```diff
@@ -0,0 +1,32 @@
+VAGRANTFILE_API_VERSION = "2"
+N = 2 
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+    (1..N).each do |machine_id|
+        config.vm.define "machine#{machine_id}" do |machine|
+            machine.vm.box = "centos/6"
+            machine.vm.hostname = "machine#{machine_id}"
```
mfranczy: ugly names machine1 and machine2 :P it's better to change names to something different like centos6-wordpress etc.

demo/vmdefs/available/centos6-wordpress/Vagrantfile
```diff
@@ -0,0 +1,32 @@
+VAGRANTFILE_API_VERSION = "2"
+N = 2
```
mfranczy: nah, is it possible to create hash with hostname and ip and then iterate over that? I don't like the idea how you calculate ip and set hostname.

From tips&trick site (https://www.vagrantup.com/docs/vagrantfile/tips.html)
```The Vagrantfile is a very flexible configuration format. Since it is just Ruby, there is a lot you can do with it. However, in that same vein, since it is Ruby, there are a lot of ways you can shoot yourself in the foot. When using some of the tips and tricks on this page, please take care to use them correctly.```

Comment (mfranczy) [reply demo/vmdefs/available/centos6-wordpress/Vagrantfile]: I see it comes from https://www.vagrantup.com/docs/provisioning/ansible.html#ansible-parallel-execution but still I think we should change that. What others think?

File: ansible_roles/wordpress/tasks/main.yml [hunk 4]
```diff
@@ -0,0 +1,32 @@
+- name: enable epel
```
vmindru: i would recomend to split this into a separate repos' role, it's easyer to maintain stuff in the future. Or you will endup enabling/disabling repos multiple times from multiple roles. this is just a recomandation feel free to ignore it.

ansible_roles/wordpress/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+  become_method: sudo
+
+- name: Copy WordPress config file
+  template: 
+    src: wp-config.php
+    dest: /etc/wordpress/
+
+- name: Copy conf for apache
+  template:
+    src: wordpress.tpl
```
vmindru: the convetion is to use j2 extension not tpl to make it clear this are jina2 templates

vmindru: also you seem to use that template as is without any vars inside it's faster to use copy in this case.

demo/vmdefs/available/centos6-wordpress/Vagrantfile
vmindru: yes 

---
- hosts: A
  tasks:  do 

- hosts: B
   tasks: do

vinzenz:
Jenkins: ok to test

PR #287, opened 2017-11-01
ADD: Docker image containing development environment

GAZDOWN:


dev-images/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM fedora:26
+
+ENV BUILD_SCRIPT        build_leapp_environment.sh
+ENV BUILD_SCRIPT_URL    http://leapp-to.github.io/files/$BUILD_SCRIPT
+ENV HOME                /root
+ENV LEAPP_HOME          $HOME/leapp
+
+
+RUN dnf update -y
+RUN dnf -y install 'dnf-command(copr)' sudo libvirt docker && \
```
shaded-enmity: Join these two together so that the `clean` operation cleans up the `update` operation as well.

dev-images/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN echo "echo \"Welcome in leapp development image. The development environment is prepared for you in /root/leapp\"" >> $HOME/.bashrc && \
+    echo "echo -e \"\n\$(ls -l ${LEAPP_HOME})\n\"" >> $HOME/.bashrc && \
+    echo "export SHELL=/bin/bash" >> $HOME/.bashrc
+
+
+# Move this to CMD
+## Run the build script before the container starts. This way the git data are
+## not distributed in the image and will be updated each time the container
+## is created.
+CMD /bin/bash -c "cd ${HOME} && curl -O ${BUILD_SCRIPT_URL} && [ ! -e ${HOME}/.installed ] && $SHELL ${BUILD_SCRIPT} 2>&1 | tee ${HOME}/leapp-setup.log && touch ${HOME}/.installed; exec $SHELL"
```
shaded-enmity: This doesn't look right - if I kill my development container (by accident) I lose all my work and more importantly I have no sane way of getting the data out of the container if I didn't set up bind mount(s) upfront accordingly.

The image should contain everything I need to get started developing & testing in a clean environment but all the repositories should be bind mounted from *the host* because it is a persistent state I care about.

GAZDOWN:
Jenkins: ok to test

PR #289, opened 2017-12-17
centos-ci/bootstrap.sh: fix epel release url

thrix:
Hardcoding epel version will work only temporarily. Idea from:
https://github.com/geerlingguy/ansible-role-repo-epel/blob/master/defaults/main.yml

Signed-off-by: Miroslav Vadkerti <mvadkert@redhat.com>


thrix:
In the future would be awesome to rewrite bootstrap.sh to ansible so we can use roles for tasks like this. This is just a quick fix to unblock the CI, which seems to be failing:

https://ci.centos.org/job/leapp-integration-tests/

PR #295, opened 2018-02-06
Use Ansible playbook for provisioning instead of shell script

vojtechsokol:


dev-images/vagrant/Vagrantfile
```diff
@@ -16,15 +17,18 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
-    config.vm.provision "shell", inline: "dnf -y update && dnf -y install docker libvirt"
+    config.vm.provision "shell", inline: "dnf -y upgrade && dnf -y install docker"
 
-    config.vm.provision "shell", inline: "cd #{HOME} && curl -O #{BUILD_SCRIPT_URL} && $SHELL #{BUILD_SCRIPT} 2>&1 | tee #{HOME}/leapp-setup.log", privileged: false
-    config.vm.provision "shell", inline: "systemctl enable docker libvirtd"
-    config.vm.provision "shell", inline: "systemctl start docker libvirtd"
+    config.vm.provision "shell", inline: "dnf install -y git ansible && "\
+        "cd #{HOME} && git clone #{GIT_URL} #{GIT_CLONE_DIR} && "\
+        "cd #{GIT_CLONE_DIR} && git checkout #{GIT_BRANCH} && "\
+        "cd ansible && ansible-playbook leapp_dev.yml 2>&1 | tee #{HOME}/leapp-setup.log"
```
GAZDOWN: IMO:
cd ansible && ansible-playbook leapp_dev.yml

would be better as:

ansible-playbook ansible/leapp_dev.yml

dev-images/vagrant/Vagrantfile
```diff
@@ -16,15 +17,18 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
         libvirt.management_network_address = '192.168.124.0/24'
     end
 
-    config.vm.provision "shell", inline: "dnf -y update && dnf -y install docker libvirt"
+    config.vm.provision "shell", inline: "dnf -y upgrade && dnf -y install docker"
 
-    config.vm.provision "shell", inline: "cd #{HOME} && curl -O #{BUILD_SCRIPT_URL} && $SHELL #{BUILD_SCRIPT} 2>&1 | tee #{HOME}/leapp-setup.log", privileged: false
-    config.vm.provision "shell", inline: "systemctl enable docker libvirtd"
-    config.vm.provision "shell", inline: "systemctl start docker libvirtd"
+    config.vm.provision "shell", inline: "dnf install -y git ansible && "\
```
GAZDOWN: yum shouldn't be used here, since this should be run with parameter: privileged: false. I suggest to move the dnf commands to the line above

vinzenz: @GAZDOWN what privileged false?

GAZDOWN: we do not want to run the complete ansible playbook with root privileges during the provision phase. In other words, we want to have everything available for vagrant user. When it comes to installation of binaries and dependencies, where you need root privileges, ansible will take care of this (become).

vojtechsokol: I was trying to be consistent with the Dockerfile, so the upgrade and docker installation is separated from the rest (git & ansible).

dev-images/vagrant/Vagrantfile
vojtechsokol: I was trying to be consistent with the Dockerfile again, so it is executed as `cd ansible && ansible-playbook leapp_dev.yml`. If the change is desired I will apply it also on the Dockerfile.

PR #299, opened 2018-04-09
Add docstring to leapp.actors

jzigmund:


jzigmund:
@vinzenz added 2 commits to the PR

docs/source/conf.py
```diff
@@ -171,24 +173,21 @@
...
     if name.startswith('with_meta_base_') or name == 'mro':
         return True
     if name == 'commands' and what == 'module':
         return True
     return skip
 
-
 def setup(app):
+
+    # prefix each section label with the name of the document it is in, followed by a colon
```
vinzenz: This comment should be removed now, that we do our own plugin instead

jzigmund:
@vinzenz fixed minor issues after your review, updated PR.

leapp/actors/__init__.py
```diff
@@ -13,73 +13,130 @@
...
             os.environ.pop('LEAPP_CURRENT_ACTOR', None)
 
     def process(self, *args, **kwargs):
+        """ Main processing method (in inherited actors, the function needs to be defined to be able process)"""
         raise NotImplementedError()
 
-    def produce(self, *args):
+    def produce(self, *models):
+        """
+        After running :py:func:`process`, it prepares processed data - results to next processing in workflow.
```
vinzenz: No this is not correct it is immediately when produce is called.
User rather this:

By calling produce model instances are store as messages. Those messages can be then consumed by other actors.

leapp/actors/__init__.py
```diff
@@ -13,73 +13,130 @@
...
+
+        :param models: what data models is going to be produced
```
vinzenz: :param models: Messages to be sent (those model types have to be specified in :py:attr:`~produces`
:type models: Variable number of :py:class:`leapp.models.Model` derived classes

leapp/actors/__init__.py
```diff
@@ -13,73 +13,130 @@
...
                         'message': {
                             'data': message_data,
                             'hash': message_hash
                         }
                     })
 
-    def consume(self, *types):
+    def consume(self, *models):
+        """
+        Load any given model data types to be able to process them.
```
vinzenz: Retrieve messages specified in the actors :py:attr:`~consumes` attribute and can be filter the message types by `models`

leapp/actors/__init__.py
```diff
@@ -13,73 +13,130 @@
...
+
+        :param models: what data models is going to be processed
+        :type models: tuple of derived classes from :py:class:`leapp.models.Model`
```
vinzenz: var arg not really tuple even though it looks like a tuple in the code

leapp/actors/__init__.py
```diff
@@ -13,73 +13,130 @@
...
                             'hash': message_hash
                         }
                     })
 
-    def consume(self, *types):
+    def consume(self, *models):
+        """
+        Load any given model data types to be able to process them.
+
+        :param models: what data models is going to be processed
```
vinzenz: Models to use as filter for the messages to return

PR #301, opened 2018-04-10
Fix language in devenv-install.md

jhornice:
I think that the following sentence should be reworded to be more descriptive:

snactor: This tool is designed to get quickly started with leapp actor development


vinzenz:
Thanks!

Issue #336, opened 2018-08-09
Error while trying to mock open builtin function during tests


artmello:
While writing tests for some actors that read data from files, we find a issue while trying to mock the builtin.open function. Here is what we have:

SampleActor:
```
from leapp.actors import Actor
from leapp.tags import SampleTag
from leapp.models import Output


class SampleFile(Actor):
    name = 'sample_file'
    description = 'Sample actor to read a file.'
    consumes = ()
    produces = (Output,)
    tags = (SampleTag,)

    def process(self):
        with open(self.get_file_path('sample.txt')) as f:
            self.produce(Output(data=f.read()))
```
OutputModel:
```
from leapp.models import Model, fields
from leapp.topics import SampleTopic


class Output(Model):
    topic = SampleTopic
    data = fields.String()
```

And the tests.py containing our single testcase:
```
from mock import patch, mock_open

from leapp.snactor.fixture import current_actor_context
from leapp.models import Output

def test_actor_execution(current_actor_context):
    sample_data = 'Burn the Witch!'
    with patch("__builtin__.open", mock_open(read_data=sample_data)) as mock_file:
        current_actor_context.run()
        assert current_actor_context.consume(Output)
        assert current_actor_context.consume(Output)[0].data == sample_data
```

This is our files structure:
```
.
├── actor.py
├── files
│   └── sample.txt
└── tests
    └── tests.py

2 directories, 3 files
```

When we try to execute the test running py.test, we got this:
```
============================= test session starts =============================
platform linux2 -- Python 2.7.15, pytest-3.6.4, py-1.5.4, pluggy-0.7.1
rootdir: /home/vagrant/mock_issue/actors/sampleactor/tests, inifile:
plugins: ordering-0.5, leapp-0.3
collected 1 item                                                              

tests.py F                                                              [100%]

================================== FAILURES ===================================
____________________________ test_actor_execution _____________________________

>   current_actor_context.run()

tests.py:46: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/home/vagrant/devel/leapp-dev/lib/python2.7/site-packages/leapp/snactor/fixture.py:104: in run
    self._actor(messaging=self._messaging).run()
[...]

PR #337, opened 2018-08-10
Guard the __builtin__.open function before mock

pirat89:
When the \_\_builtin\_\_.open function is mocked inside a unit test, test will fail because the environment cannot be initialised correctly. Problem happens when the os.fdopenfcuntion is called to create file object for the given stdin FD.

This is hackish workround, which stores the original function object during an import of the actor_definition module and ensure this one is be temporary restored when os.fdopen is called.

There should be definitely better solution, so feel free to close unmerged once you find/create better one.

Issue #336

### DIFF
--- a/leapp/repository/actor_definition.py
+++ b/leapp/repository/actor_definition.py
@@ -12,6 +12,13 @@ from leapp.exceptions import ActorInspectionFailedError, MultipleActorsError, Un
 from leapp.repository import DefinitionKind
 from leapp.repository.loader import library_loader
 
+# THE MOST DIRTY HACK
+# In case the open function is mocked, e.g. inside a unit test, the run of an
+# actor files on most systems when os.fdopen is called. Store the original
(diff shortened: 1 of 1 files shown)


pirat89:
Provided solution is compatible with Python2 only. As I hope this solution will not be used in the end, I will keep it as it is for now, otherwise it will require change to be compatible with Python 3.

vojtechsokol:
Could you please merge latest changes from master to be able to use the *copr_build* target from the Makefile?

vinzenz:
I guess we will be abandoning this one, I am not really in favour of this since it interferes with the functionality of the framework.

pirat89:
@vinzenz Yeah, I think we agreed we are not able to mock everything and the deal is we will run all tests in vagrant box - so I would go with it that people have to write the code they will be able to test it or they have to write it in the way they do not want to run them on the host system locally. As this was just proposal that helped to discuss this problem, I am closing it.

PR #344, opened 2018-09-03
Update leapp.spec

jmazanek:
Added the dependency to ethtool. It is needed for netifacescanner actor as well as future extensions to it


vinzenz:
https://github.com/leapp-to/leapp-actors/blob/master/leapp-repository.spec should be updated not this one.
Please put the dependency there not on the framework, thanks.

PR #350, opened 2018-09-20
Add manpages for leapp and snactor executables

vojtechsokol:
Manpages are generated by help2man tool by a Makefile target `manpages`.
What is missing are inculde files (`man/{leapp,snactor}.h2m`) to be used by help2man tools to make the output look less stubble.

### DIFF
--- a/man/leapp.1
+++ b/man/leapp.1
@@ -1,0 +1,34 @@
+.TH LEAPP "1" "30/09/2018" "leapp 0.3" "User Commands"
+.SH NAME
+leapp - command-line interface to for App & OS Modernization Framework Leapp
+.SH SYNOPSIS
+leapp [-h|--help] [--version] [--debug] <command>
+.SH DESCRIPTION
+The leapp tool is an end-user application designed to run specific workflows.
+.SH OPTIONS
+-h, --help
+    show this help message and exit
+
+--version
+    show program's version number and exit
+
+--debug
+    Enables debug logging
+.SH "LEAPP COMMANDS"
+upgrade
+    Upgrades the current system to the next available major version.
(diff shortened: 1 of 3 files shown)


vojtechsokol:
Manpages are no loner generated by help2man tool (and Makefile target `manpages` is gone) as it was inflexible and hard to customize.

man/snactor.1
```diff
@@ -0,0 +1,68 @@
...
+discover
+    Discovers all available entities in the current repository
+
+new-actor
+    Creates a new actor
+
+new-model
+    Creates a new model
+
+new-project
```
vinzenz: please remove this, since it's deprecated it makes no sense in keeping it here

man/leapp.1
```diff
@@ -0,0 +1,34 @@
...
+leapp [-h|--help] [--version] [--debug] <command>
+.SH DESCRIPTION
+The leapp tool is an end-user application designed to run specific workflows.
+.SH OPTIONS
+-h, --help
+    show this help message and exit
+
+--version
+    show program's version number and exit
+
```
vinzenz: I am missing the --whitelist-experimental option to be documented

vojtechsokol: Well thinking of it there is more that is missing, the manpage from this PR olny covers the snactor command, common options and a list snactor commands. Command specific options are missing, e.g. `--whitelist-experimental` for `run` command, `--json` for `discover` etc.

The question is how to deal with it - either list these options in the `SNACTOR COMMANDS` section of man or use the approach similar to what `git` has, i.e. `man snactor` would print the common options and list of subcommands and then there would be `man snactor <subcommand>` manpage for each subcommand of the snactor tool.

PR #361, opened 2018-10-24
WIP: Remove allow_null and missing from models fields

pirat89:
We do not see use-case to be able to distinguish whether value of model's attribute is None or was not be set at all. The original solution was confusing and there was not any actor writer that would expect such non-python logic. To make everything clear, remove the "missing" function and allow_null and simplify the logic just to use the "required" value for fields.

So the current logic:
  - If required is set to True for a field, expect that field cannot
    contains the None value.
  - If required is set to False (default) for a field, the field can
    contains the None value.

### DIFF
--- a/leapp/models/__init__.py
+++ b/leapp/models/__init__.py
@@ -133,7 +133,7 @@ class ErrorModel(Model):
 
     message = fields.String(required=True)
     severity = fields.StringEnum(required=True, choices=ErrorSeverity.ALLOWED_VALUES, default=ErrorSeverity.ERROR)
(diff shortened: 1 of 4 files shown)


pirat89:
@vinzenz The solution is not complete. Tests are not modified and I guess documentation should be modified as well in case there is any inside this repository which explains use of `allow_null` and `required`.

pirat89:
Tests are obviously still broken - or the code :) - anyway, another work is postponed to the next team discussion about this change.

pirat89:
Closing the PR as the problem will be resolved in different way.

PR #363, opened 2018-10-25
Set & use the LEAPP_DEBUG env properly

pirat89:
tl;dr;
- Do not override the LEAPP_DEBUG value when the --debug option is
  not used and LEAPP_DEBUG is already set.
- Set a logging level to DEBUG only when LEAPP_DEBUG == 1, otherwise
  use the INFO level.

Previously, the LEAPP tool ignored LEAPP_DEBUG set from the
commandline completely. So the value has been set only based on
use of the --debug option. Additionaly, the logging level has been
set to DEBUG in case the LEAPP_DEBUG has not been set.

### DIFF
--- a/leapp/cli/__init__.py
+++ b/leapp/cli/__init__.py
@@ -9,7 +9,7 @@ from leapp import VERSION
 @command('')
 @command_opt('debug', is_flag=True, help='Enable debug logging', inherit=True)
 def cli(args):
-    os.environ['LEAPP_DEBUG'] = '1' if args.debug else '0'
+    os.environ['LEAPP_DEBUG'] = '1' if args.debug else os.environ.get('LEAPP_DEBUG', '0')
 
 
 def main():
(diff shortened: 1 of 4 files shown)


leapp/cli/__init__.py
```diff
@@ -9,7 +9,7 @@
 @command('')
 @command_opt('debug', is_flag=True, help='Enable debug logging', inherit=True)
 def cli(args):
-    os.environ['LEAPP_DEBUG'] = '1' if args.debug else '0'
+    os.environ['LEAPP_DEBUG'] = '1' if args.debug else os.environ.get('LEAPP_DEBUG', 0)
```
vinzenz: '0' not 0

leapp/snactor/__init__.py
```diff
@@ -59,7 +59,7 @@ def cli(args):
         config_file_path = '/etc/leapp/leapp.conf'
 
     os.environ['LEAPP_CONFIG'] = config_file_path
-    os.environ['LEAPP_DEBUG'] = '1' if args.debug else '0'
+    os.environ['LEAPP_DEBUG'] = '1' if args.debug else os.environ.get('LEAPP_DEBUG', 0)
```
vinzenz: '0' not 0 as default

vinzenz:
CI failures are unrelated - Ignoring them therefore

PR #364, opened 2018-11-02
Initial localization support

vinzenz:
This commit introduces the initial changes necessary for allowing,
our code and actors to be localizable with gettext.

Magic builtins `_()` and `P_()` have been introduced with inspiration
taken from the DNF implementation to solve this.

Usage examples:

* Localization without need of pluralization:
        ```
         localized_string = _('localized')
        ```
* Localization with the need of pluralization:
        ```
        localized_string = P_('%d item', '%d items', 2)
        ```


vojtechsokol:
Could you please merge latest changes from master to be able to use the *copr_build* target from the Makefile?

PR #366, opened 2018-11-05
upgrade: Auto reboot functionality

vinzenz:
Previously reboots were always executed automatically, however this
behaviour isn't always wanted. Users and testers might need to do
things before the actual reboot.

This patch introduces a --reboot flag to the leapp upgrade
command line tool which will instruct the framework to perform
the reboot when it is requested. When the flag is not present
the upgrade will not performed and just a message will be presented
to the user asking them to reboot.

### DIFF
--- a/leapp/dialogs/__init__.py
+++ b/leapp/dialogs/__init__.py
@@ -19,3 +19,12 @@ class RawMessageDialog(Dialog):
 
     def __init__(self, message):
         super(RawMessageDialog, self).__init__(scope=None, reason=message)
+
+    def request_answers(self, store, renderer):
+        """
+        :param store: AnswerStore instance
+        :param renderer: Target renderer instance
+        :return: Dictionary with answers once retrieved
+        """
+        renderer.render(self)
+        return {}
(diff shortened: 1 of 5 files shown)


leapp/dialogs/__init__.py
```diff
@@ -19,3 +19,12 @@ class RawMessageDialog(Dialog):
 
     def __init__(self, message):
         super(RawMessageDialog, self).__init__(scope=None, reason=message)
+
+    def request_answers(self, store, renderer):
```
Rezney: This function has a same name, signature and a docstring as one in "leapp/dialogs/dialog.py". Would not it be better to differentiate it? Or in case this is somehow on purpose to mention why it is so?

Rezney: Ahh, ok this is part of functionality...

Rezney:
LGTM. Tested and it is working fine.

tcerna:
It stops before first reboot and admin can work with the machine. Good job! Thank you.

PR #367, opened 2018-11-05
upgrade: Remember commandline flags when resuming

vinzenz:
Previously when running leapp upgrade flags that have been passed
during the initial run, were not restored when trying to resume.
As a result of that, using --whitelist-experimental or --debug
on the command line were not respected within the initram phase
any more because those flags were forgotten.
This patch introduces a storing of those flags and passing them
on to the other phases with the same execution context (session).


vinzenz:
Verified and works

Issue #374, opened 2018-11-08
Problem with reusing model definition


artmello:
All the following were executed insider a RHEL 7 VM using python2-leapp-0.3-1.devel.20181107.git.232.noarch

We have the following scenario: One actor is suppose to scan and generate a list of items and a second actor is suppose to receive such list and filter the relevant entries. Since the elements on both lists will be the same, the model that defines each item is not duplicated.

Here are the models:

models/all.py
```python
from leapp.models import Model, fields
from leapp.topics import SystemInfoTopic

class Item(Model):
    topic = SystemInfoTopic
    data = fields.Integer()


class All(Model):
    topic = SystemInfoTopic
    items = fields.List(fields.Model(Item), required=True, default=[])
```
models/filtered.py
```python
from leapp.models import Item, Model, fields
from leapp.topics import SystemInfoTopic


class Filtered(Model):
    topic = SystemInfoTopic
    items = fields.List(fields.Model(Item), required=True, default=[])
```
and actors

actors/scanner/actor.py
```python
from leapp.actors import Actor
from leapp.models import All, Item
from leapp.tags import ScanTag


class Scanner(Actor):
    name = 'scanner'
    description = 'No description has been provided for the scanner actor.'
    consumes = ()
    produces = (All,)
    tags = (ScanTag)

    def process(self):
        result = All()
        for i in range(10):
            result.items.append(Item(data=i))
        self.produce(result)
```
actors/filter/actor.py
```python
from leapp.actors import Actor
from leapp.models import Filtered, All, Item
from leapp.tags import ScanTag

class Filter(Actor):
    name = 'filter'
    description = 'No description has been provided for the filter actor.'
    consumes = (All,)
    produces = (Filtered,)
    tags = (ScanTag)

    def process(self):
        result = Filtered()
        all_items = next(self.consume(All), All())
        for i in all_items.items:
            if i.data % 2 == 0:
                result.items.append(i)

        self.produce(result)
```
When we execute both actors the output is the one expected:
```json
[
  {
    "stamp": "2018-11-08T19:19:12.936713Z", 
    "hostname": "localhost.localdomain", 
    "actor": "filter", 
    "topic": "system_info", 
    "context": "57b5c57e-ddb2-48e9-9072-dea2e86112e2", 
    "phase": "NON-WORKFLOW-EXECUTION", 
    "message": {
      "hash": "27d4c45f3807f25b7d550829cefe244ab73eb4056cce15a4cfee3f579adee05a", 
[...]

PR #376, opened 2018-11-09
models: Support model references in fields.Model

vinzenz:
Previously the Model field type, did not check whether or not it
received a reference as model to use, which resulted in some validation
error if the real model was passed to it. This PR fixes this problem
by trying to get the _resolved field from the model and defaults
to the actual model type that has been passed.

Fixes https://github.com/leapp-to/leapp/issues/374

Related-issue: https://github.com/leapp-to/leapp/issues/374


shaded-enmity:
Tested with c&p setup from #374 and LGTM @artmello can you confirm?

artmello:
LGTM. Thanks for fixing it

PR #379, opened 2018-11-21
Introduce nullable field type

Rezney:
Currently we are using "missing" function as a placeholder
in cases where no value was specified in the particular field.
This might be confusing and can introduce issues when not
understood correctly. This patch is removing concept of the
missing function together with "allow_null" and "required"
parameters and switch to using a new nullable field type.
In this case every field which is not nullable has to be
specified. If we want to allow a field to be None we need to
wrap it around the nullable field.

E.g. fields.nullable(fields.String())

### DIFF
--- a/docs/source/messaging.md
+++ b/docs/source/messaging.md
@@ -18,7 +18,6 @@ Assign the SystemInfoTopic to the new model and add two fields:
 * The `name` field represents the hostname.
 * The `ips` field contains a list of strings with IPv4 or IPv6 addresses.
 
-Both fields are required in this scenario.
 
 ```python
 from leapp.models import Model, fields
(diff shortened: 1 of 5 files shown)


Rezney:
Thanks @vinzenz for helping with this...

Rezney:
There are still documentation changes missing and probably the tests will need some more love but this is mainly for discussion.

shaded-enmity:
From a cursory look LGTM, however I'd rename `nullable` to `Nullable` to stay consistent with the capitalization of other classes. 

I'm gonna give this deeper look in the afternoon.

Rezney:
The suggestions were implemented. The docs should be ok. The only thing which may need further polishing are tests. It then depends on how quick (should this solution be accepted) we want to implement it, whether to open a proper follow-up on testing or wait and work on it here.

Rezney:
Was testing further and it turned out I omitted one fail:

```python
  File "/usr/lib/python2.7/site-packages/leapp/models/fields/__init__.py", line 379, in _validate_model_value
    self._elem_type._validate_model_value(entry, name='{}[{}]'.format(name, idx))
  File "/usr/lib/python2.7/site-packages/leapp/models/fields/__init__.py", line 432, in _validate_model_value
    resolved_model.__name__, name, type(value)))
ModelViolationError: Expected an instance of RPM for the items[0] attribute but got <class 'installedrpm.RPM'>
```
which seems to be related to the fact that at this part the model is "_resolved". Fixed.

PR #380, opened 2018-11-21
Avoid showing multiple tracebacks to users

Rezney:
At the moment when an actor fails, two tracebacks are
thrown to the stderr. One for an actual failure in the actor
and the other (the last one) about the actor failure itself.
As the last traceback is not useful at all we shall keep only
the first one.

### DIFF
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -10,7 +10,7 @@ from leapp.logger import configure_logger
 from leapp.repository.scan import find_and_scan_repositories
 from leapp.utils.audit import Execution, get_connection, get_checkpoints
 from leapp.utils.clicmd import command, command_opt
-from leapp.utils.output import report_errors
+from leapp.utils.output import report_errors, beautify_actor_exception
 
 
 def load_repositories_from(name, repo_path, manager=None):
@@ -91,5 +91,7 @@ def upgrade(args):
             msg = 'No such Actor: {}'.format(actor_name)
             logger.error(msg)
             raise CommandError(msg)
-    workflow.run(context=context, skip_phases_until=skip_phases_until)
+    with beautify_actor_exception():
+        workflow.run(context=context, skip_phases_until=skip_phases_until)
+
     report_errors(workflow.errors)
--- a/leapp/snactor/commands/run.py
+++ b/leapp/snactor/commands/run.py
@@ -6,7 +6,7 @@ from leapp.utils.clicmd import command, command_opt, command_arg
 from leapp.utils.repository import requires_repository, find_repository_basedir
 from leapp.logger import configure_logger
 from leapp.messaging.inprocess import InProcessMessaging
-from leapp.utils.output import report_errors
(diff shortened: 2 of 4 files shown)


leapp/cli/upgrade/__init__.py
```diff
@@ -91,5 +91,14 @@ def upgrade(args):
             msg = 'No such Actor: {}'.format(actor_name)
             logger.error(msg)
             raise CommandError(msg)
-    workflow.run(context=context, skip_phases_until=skip_phases_until)
+    try:
+        workflow.run(context=context, skip_phases_until=skip_phases_until)
+    except LeappRuntimeError as e:
+        msg = '{} - Please check the exception above'.format(e.message)
```
shaded-enmity: I think the exception handler should be kept in one place (`leapp.utils`) and imported where needed.

vinzenz: Good idea, and that could be even done with some kind of connection manager
```python
with exception_beautifier():
    workflow.run(context=context, skip_phases_until=skip_phases_until)
```

Rezney: Ack, thanks for the ideas...

shaded-enmity:
LGTM @Rezney - out of curiosity can you please paste here how the final message looks like?

Rezney:
The final message looks like:

```bash
(venv) [work@client traceback]$ snactor run HostnameScanner
2018-11-21 15:21:04.21  INFO     PID: 2247 leapp: Logging has been initialized
2018-11-21 15:21:04.74  INFO     PID: 2247 leapp.repository.traceback: A new repository 'traceback' is initialized at /tmp/fail_actor/traceback
Process Process-3:
  File "/usr/lib64/python2.7/multiprocessing/process.py", line 267, in _bootstrap
    self.run()
  File "/usr/lib64/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/tmp/fail_actor/tut/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 61, in _do_run
    target_actor(logger=logger, messaging=messaging).run(*args, **kwargs)
  File "/tmp/fail_actor/tut/lib/python2.7/site-packages/leapp/actors/__init__.py", line 121, in run
    self.process(*args)
  File "/tmp/fail_actor/traceback/actors/hostnamescanner/actor.py", line 15, in process
    os.unlink('bullshit')
OSError: [Errno 2] No such file or directory: 'nonexistent'

===================================================================================================
Actor hostname_scanner unexpectedly terminated with exit code: 1 - Please check the above traceback
===================================================================================================

(venv) [work@client traceback]$
```

Just changed the additional message to "Please check the above traceback".

If the patch was not tested yet I asked @jmikovic to take a look.

vinzenz:
@Rezney well, not always when the exit code is not 0, it was an exception so 'the above traceback' might be misleading.

Rezney:
@vinzenz fair enough...

Rezney:
Thanks @jmikovic for testing and catching the missing context manager import...

PR #381, opened 2018-11-23
Release 0.4.0

vinzenz:


packaging/leapp.spec
```diff
@@ -211,5 +211,8 @@ install -m 0644 -p man/snactor.1 %{buildroot}%{_mandir}/man1/
 %endif
 
 %changelog
+* Fri Nov 22 2018 Vinzenz Feenstra <evilissimo@gmail.com> - 0.4.0-1
+- First release
+
 * Mon Apr 16 2018 Vinzenz Feenstra <evilissimo@gmail.com> - 0.3-1
 - Initial rpm
```
pirat89: Do not modify changelog in upstream as changelog is supposed to be for downstream. For upstream is better to keep it separately - part of release log or changelog in separate file. Look at this: https://github.com/oamg/leapp-repository/blob/68b7596bd7459d6866eff338616ba5630edd3cd8/packaging/leapp-repository.spec#L82

PR #385, opened 2018-11-29
Timestamp fix

vojtechsokol:
Simpler fix for persistent TIMESTAMP (#384)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ MASTER_BRANCH=master
 # upstream solution). For upstream builds N_REL=1;
 N_REL=`_NR=$${PR:+0}; if test "$${_NR:-1}" == "1"; then _NR=$${MR:+0}; fi; git rev-parse --abbrev-ref HEAD | grep -q "^$(MASTER_BRANCH)$$" || _NR=0;  echo $${_NR:-1}`
 
-_TIMESTAMP=$${TIMESTAMP:-`date +%Y%m%d%H%MZ -u`}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ MASTER_BRANCH=master
 # upstream solution). For upstream builds N_REL=1;
 N_REL=`_NR=$${PR:+0}; if test "$${_NR:-1}" == "1"; then _NR=$${MR:+0}; fi; git rev-parse --abbrev-ref HEAD | grep -q "^$(MASTER_BRANCH)$$" || _NR=0;  echo $${_NR:-1}`
 
-_TIMESTAMP=$${TIMESTAMP:-`date +%Y%m%d%H%MZ -u`}
+TIMESTAMP=$(shell /bin/date "+%Y%m%d%H%MZ" -u)
```
pirat89: Great. I just suggest to use:
`TIMESTAMP=$${__TIMESTAMP:-$(shell /bin/date "+%Y%m%d%H%MZ" -u)}`

In case that somewhere we call $(MAKE), we should be able to keep TIMESTAMP persistant, so we should call it `__TIMESTAMP=$(TIMESTAMP) $(MAKE)` - I know it's not relevant for this one, but as we want to keep the solution unified, would be better to change here as well.

pirat89:
We found this is not working as every target is run as separate shell and the substitution is evaluated separately. so this is not working solution.

PR #387, opened 2018-12-05
Introduce a way of running external programs

shaded-enmity:
This function provides stable interface across Python2 and Python3.
Besides, the main difference between this and the various `subprocess`
related functions is that we can easily "tee" the output to stdout/stderr
while buffering it internally for auditing.

Note that the `_call` function should provide internal interface for Leapp and that we should provide additional "facade" on top of that function to simplify the interface (and perhaps mimic the `subprocess` interface).

You can easily test this function by calling the file directly:

```
$ python2 call.py bash -c 'echo 123 ; (>&2 echo err) ; exit 255'
123
err
{'exit_code': 255,
 'pid': 16412,
 'signal': 0,
 'stderr': u'err\n',
 'stdout': u'123\n'}
$ python3 call.py dnf update
Error: This command has to be run under the root user.
{'exit_code': 1,
 'pid': 13591,
 'signal': 0,
 'stderr': u'Error: This command has to be run under the root user.\n',
 'stdout': u''}
```

### DIFF
--- a/leapp/libraries/stdlib/call.py
+++ b/leapp/libraries/stdlib/call.py
@@ -1,0 +1,190 @@
+from __future__ import print_function
+
+import select
+import os
+import sys
+
+
+STDIN = 0
+STDOUT = 1
+STDERR = 2
+
+
+def _multiplex(ep, read_fds, callback_raw, callback_linebuffered,
+               encoding='utf-8', write=None, timeout=1, buffer_size=80):
+    # Register the file descriptors (stdout + stderr) with the epoll object
+    # so that we'll get notifications when data are ready to read
+    for fd in read_fds:
+        ep.register(fd, select.EPOLLIN | select.EPOLLPRI)
+
+    # Register a write file descritpr
+    if write:
+        ep.register(write[0], select.EPOLLOUT)
+
+    # Offset into the `write[1]` buffer where we should continue writing to stdin
+    offset = 0
+
+    # We need to keep track of which file descriptors have already been drained
+    # because when running under `pytest` it seems that all `epoll` events are
+    # received twice so using solely `ep.unregister(fd)` will not work
+    hupped = set()
+    # Total number of 'hupped' file descriptors we expect
+    num_expected = len(read_fds) + (1 if write else 0)
+    # Set up file-descriptor specific buffers where we'll buffer the output
+    buf = {fd: bytes() for fd in read_fds}
+    linebufs = {fd: '' for fd in read_fds}
+
+    while not ep.closed and len(hupped) != num_expected:
+        events = ep.poll(timeout)
+        for fd, event in events:
+            if event == select.EPOLLHUP:
+                hupped.add(fd)
+                ep.unregister(fd)
+            if event & (select.EPOLLIN | select.EPOLLPRI) != 0:
+                read = os.read(fd, buffer_size)
+                callback_raw(fd, read)
+                linebufs[fd] += read.decode(encoding)
+                while '\n' in linebufs[fd]:
+                    pre, post = linebufs[fd].split('\n', 1)
+                    linebufs[fd] = post
+                    callback_linebuffered(fd, pre)
+                buf[fd] += read
+            elif event == select.EPOLLOUT:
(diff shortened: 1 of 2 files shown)


leapp/libraries/common/call.py
```diff
@@ -0,0 +1,119 @@
+from __future__ import print_function
+
+import select
+import os
+import sys
+
+
+STDOUT = 1
```
pirat89: Wouldn't be better to use: `sys.stderr.fileno()` and `sys.stdout.fileno()`

pirat89:
@vinzenz I guess this one should be under `stdlib`, right? `common` is meant to be part of *leapp-repository*.

leapp/libraries/common/call.py
```diff
@@ -35,7 +39,7 @@ def multiplex_reads(ep, fds, callback_raw, callback_linebuffered, timeout=1, buf
                 callback_linebuffered(fd, pre)
             buf[fd] += read
 
-    # process leftoevers from line buffering
+    # Process leftoevers from line buffering
```
vinzenz: leftoevers => leftovers

Rezney:
+1 for adding the comments

leapp/libraries/common/call.py
```diff
@@ -0,0 +1,135 @@
...
+    pid = os.fork()
+    if pid > 0:
+        # Since pid > 0, we are in the parent process, so we have to close the write-end
+        # file descriptors
+        os.close(wstdout)
+        os.close(wstderr)
+
+        read = multiplex_reads(
+            ep,
+            [stdout, stderr],
```
vinzenz: I guess that it won't matter that much, but you never change this list, I think it's a good idea to have it a tuple, lists are just noice IMHO but well I will let this be up to you.

leapp/libraries/common/call.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # Process leftoevers from line buffering
+    for (fd, lb) in linebufs.items():
+        if lb:
+            callback_linebuffered(fd, lb)
+
+    return buf
+
+
+def _call(command, callback_raw=lambda fd, value: None, callback_linebuffered=lambda fd, value: None,
```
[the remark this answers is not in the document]
shaded-enmity: All should be private FTTB

leapp/libraries/common/call.py
```diff
@@ -0,0 +1,135 @@
...
+
+STDOUT = sys.stdout.fileno()
+STDERR = sys.stderr.fileno()
+
+
+def decode(data):
+    return data.decode('utf-8')
+
+
+def multiplex_reads(ep, fds, callback_raw, callback_linebuffered, timeout=1, buffer_size=72):
```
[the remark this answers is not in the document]
shaded-enmity: Changed size to 80 and added notice into docstring

shaded-enmity:
@vinzenz @Rezney @pirat89 Hopefully addressed all the comments. (besides the testing which i'm going to take care of now)

I also extended the code so that it is possible to pass `stdin` to the child process as well (either directly as a string, or a file descriptor).

shaded-enmity:
Ok folks, I think that we're good to go here.

As for the final facade I'd start with the following (the implementation part is pseudocode):
```python
def call(command, stdout=True, stderr=True, stdin=fd/str, split=True):
     Audit.add('COMMAND_START', command, bytes(stdin), user.uid, user.gid)
     def linebuffer_callback(fd, data):
         log(data)
     r = _call(command, linebuffer_callback=linebuffer_callback, stdin=stdin)
     Audit.add('COMMAND_FINISH', r['stdout'], r['stderr'], r['error_code'], r['pid'], r['signal'])
     if not stdout:
         del r['stdout']
     elif split:
         r['stdout'] = r['stdout'].splitlines()
     if not stderr:
         del r['stderr']
     elif split:
         r['stderr'] = r['stderr'].splitlines()
         
     return r
```
If either stdout/stderr are False, their output is not included in the return dictionary (just the error_code, pid and signal if both are off)

This should be compatible with the old interface, while providing new means of control for users.

PR #388, opened 2018-12-05
WIP: Introduction of leapp.libraries.stdlib library

radomil:
leapp.libraries.stdlib library created
 - represents a location for functions that otherwise would be/were defined multiple times across leapp actors. And at the same time incorporated functions are really useful for other actors.

Library leapp.libraries.stdlib now consists of:
 Function
    - call - supporting python2/python3 (six module)

### DIFF
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -1,0 +1,32 @@
+"""
+:py:mod:`leapp.libraries.stdlib`
+represents a location for functions that otherwise would be defined multiple times across leapp actors
+and at the same time, they are really useful for other actors.
+"""
+import six
+import subprocess
+import os
+
+
+def check_output(args, split=True):
+    """
+    Call an external program, capture and automatically utf-8 decode its ouput.
+    Then, supress output to stderr and redirect to /dev/null.
+
+    :param args: Command to execute
+    :type args: list
+    :param split: Split the output on newlines
+    :type split: bool
+    :return: stdout output, 'utf-8' decoded, split by lines if split=True
+    :rtype: unicode/str or [unicode/str] if split=True
+    """
+
+    r = None
+    with open(os.devnull, mode='w') as err:
+        if six.PY3:
+            r = subprocess.check_output(args, stderr=err, encoding='utf-8')
+        else:
+            r = subprocess.check_output(args, stderr=err).decode('utf-8')
+    if split:
+        return r.splitlines()
+    return r
--- a/tests/scripts/test_stdlib.py
+++ b/tests/scripts/test_stdlib.py
@@ -1,0 +1,21 @@
+from leapp.libraries.stdlib import check_output
+
+
+def test_check_single_line_output():
+    a_command = ['echo', 'This a single line test!']
+    assert check_output(a_command) == [u'This a single line test!']
+
+
+def test_check_single_line_output_no_split():
+    a_command = ['echo', 'This a single line No Split test!']
+    assert check_output(a_command, split=False) == u'This a single line No Split test!\n'
+
+
+def test_check_multiline_output():
+    a_command = ['echo', 'This a multi-\nline test!']
+    assert check_output(a_command) == [u'This a multi-', u'line test!']
+
+
+def test_check_multiline_output_no_split():
+    a_command = ['echo', 'This a multi-\nline No Split test!']
+    assert check_output(a_command, split=False) == u'This a multi-\nline No Split test!\n'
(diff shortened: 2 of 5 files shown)


leapp/libraries/stdlib/__init__.py
```diff
@@ -0,0 +1,25 @@
+"""
+:py:mod:`leapp.libraries.stdlib` represents a location for functions that otherwise would be defined multiple times across leapp actors
+ and at the same time are really useful for other actors.
+
+Example:
+    If the function call() is called from an actor
```
vinzenz: This example does not really contribute to anything. This is basic python functionality, I would not bother making an example.

leapp/libraries/stdlib/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+      from leapp.libraries.stdlib import call
+
+"""
+import six
+import subprocess
+import os
+
+def call(args, split=True):
+    ''' Call external processes with some additional sugar '''
```
vinzenz: Please document this function properly 'With some additional sugar' isn't really a  good documentation, plus I am missing parameter description:
```python
    """
    Calls an external program and captures it ouput. Additionally the output is automatically utf-8 decoded, and output to stderr is suppressed and redirected to /dev/null.

    :param args: Command to execute
    :type args: list
    :param split: Split the output on newlines?
    :type split: bool
    :return: stdout output, 'utf-8' decoded, split by lines if split=True
    :rtype: unicode/str or [unicode/str] if split=True
   """
```

Also I would call this function rather 'check_output' since this is what it is using and gives a bit of a background to people using this function on what this actually is supposed to be doing. 
The name `call` is misleading IMHO, this was fine for private libraries but stdlib functions need to be named better.

radomil:
@vinzenz , changes incorporated. Now, it consists of one commit only.

leapp/libraries/stdlib/__init__.py
```diff
@@ -0,0 +1,31 @@
+"""
+:py:mod:`leapp.libraries.stdlib` represents a location for functions that otherwise would be defined multiple times across leapp actors
```
bocekm: Keep the lines up to 120 chars. This one has 136.

leapp/libraries/stdlib/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+ and at the same time are really useful for other actors.
+"""
+import six
+import subprocess
+import os
+
+
+def check_output(args, split=True):
+    """
+    Calls an external program and captures its ouput. Additionally the output is automatically utf-8 decoded,
```
bocekm: _Calls_ -> _Call_
Use imperative mood/command style, based on [PEP-257](https://www.python.org/dev/peps/pep-0257/) referred by PEP-8:
> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

That reminds me to add PEP-257 to the contribution guidelines.

radomil: updated

leapp/libraries/stdlib/__init__.py
radomil: updated

leapp/libraries/stdlib/__init__.py
pirat89: I would extend the info, that we talked with @shaded-enmity about the extensions of the function, so check_output doesn't fit as well because it suggests it should be more like known `subprocess.check_output` function. However it is mentioned to do more things and `call` is generic name, which sounds like something what I will find first. So I would go with call even. Anyway, I can go even with the check_output function, as we can inside call the generic `call` function with expected params.

tests/scripts/test_stdlib.py
```diff
@@ -0,0 +1,7 @@
+from subprocess import check_output, CalledProcessError
+from leapp.libraries.stdlib import check_output as tested_check_output
+
+
+def test_check_output():
+    a_command = ['pwd']
+    assert check_output(a_command) == tested_check_output(a_command)
```
pirat89: That's not sufficient testing. Actually, it doesn't mean anything. You should check, you will get specific output, not the same output that provides original `check_output` function. This test doesn't tell you anything. Better would be to use even `['echo', '<string>']` and check you get '<string>'. In addition, the function has additional parameter, you don't test it. We expect that library functions will be well tested. So I expect more tests, that will show the correct functionality of the function.

radomil:
Based on discussion, code has been updated. Please, revisit your reviews @vinzenz @pirat89.

PR #398, opened 2018-12-07
Fix broken link to "Repository Directory Layout"

AloisMahdal:


bocekm:
I'm fixing this also under https://github.com/oamg/leapp/pull/394

PR #402, opened 2018-12-13
Update the unit testing tutorial

bocekm:


docs/source/unit-testing.md
```diff
@@ -27,41 +21,35 @@ actors\
...
 
 ### Naming conventions
 
 To have the tests found and carried out by
-[pytest framework](https://pytest.org), follow these rules:
-
-- All tests have to reside in the `test_*.py` or `*_test.py` files.
-- Test functions outside of the class have to be prefixed by `test_`.
-- Test methods with the `test_` prefix have to reside in the `Test` prefixed classes.
-- Tests should have unique filenames.
```
pirat89: Is it fixed? I think that there was reason for that because of our CI. (ping: @examon @vojtechsokol ??)

docs/source/unit-testing.md
```diff
@@ -27,41 +21,35 @@ actors\
...
+ - be prefixed by `test_`.
 
-See the pytest [documentation](https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code).
+See the [pytest documentation](https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code).
 
 ### Writing tests that execute the whole actor
 
 Now let's assume you would want to write a test that executes the actor.
-This is how your `test_mytest.py` from above could look like:
+This is how your `unit_test.py` from above could look like:
```
pirat89: just as the run of actor is not unit test, I would rather avoid use of the *unit* word here :)

bocekm: Initially I was thinking to mention here both `unit_test.py` and `component_test.py` to differentiate the two test levels as per https://github.com/oamg/leapp-repository/pull/14. But we can't seem to agree on the naming of the other test level - functional/component/end-to-end: https://github.com/oamg/leapp-repository/pull/14#pullrequestreview-183396803. Till the time we agree on that, I can revert back to `test_mytest.py` here.

docs/source/unit-testing.md
bocekm: To be more specific - the requirement of unique testfile name is actually a requirement of `pytest`. But that would apply only if we run all tests of all actors at once, like `pytest <leapp-repository/repos>`. But we call `pytest` per actor, like `pytest <specific actor folder>`. This way the need for unique filenames is limited to the `<actor>/tests/` folder which is fine because you can't have two files with the same name in a folder. The only corner case would be if someone creates subfolders in the `<actor>/tests/` folder and create same-name testfiles under them. That is a negligible risk though IMO.

pirat89: Thanks for explanation! :)

docs/source/unit-testing.md
bocekm: Reverted.

PR #405, opened 2019-01-08
stdlib: Actor API convenience library

vinzenz:
This patch introduces the actor convenience api for actors and
repository libraries and provides an easy access to the functions that
actors have access to.

Such as:
- Producing messages
- Consuming messages
- Retrieving files or directory paths
    - files that are bundled with the actor itself
    - shared files on repository level
- Access to the currently running actor instance
- Access to the logger of the currently running actor instance
- Reporting errors

### DIFF
--- a/leapp/actors/__init__.py
+++ b/leapp/actors/__init__.py
@@ -20,6 +20,12 @@ class Actor(object):
     actors in the workflow.
     """
 
+    current_instance = None
+    """
+    Instance of the currently executed actor. Within a process only exist one Actor instance. This will allow
+    convenience functions for library developers to be available.
+    """
+
     ErrorSeverity = ErrorSeverity
     """ Convenience forward for the :py:class:`leapp.models.error_severity.ErrorSeverity` constants. """
 
@@ -60,6 +66,7 @@ class Actor(object):
     """
 
     def __init__(self, messaging=None, logger=None):
+        Actor.current_instance = self
         install_translation_for_actor(type(self))
         self._messaging = messaging
         self.log = (logger or logging.getLogger('leapp.actors')).getChild(self.name)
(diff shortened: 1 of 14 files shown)


radomil:
LGTM

leapp/libraries/stdlib/api.py
```diff
@@ -0,0 +1,163 @@
...
+ErrorSeverity = Actor.ErrorSeverity
+
+
+def current_actor():
+    """
+    Retrieve the Actor class instance of the current active actor.
+    :return: Instance of the currently instantiated actor.
+    :rtype: leapp.actors.Actor
+    """
+    return Actor.current_actor
```
pirat89:
```suggestion
    return Actor.current_instance
```

leapp/actors/__init__.py
```diff
@@ -92,36 +100,102 @@ def common_files_paths(self):
...
+    def get_file_path(self, name):
+        """
+        Finds the first matching file path within :py:attr:`files_paths`.
+
+        :param name: Name of the file
+        :type name: str
+        :return: Found file path
+        :rtype: str or None
+        """
+        return self._get_folder_path(self.files_paths, name)
```
pirat89:
```suggestion
        return self._get_file_path(self.files_paths, name)
```

leapp/actors/__init__.py
```diff
@@ -92,36 +100,102 @@ def common_files_paths(self):
         """ Returns all common repository file paths. """
         return os.getenv("LEAPP_COMMON_FILES", "").split(":")
 
-    def get_folder_path(self, name):
+    def _get_folder_path(self, directories, name):
```
pirat89: why we are mixing `folder` and `directory` (not only here)? I prefer to use *directory* equivalent everywhere. So `_get_dir_path`, ....

pirat89:
Discussed in person. There are missing tests just for `request_answer` function. As we need this PR really already, we will take it just with all those tests we have now as it's enough now and an issue will be created to add those missing tests later.

Issue #407, opened 2019-01-08
Installing leapp breaks pytest on rhel7


drehak:
**Reproducer:**

1. Install packages `leapp` and `pytest` on a RHEL7 system
2. Run `py.test`:

```
Traceback (most recent call last):
  File "/usr/bin/py.test", line 9, in <module>
    load_entry_point('pytest==2.7.0', 'console_scripts', 'py.test-2.7')()
  File "/usr/lib/python2.7/site-packages/_pytest/config.py", line 32, in main
    config = _prepareconfig(args, plugins)
  File "/usr/lib/python2.7/site-packages/_pytest/config.py", line 85, in _prepareconfig
    pluginmanager=pluginmanager, args=args)
  File "/usr/lib/python2.7/site-packages/_pytest/core.py", line 521, in __call__
    return self._docall(self.methods, kwargs)
  File "/usr/lib/python2.7/site-packages/_pytest/core.py", line 528, in _docall
    firstresult=self.firstresult).execute()
  File "/usr/lib/python2.7/site-packages/_pytest/core.py", line 393, in execute
    return wrapped_call(method(*args), self.execute)
  File "/usr/lib/python2.7/site-packages/_pytest/core.py", line 109, in wrapped_call
    wrap_controller.send(call_outcome)
  File "/usr/lib/python2.7/site-packages/_pytest/helpconfig.py", line 28, in pytest_cmdline_parse
    config = outcome.get_result()
  File "/usr/lib/python2.7/site-packages/_pytest/core.py", line 138, in get_result
    py.builtin._reraise(*ex)
  File "/usr/lib/python2.7/site-packages/_pytest/core.py", line 123, in __init__
    self.result = func()
  File "/usr/lib/python2.7/site-packages/_pytest/core.py", line 394, in execute
    res = method(*args)
  File "/usr/lib/python2.7/site-packages/_pytest/config.py", line 636, in pytest_cmdline_parse
    self.parse(args)
  File "/usr/lib/python2.7/site-packages/_pytest/config.py", line 746, in parse
    self._preparse(args)
  File "/usr/lib/python2.7/site-packages/_pytest/config.py", line 713, in _preparse
    self.pluginmanager.consider_setuptools_entrypoints()
  File "/usr/lib/python2.7/site-packages/_pytest/core.py", line 278, in consider_setuptools_entrypoints
    plugin = ep.load()
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 2260, in load
[...]

PR #409, opened 2019-01-10
snactor.fixture: Conditional pyfunc hook

vinzenz:
pytest.hookimpl was introduced in pytest 2.8 - Older version would break
on the fact that we use the decorator. This patch will make the
definition of the hook conditional based on the existance of the
decorator in the module.

Fixes: https://github.com/oamg/leapp/issues/407


pirat89:
looks good, just there is some pyfu...nc breaking the tests :) fix that first

PR #411, opened 2019-01-10
Differentiate unit/component tests in tutorial

bocekm:
- and mention the use of Makefile for test dependencies


pirat89:
@bocekm Yes, I mean this. And I would require such names. Otherwise there is not any reason to split unit tests and component tests into two files. The requirement of those names is only reason for me to modify documentation in this way.

bocekm:
Well, I'm against requiring them. You can have multiple library modules and then it would be better to have one test file per one library module. The current naming convention is there:
```
All test functions have to:
 - reside in `test_*.py` or `*_test.py` files,
```
I would keep it this way. Somebody can even have unit tests and component test in one file. I think it's not a problem.

pirat89:
@bocekm You are right. You persuaded me.

PR #412, opened 2019-01-11
Add exceptions for stopping actor execution

jmikovic:


leapp/workflows/__init__.py
```diff
@@ -191,7 +192,12 @@ def run(self, context=None, until_phase=None, until_actor=None, skip_phases_unti
                                                                                        actor=actor.name))
                     messaging = InProcessMessaging()
                     messaging.load(actor.consumes)
-                    actor(logger=current_logger, messaging=messaging).run()
+                    try:
```
vinzenz: I would rather put it within the run function of the actor

leapp/exceptions.py
```diff
@@ -105,3 +105,13 @@ def __init__(self, message):
...
+
+
+class StopActorExecution(Exception):
+    def __init__(self):
+        super(StopActorExecution, self).__init__()
+
+
+class StopActorExecutionError(LeappError):
+    def __init__(self, message):
+        super(StopActorExecutionError, self).__init__(message)
```
vinzenz: You would want also the additional options covered that are covered by report_error such as details and severity

leapp/workflows/__init__.py
```diff
@@ -191,7 +192,12 @@ def run(self, context=None, until_phase=None, until_actor=None, skip_phases_unti
...
+                        actor(logger=current_logger, messaging=messaging).run()
+                    except StopActorExecution:
+                        pass
+                    except StopActorExecutionError as err:
+                        actor.report_error(err.message)
```
vinzenz: here one would want to be able to pass severity and details to report_error

vinzenz:
Also I am missing the documentation part, 'How to use it'

jmikovic:
I realize that documentation is missing, that's why I put it as WIP

leapp/exceptions.py
```diff
@@ -105,3 +105,15 @@ def __init__(self, message):
...
         super(LeappRuntimeError, self).__init__(message)
+
+
+class StopActorExecution(Exception):
+    def __init__(self):
+        super(StopActorExecution, self).__init__()
+
+
+class StopActorExecutionError(LeappError):
+    def __init__(self, message,  severity='error', details=None):
```
vinzenz: please do not use this literal, but the appropriate ErrorSeverity constant

vinzenz:
```python
from leapp.models.error_severity import ErrorSeverity

def __init__(self, message,  severity='error', details=None):
```

leapp/exceptions.py
pirat89: Thinking that I am missing function to stop execution. Instead of raw raise of exception, there would be function for that like `stop_execution`, `exit_error`, `exit`, ... Maybe make it part of API.

vinzenz: Irrelevant to this, this is a convenience function that would be added on top of it.

PR #416, opened 2019-01-14
WIP: Move GitHub Pages here from obsolete leapp-to org

bocekm:
- we have a new GitHub org OAMG but till now we've been having GitHub
Pages still under the obsolete leapp-to org (leapp-to.github.io)
- moved from https://github.com/leapp-to/leapp-to.github.io

TODO:
- move the content to a completely separate directory (update _config.yml accordingly)


bocekm:
To avoid melding source for two different documentations in the _docs/_ we've moved the code to the [`gh-pages` branch](https://github.com/oamg/leapp/tree/gh-pages).

PR #424, opened 2019-01-22
Add metapackage for dependencies

pirat89:
Add metapackage for dependencies
    
    Currently we are not able to use python dependencies correctly to
    ensure the RPM will not be removed during the upgrade of the system.
    As a workround, put all such system dependencies into the leapp-deps
    metapackage.
    
    This metapackage is supposed to be built for the same system where
    leapp should be installed. When different dependencies are needed
    on target system (e.g. during the in-place upgrade) repository is
    reponsible to deliver new meta package with such dependencies.
    
    Leapp framework from this point will require specific capability:
      leapp-framework-dependencies
    
    which has to be provided by such meta package. Any time the
    dependencies are changed, the capability number has to be incremented
    by one. So packages provided by leapp repository has to reflect such
    change as well.

### DIFF
--- a/packaging/leapp.spec
+++ b/packaging/leapp.spec
@@ -30,6 +30,7 @@ Requires: python2-%{name} = %{version}-%{release}
 %endif
 Requires: leapp-repository >= %{version}
 
+
 %description
 Leapp tool for handling upgrades.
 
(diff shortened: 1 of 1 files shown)


shaded-enmity:
LGTM

PR #426, opened 2019-01-23
add api function for showing messages to user

jmikovic:


vinzenz:
Since the test is missing even for request_answers it's ok that it's missing here

PR #427, opened 2019-01-23
Ext program exec function leapp stdlib call with audit

radomil:
Leapp stdlib Call function enrichment with audit data generation and storage.
Names of these new_... functions are to be changed to just simply to call, checked_call, later when replacing the current call definition.

### DIFF
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -14,13 +14,14 @@ import six
 from leapp.exceptions import LeappError
 from leapp.utils.audit import create_audit_entry
 from leapp.libraries.stdlib import api
-from leapp.libraries.stdlib.call import _call
+from leapp.libraries.stdlib.call import _call, STDOUT
 
 
 class CalledProcessError(LeappError):
-    """ Leapp Call Process Exception Error.
+    """
+    Leapp Call Process Exception Error.
 
-    Raised when the result of a called process is a none zero return code.
+    Raised when the result of a called process is of a none zero return code.
     """
     def __init__(self, message, command, result):
         """
@@ -101,55 +102,50 @@ def call(args, split=True):
 def _logging_handler(fd_info, buffer):
     """
     Log while in a leapp debug mode
-    :param fd_info:
-    :param buffer:
+
+    :param fd_info: Contains File descriptor type
+    :type fd_info: tuple
+    :param buffer: buffer interface
+    :type buffer: bytes array
     """
     (_unused, fd_type) = fd_info
     if os.getenv('LEAPP_DEBUG', '0') == '1':
-        if fd_type == _call.STDOUT:
+        if fd_type == STDOUT:
             sys.stdout.write(buffer)
         else:
             sys.stderr.write(buffer)
 
 
-def run(args):
+def run(args, split=False):
     """
-    Run the program described by args
-    and wait for the program to complete and return the results as a dict.
+    Run a command and return its result as a dict.
+
     The execution of the program and it's results are captured by the audit.
 
     :param args: Command to execute
-    :return: stdout output, 'utf-8' decoded
+    :type args: list or tuple
+    :param split: Split the output on newlines
+    :type split: bool
+    :return: {'stdout' : stdout, 'stderr': 'signal': signal, 'exit_code': exit_code, 'pid': pid}
+    :rtype: dict
     """
     _id = str(uuid.uuid4())
-    result_data = None
+    result = None
     try:
         create_audit_entry('process-start', {'id': _id, 'parameters': args})
-        result_data = _call(args, callback_raw=_logging_handler)
+        result = _call(args, callback_raw=_logging_handler)
+        if result['exit_code'] != 0:
+            raise CalledProcessError(
+                message="A Leapp Command Error occurred. ",
+                command=args,
+                result=result
+            )
+        else:
+            if split:
+                result.update({
+                    'stdout': result['stdout'].splitlines()
+                })
     finally:
         create_audit_entry('process-end', _id)
-        create_audit_entry('process-result', {'id': _id, 'parameters': args, 'result': result_data})
-    return result_data
-
-
-def checked_call(args, split=False):
-    """
-    Run the program described by args and wait for the program to complete and return the results as a dict.
-    The execution of the program and it's results are captured by the audit.
-
-
-    :param args: Command to execute
-    :param split: Split the output on newlines
-    :return: stdout output, 'utf-8' decoded, split by lines if split=True
-    """
-    result = run(args)
-    if result['exit_code'] != 0:
-        raise CalledProcessError(
-            message="A Leapp Command Error occurred. ",
-            command=args,
-            result=result
-        )
-    else:
-        if split:
-            return result['stdout'].splitlines()
-        return result['stdout']
+        create_audit_entry('process-result', {'id': _id, 'parameters': args, 'result': result})
+    return result
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -17,7 +17,9 @@ from leapp.libraries.stdlib import call, api
 
 
 class CalledProcessError(LeappError):
-    """ Leap Call Process Exception Error
+    """ Leapp Call Process Exception Error
+
+    Raised when the result of a called process is a none zero return code.
     """
     def __init__(self, message, command, result):
         super(CalledProcessError, self).__init__(message)
@@ -26,22 +28,42 @@ class CalledProcessError(LeappError):
 
     @property
     def stdout(self):
+        """
+        Retrieve the stdout.
+        :return:
+        """
         return self._result.get('stdout')
 
     @property
     def stderr(self):
+        """
+        Retrieve the stderr.
+        :return:
+        """
         return self._result.get('stderr')
 
     @property
     def exit_code(self):
+        """
+        Retrieve the exit code.
+        :return:
+        """
         return self._result.get('exit_code')
 
     @property
     def signal(self):
+        """
+        Retrieve the signal which the process was signalled by.
+        :return:
+        """
         return self._result.get('signal')
 
     @property
     def pid(self):
+        """
+        Retrieve the pid of the finished process.
+        :return:
+        """
         return self._result.get('pid')
 
 
@@ -70,6 +92,11 @@ def call(args, split=True):
 
 
 def _logging_handler(fd_info, buffer):
+    """
+    Log while in a leapp debug mode
+    :param fd_info:
+    :param buffer:
+    """
     (_unused, fd_type) = fd_info
     if os.getenv('LEAPP_DEBUG', '0') == '1':
         if fd_type == call.STDOUT:
@@ -78,9 +105,11 @@ def _logging_handler(fd_info, buffer):
             sys.stderr.write(buffer)
 
 
-def new_call(args):
+def run(args):
     """
-    Call the Leapp's stdlib _call
+    Run the program described by args
+    and wait for the program to complete and return the results as a dict.
+    The execution of the program and it's results are captured by the audit.
 
     :param args: Command to execute
     :return: stdout output, 'utf-8' decoded
@@ -96,19 +125,21 @@ def new_call(args):
     return result_data
 
 
-def new_checked_call(args, split=False):
+def checked_call(args, split=False):
     """
-    Call the audited call and check result, split the result into multiple lines if requested
+    Run the program described by args and wait for the program to complete and return the results as a dict.
+    The execution of the program and it's results are captured by the audit.
+
 
     :param args: Command to execute
     :param split: Split the output on newlines
     :return: stdout output, 'utf-8' decoded, split by lines if split=True
     """
-    result = new_call(args)
+    result = run(args)
     if result['exit_code'] != 0:
         raise CalledProcessError(
             message="A Leapp Command Error occurred. ",
-            command=str(args[0]),
+            command=args,
             result=result
         )
     else:
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -13,7 +13,8 @@ import six
 
 from leapp.exceptions import LeappError
 from leapp.utils.audit import create_audit_entry
-from leapp.libraries.stdlib import call, api
+from leapp.libraries.stdlib import api
+from leapp.libraries.stdlib.call import _call
 
 
 class CalledProcessError(LeappError):
@@ -99,7 +100,7 @@ def _logging_handler(fd_info, buffer):
     """
     (_unused, fd_type) = fd_info
     if os.getenv('LEAPP_DEBUG', '0') == '1':
-        if fd_type == call.STDOUT:
+        if fd_type == _call.STDOUT:
             sys.stdout.write(buffer)
         else:
             sys.stderr.write(buffer)
@@ -118,7 +119,7 @@ def run(args):
     result_data = None
     try:
         create_audit_entry('process-start', {'id': _id, 'parameters': args})
-        result_data = call._call(args, callback_raw=_logging_handler)
+        result_data = _call(args, callback_raw=_logging_handler)
     finally:
         create_audit_entry('process-end', _id)
         create_audit_entry('process-result', {'id': _id, 'parameters': args, 'result': result_data})
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -6,24 +6,20 @@ and at the same time, they are really useful for other actors.
 import os
 import sys
 
-import six
 import subprocess
 import uuid
 
+import six
+
+from leapp.exceptions import CommandError
 from leapp.utils.audit import create_audit_entry
-from leapp.libraries.stdlib import call
-
-
-class CalledProcessError(Exception):
-    pass
-
-from leapp.libraries.stdlib import api
+from leapp.libraries.stdlib import call, api
 
 
 def call(args, split=True):
     """
-    Call an external program, capture and automatically utf-8 decode its ouput.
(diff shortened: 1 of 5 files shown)


leapp/libraries/stdlib/call.py
```diff
@@ -37,17 +37,18 @@ def _multiplex(ep, read_fds, callback_raw, callback_linebuffered,
     while not ep.closed and len(hupped) != num_expected:
         events = ep.poll(timeout)
         for fd, event in events:
+            translated_fd = read_fds.index(fd) + 1
```
shaded-enmity: We'll need to think a little bit more about this, technically this works because the index of the element in the array corresponds to the actual numerical value of STDOUT/STDERR, however the value of `translated_fd` does not represent the real file descriptor, because both `stdout` and `stderr` are actually pipes (https://github.com/oamg/leapp/blob/5d1ee15732f5be216eab60560f9682a5719ec4ca/leapp/libraries/stdlib/call.py#L108).

I'd rename `translated_fd` to `fd_type`, to better reflect what the value represents, and change the semantics of `callback_{raw,linebuffered}` to treat the first argument as a tuple -- (fd, fd_type), this way we can still distinguish whether we're dealing with `stdout` or `stderr` while retaining the access to the real file descriptor.

leapp/libraries/stdlib/__init__.py
```diff
@@ -30,3 +40,31 @@ def call(args, split=True):
...
+
+def new_call(args, split=False):
+    _id = str(uuid.uuid4())
+    result_data = None
+    try:
+        create_audit_entry('process-start', {'id': _id, 'parameters': args})
+        result_data = call._call(args, callback_raw=_logging_handler)
+    finally:
+        create_audit_entry('process-end', _id)
+        create_audit_entry('process-result', {'id': _id, 'parameters': args, 'result': result_data})
```
shaded-enmity: @vinzenz If `result_data` is buffer/bytes via: https://github.com/oamg/leapp/blob/5d1ee15732f5be216eab60560f9682a5719ec4ca/leapp/libraries/stdlib/call.py#L164 can the Audit serializer deal with that?

leapp/utils/audit/__init__.py
```diff
@@ -257,6 +257,17 @@ def do_store(self, connection):
         self._message_id = cursor.lastrowid
 
 
+def create_audit_entry(event, data):
```
shaded-enmity: Missing docstrings for the function

leapp/libraries/stdlib/__init__.py
```diff
@@ -30,3 +40,31 @@ def call(args, split=True):
...
+def _logging_handler(fd, buffer):
+    if os.getenv('LEAPP_DEBUG', '0') == '1':
+        if fd == call.STDOUT:
+            sys.stdout.write(buffer)
+        else:
+            sys.stderr.write(buffer)
+
+
+def new_call(args, split=False):
+    _id = str(uuid.uuid4())
```
shaded-enmity: I'm wondering why are we creating the ID here, and not in the database.

leapp/libraries/stdlib/__init__.py
vinzenz: AFAIK yes because it ahould be JSON serializable and that’s all that is required

leapp/libraries/stdlib/__init__.py
vinzenz: Because it is not a ID in sense of a key in the db but a correlation ID for the call to be able to correlate multiple events.

leapp/libraries/stdlib/__init__.py
```diff
@@ -30,3 +40,45 @@ def call(args, split=True):
     if split:
         return r.splitlines()
     return r
+
+
+def _logging_handler(fd, buffer):
```
shaded-enmity: Update callback function signature
```suggestion
def _logging_handler((fd, fd_type), buffer):
```

Also in the body of the function we should check `fd_type` and not `fd`

shaded-enmity:
Looking good, the only thing missing are tests, i.e. make sure that the calls are properly audited.

leapp/libraries/stdlib/__init__.py
```diff
@@ -3,9 +3,19 @@
...
 import six
 import subprocess
-import os
+import uuid
+
+from leapp.utils.audit import create_audit_entry
+from leapp.libraries.stdlib import call
+
+
+class CalledProcessError(Exception):
```
vinzenz: This should be based on [leapp.exceptions.LeappError](https://leapp.readthedocs.io/en/latest/pydoc/leapp.html#leapp.exceptions.LeappError) and also contain stderr, stdout, return code etc

leapp/libraries/stdlib/__init__.py
```diff
@@ -3,9 +3,19 @@
 represents a location for functions that otherwise would be defined multiple times across leapp actors
 and at the same time, they are really useful for other actors.
 """
+import os
+import sys
+
```
vinzenz: this space should be here and the imports need to be ordered by name.

leapp/libraries/stdlib/__init__.py
```diff
@@ -3,9 +3,19 @@
...
 import six
```
vinzenz: import six should be after python standard library imports (according to pep8)

radomil:
leapp-ci build

leapp/libraries/stdlib/__init__.py
pirat89: @vinzenz  `LeappError` or `CommandError`? (the second one is used)

radomil:
leapp-ci build

radomil:
Failing make command
```
make: *** [test] Error 1
The command "if [ ! -z "$CONTAINER" ]; then make container-test; else make test; fi" exited with 2.
```

vinzenz:
_______________________ FLAKE8-check(ignoring F811 F401) _______________________
/home/travis/build/oamg/leapp/leapp/libraries/stdlib/__init__.py:80:85: W504 line break after binary operator

radomil:
1 out of 3 environments tests/build passes. The flake8 warning in the other 2 environments doesn't seem to as a build breaker. At least I will reformat that line.

vinzenz:
Well, the container version runs in a CentOS environment, to replicate as close as possible the RHEL7 environment. So if checks don't fail there, it's because the checking tool doesn't check for that yet, however the other two environments are more up to date.

Anyway the problem is the Flake8 check result

leapp/libraries/stdlib/__init__.py
radomil: @shaded-enmity ,  suggested change has been done
https://travis-ci.org/oamg/leapp/builds/492657740
Failing in python 3.6 env (SyntaxError: invalid syntax)
In 2.7 env it's ok

leapp/libraries/stdlib/__init__.py
vinzenz: @pirat89 LeappError is the baseclass of all leapp Exceptions

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +38,51 @@ def call(args, split=True):
     if split:
         return r.splitlines()
     return r
+
+
+def _logging_handler((fd, fd_type), buffer):
```
```suggestion
def _logging_handler(fd_info, buffer):
    (fd, fd_type) = fd_info
```

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +38,51 @@ def call(args, split=True):
...
+    :param args: Command to execute
+    :param split: Split the output on newlines
+    :return: stdout output, 'utf-8' decoded, split by lines if split=True
+    """
+    result = new_call(args)
+    if result['exit_code'] != 0:
+        failed_command = "Command: " + str(args[0])
+        command_stdout = "Result stdout: " + result['stdout']
+        error_message = " stderr: " + result['stderr']
+        raise CommandError(
```
vinzenz: No, no, this is no good, do not use CommandError here. This is not for this purpose. It was meant to be used Leapp tools CLI handling.

The idea of defining a CalledProcessError was much better, since that's what people are used to.
```python
class CalledProcessError(LeappError):
    def __init__(self, message, command, result):
        super(CalledProcessError, self).__init__(message)
        self._result = result
        self.command = command
    
    @property
    def stdout(self):
        self._result.get('stdout')

    @property
    def stderr(self):
        self._result.get('stderr')

    @property
    def exit_code(self):
        self._result.get('exit_code')

    @property
    def signal(self):
        self._result.get('signal')

    @property
    def pid(self):
        self._result.get('pid')
```
Something along these lines. This way you can get the data and don't try to consider stderr to be an error message. This is not a correct assumption. This can be tons of text....

leapp/libraries/stdlib/__init__.py
radomil: solved

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +67,51 @@ def call(args, split=True):
...
+
+    :param args: Command to execute
+    :param split: Split the output on newlines
+    :return: stdout output, 'utf-8' decoded, split by lines if split=True
+    """
+    result = new_call(args)
+    if result['exit_code'] != 0:
+        raise CalledProcessError(
+            message="A Leapp Command Error occurred. ",
+            command=str(args[0]),
```
vinzenz: id' expect the whole 'args' here, not just the first command.

radomil: sure, updated

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +67,51 @@ def call(args, split=True):
...
+    if os.getenv('LEAPP_DEBUG', '0') == '1':
+        if fd_type == call.STDOUT:
+            sys.stdout.write(buffer)
+        else:
+            sys.stderr.write(buffer)
+
+
+def new_call(args):
+    """
+    Call the Leapp's stdlib _call
```
vinzenz: That’s not a good docstring, it tells me only something about the usage of an internal function.

Better would be perform something like:
Run the program described by args and wait for the program to complete and return the results as a dict. The execution of the program and it’s results are captured by the audit.

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +67,51 @@ def call(args, split=True):
...
+    try:
+        create_audit_entry('process-start', {'id': _id, 'parameters': args})
+        result_data = call._call(args, callback_raw=_logging_handler)
+    finally:
+        create_audit_entry('process-end', _id)
+        create_audit_entry('process-result', {'id': _id, 'parameters': args, 'result': result_data})
+    return result_data
+
+
+def new_checked_call(args, split=False):
```
vinzenz: Since we do not have a checked_call yet just all it that

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +67,51 @@ def call(args, split=True):
...
+def _logging_handler(fd_info, buffer):
+    (_unused, fd_type) = fd_info
+    if os.getenv('LEAPP_DEBUG', '0') == '1':
+        if fd_type == call.STDOUT:
+            sys.stdout.write(buffer)
+        else:
+            sys.stderr.write(buffer)
+
+
+def new_call(args):
```
vinzenz: How about audited_call or just run? That would be similar to subprocess.

leapp/libraries/stdlib/__init__.py
```diff
@@ -3,17 +3,52 @@
...
+from leapp.exceptions import LeappError
+from leapp.utils.audit import create_audit_entry
+from leapp.libraries.stdlib import call, api
+
 
-from leapp.libraries.stdlib import api
+class CalledProcessError(LeappError):
+    """ Leap Call Process Exception Error
+    """
+    def __init__(self, message, command, result):
```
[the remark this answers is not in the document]
bocekm: I seems to me this comment has not been resolved even though it's been marked as such.

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +90,60 @@ def call(args, split=True):
     if split:
         return r.splitlines()
     return r
+
+
+def _logging_handler(fd_info, buffer):
+    """
+    Log while in a leapp debug mode
+    :param fd_info:
+    :param buffer:
```
bocekm: Complete the param description.

leapp/utils/audit/__init__.py
```diff
@@ -257,6 +257,50 @@ def do_store(self, connection):
...
+                audit.stamp       AS stamp,
+                audit.data        AS data
+              FROM
+                audit
+              WHERE
+                audit.event = ?
+              ORDER BY stamp ASC;
+        ''', (event,))
+        # ''', (context, event))
+        # audit.context = ? AND audit.event = ?
```
bocekm: Remove the commented out code.

leapp/libraries/stdlib/__init__.py
```diff
@@ -3,17 +3,75 @@
...
+import six
+
+from leapp.exceptions import LeappError
+from leapp.utils.audit import create_audit_entry
 from leapp.libraries.stdlib import api
+from leapp.libraries.stdlib.call import _call
+
+
+class CalledProcessError(LeappError):
+    """ Leapp Call Process Exception Error
```
bocekm: Move the sentence _Leapp Call Process Exception Error_ to a new line.

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +90,60 @@ def call(args, split=True):
     if split:
         return r.splitlines()
     return r
+
+
+def _logging_handler(fd_info, buffer):
+    """
+    Log while in a leapp debug mode
```
bocekm: Mention where is the buffer being logged - it is logged to stdout or stderr.

leapp/libraries/stdlib/__init__.py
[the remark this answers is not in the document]
radomil: @bocekm oh, ok, I documented all but this. I just pushed the update.

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +90,60 @@ def call(args, split=True):
...
+    try:
+        create_audit_entry('process-start', {'id': _id, 'parameters': args})
+        result_data = _call(args, callback_raw=_logging_handler)
+    finally:
+        create_audit_entry('process-end', _id)
+        create_audit_entry('process-result', {'id': _id, 'parameters': args, 'result': result_data})
+    return result_data
+
+
+def checked_call(args, split=False):
```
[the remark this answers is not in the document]
bocekm: After reading the implementation the docstring is incorrect. You're not returning a dict but a either string or list of strings.

leapp/libraries/stdlib/call.py
```diff
@@ -64,7 +64,7 @@ def _multiplex(ep, read_fds, callback_raw, callback_linebuffered,
     # Process leftovers from line buffering
     for (fd, lb) in linebufs.items():
         if lb:
-            callback_linebuffered(fd, lb)
+            callback_linebuffered(read_fds.index(fd) + 1, lb)
```
bocekm: Explain the magic _+1_ in a comment please.

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +90,60 @@ def call(args, split=True):
...
+
+
+def run(args):
+    """
+    Run the program described by args
+    and wait for the program to complete and return the results as a dict.
+    The execution of the program and it's results are captured by the audit.
+
+    :param args: Command to execute
+    :return: stdout output, 'utf-8' decoded
```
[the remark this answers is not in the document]
bocekm: Could you please describe _:return:_ for the _call() in its docstring?

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +96,60 @@ def call(args, split=True):
...
+
+def _logging_handler(fd_info, buffer):
+    """
+    Log while in a leapp debug mode
+    :param fd_info:
+    :param buffer:
+    """
+    (_unused, fd_type) = fd_info
+    if os.getenv('LEAPP_DEBUG', '0') == '1':
+        if fd_type == _call.STDOUT:
```
bocekm: The function __call_ has no STDOUT member, the _call_ module has.

radomil:
@vinzenz , @bocekm just pinging you for the change requests have been addressed.

leapp/libraries/stdlib/__init__.py
```diff
@@ -32,3 +90,60 @@ def call(args, split=True):
...
+            sys.stderr.write(buffer)
+
+
+def run(args):
+    """
+    Run the program described by args
+    and wait for the program to complete and return the results as a dict.
+    The execution of the program and it's results are captured by the audit.
+
+    :param args: Command to execute
```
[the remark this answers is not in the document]
bocekm: Not resolved.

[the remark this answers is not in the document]
bocekm: Sorry, it is.

leapp/libraries/stdlib/__init__.py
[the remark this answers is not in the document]
bocekm: As I write above, I'd love to see the `:return: {'stdout' : stdout, 'stderr': 'signal': signal, 'exit_code': exit_code, 'pid': pid}` added to the __call_ in _leapp/libraries/stdlib/call.py_.

leapp/libraries/stdlib/__init__.py
radomil: As it logs either to STDOUT or STDERR, its description was updated and extended accordingly.

leapp/libraries/stdlib/__init__.py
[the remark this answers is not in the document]
radomil: _call in call.py updated now.

PR #429, opened 2019-01-23
Release 0.5.0

pirat89:
- Models has been refactored to be more comprehensible and reliable
- Introduced standard library
- Introduced the actor convenience api for actors and repository libraries
- Added localization support
- Extended serialization support
- Added exception to be able to stop actor execution


bocekm:
Update to leapp/man/snactor.1 is missing.

Issue #431, opened 2019-01-24
snactor discover: wrong paths displayed for objects with the same name


drehak:
**Reproducer:**
Create a repository and a tag inside it. Run discover.
```
$ snactor repo new chain
New repository chain has been created in /home/user/chain
$ cd chain/
$ snactor new-tag chain
New tag ChainTag has been created in /home/user/chain/tags/chain.py
$ snactor discover
Repository:
  Name: chain
  Path: /home/user/chain

Actors(0):

Models(0):

Tags(1):
   - ChainTag                            tags/chain.py

Topics(0):

Workflows(0):
```
Create a topic with the same name. Now discover displays a wrong path for our tag.
```
$ snactor new-topic chain
New topic ChainTopic has been created in /home/user/chain/topics/chain.py
$ snactor discover
Repository:
  Name: chain
  Path: /home/user/chain

Actors(0):

Models(0):

Tags(1):
   - ChainTag                            topics/chain.py

Topics(1):
   - ChainTopic                          topics/chain.py

Workflows(0):
```
Similar behavior occurs after creating a model with the very same name.
```
$ snactor new-model --topic ChainTopic chain
New model Chain has been created in /home/user/chain/models/chain.py
$ snactor discover
Repository:
  Name: chain
  Path: /home/user/chain

Actors(0):

Models(1):
   - Chain                               models/chain.py

Tags(1):
   - ChainTag                            models/chain.py

Topics(1):
   - ChainTopic                          models/chain.py

Workflows(0):
```
**Expected output:**
```
$ snactor discover
Repository:
  Name: chain
  Path: /home/user/chain

Actors(0):

Models(1):
   - Chain                               models/chain.py

Tags(1):
   - ChainTag                            tags/chain.py

Topics(1):
   - ChainTopic                          topics/chain.py

Workflows(0):
```
snactor version 0.5.0

shaded-enmity:
@drehak Can you please also check the `--safe` variant of `snactor discover`?

drehak:
Looks alright.
```
$ snactor discover --safe
Repository:
  Name: chain
  Path: /home/user/chain

Actors

Models
  - Chain(Model) ............................................ models/chain.py

Tags
  - ChainTag(Tag) ........................................... tags/chain.py
```

vinzenz:
Fixed in #438

PR #438, opened 2019-02-08
snactor discover: fix wrong path

fernflower:
After registering several different objects
with the same name (ex. actor and tag),
snactor discover without --safe would return
wrong path for actual files for all objects
but the one registered last.
This fix addresses the issue by utilizing
the fact that repo structure remains the same
and adding object type check in _get_class_file
path resolution.

Closes-Issue: #431

### DIFF
--- a/leapp/repository/__init__.py
+++ b/leapp/repository/__init__.py
@@ -120,13 +120,13 @@ class Repository(object):
         if not stage or stage is _LoadStage.INITIAL:
             self.log.debug("Loading repository %s", self.name)
             self.log.debug("Loading tag modules")
-            self._load_modules(self.tags)
+            self._load_modules(self.tags, 'leapp.tags')
             self.log.debug("Loading topic modules")
-            self._load_modules(self.topics)
+            self._load_modules(self.topics, 'leapp.topics')
 
         if not stage or stage is _LoadStage.MODELS:
             self.log.debug("Loading model modules")
-            self._load_modules(self.models)
+            self._load_modules(self.models, 'leapp.models')
             if resolve:
                 from leapp.models import resolve_model_references
                 resolve_model_references()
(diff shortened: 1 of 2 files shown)


fernflower:
vinzenz, sure. Another approach I can think of it to judge based on mro()

leapp/repository/__init__.py
```diff
@@ -172,10 +172,12 @@ def _load_libraries(self, path=None, mod=None, prefix='leapp.libraries.common'):
             if is_pkg:
                 self._load_libraries(path=(os.path.dirname(loaded.__file__),), mod=loaded, prefix=mod_full_name)
 
-    def _load_modules(self, modules):
-        directories = tuple(os.path.join(self._repo_dir, os.path.dirname(module)) for module in modules)
-        for importer, name, is_pkg in pkgutil.iter_modules(directories):
-            importer.find_module(name).load_module(name)
+    def _load_modules(self, modules, mod_name=None):
```
vinzenz: unused variable

leapp/repository/__init__.py
```diff
@@ -172,10 +172,12 @@ def _load_libraries(self, path=None, mod=None, prefix='leapp.libraries.common'):
...
+        for module in modules:
+            directory = os.path.join(self._repo_dir, os.path.dirname(module))
+            prefix = 'leapp.%s.' % directory.split(os.path.sep)[-1]
```
vinzenz: Don't even try to make it that magically, just get the prefix to use from the caller. There is only function where this is called from with 4 calls where this is relevant.

vinzenz: I would just change the calls in `def load` to this:

`self._load_modules(self.tags)` => `self._load_modules(self.tags, prefix='leapp.tags.')`
`self._load_modules(self.topics)` => `self._load_modules(self.topics, prefix='leapp.topics.')`
`self._load_modules(self.models)` => `self._load_modules(self.models, prefix='leapp.models.')`
`self._load_modules(self. workflows)` => `self._load_modules(self.workflows, prefix='leapp.workflows.')`

Pass through the prefix and be done with it. That's IMHO safer in case we change something about the directory structure. The assumption that basename(dirname(module_path)) gives you the right type, isn't a valid assumption. it could be easily models/something/somemodel.py

fernflower: Makes sense, will fix that. 
But this will change the semantics of _load_modules a bit, now it will be impossible to pass a list of heterogeneous modules, as they will be registered as one type only. Probably this should be put in a docstring.

leapp/repository/__init__.py
fernflower: ooops, will fix that

vinzenz:
LGTM! Thanks a lot!

PR #441, opened 2019-02-14
snactor: Workflow run --save-output

vinzenz:
This implements a new feature for the `snactor workflow run` command,
that allows to store all messages from snactor workflow run to be
available for consumption by actor executed through `snactor run`.

Usage scenario:
---------------

One wants to write an actor that needs to consume from the
FactsCollection phase of the IPU workflow.

Before we introduced this parameter, one had to manually run every
actor, that is required to produce the messages the actor needs to
consume, manually with the --save-output parameter.

With the new feature one can now run:

$ snactor workflow run --save-output --until-phase FactsCollection ipu

And then can normally use `snactor run` to start testing the actor.

$ snactor run MyNewActor --print-output

### DIFF
--- a/leapp/snactor/commands/workflow/run.py
+++ b/leapp/snactor/commands/workflow/run.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
 import leapp.workflows
+import os
 import sys
 from leapp.exceptions import LeappError, UsageError, CommandError
 from leapp.snactor.commands.workflow import workflow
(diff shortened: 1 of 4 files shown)


pirat89:
@fernflower Yes. We discussed that with @vinzenz (if I remember well that it was about this) that this is useful and tests would take a lot of time now, so there will be created issue about that so anyone would take it and write those tests when we will have time. I will test this manually now.

shaded-enmity:
LGTM, however I'd like to see at least a skeleton for testing (test file with unimplemented `test_*` functions), so that someone else can pick it up and implement the actual logic.

PR #442, opened 2019-02-15
builds: Fedora builds - Only snactor + framework

vinzenz:


packaging/leapp.spec
```diff
@@ -203,7 +205,7 @@ install -m 0644 -p man/snactor.1 %{buildroot}%{_mandir}/man1/
 %dir %{_sysconfdir}/leapp/repos.d
 %{_bindir}/leapp
```
bocekm: I'm getting an error on fedora:
```
Checking for unpackaged file(s): /usr/lib/rpm/check-files /home/mbocek/rpmbuild/BUILDROOT/leapp-0.5.0-0.201902191126Z.1c67a50.fedora_builds.fc29.x86_64
error: Installed (but unpackaged) file(s) found:
   /usr/bin/leapp
```

bocekm: Adding the following to the %install section does the trick. It looks abit ugly/hacky though. @pirat89 can chime in.
```
%if 0%{?fedora}
rm -rf %{buildroot}%{_bindir}/leapp
%endif
```

bocekm:
The _leapp-deps_ package is still being built. @pirat89, is it necessary in case we don't build the _leapp_ package?

packaging/leapp.spec
pirat89: yep, the remove like that is valid.

pirat89:
> The _leapp-deps_ package is still being built. @pirat89, is it necessary in case we don't build the _leapp_ package?

Yes it is valid. Because the snactor and framework would depend on that package as well.

vinzenz:
```
$ sudo dnf install python2-leapp*fedora_builds* leapp-deps*fedora_builds*
Last metadata expiration check: 0:00:30 ago on Tue 26 Feb 2019 03:16:10 PM CET.
Dependencies resolved.
==============================================================================================================================================================
 Package                     Arch                 Version                                                         Repository                             Size
==============================================================================================================================================================
Installing:
 leapp-deps                  noarch               0.5.0-0.201902261159Z.4c9040f.fedora_builds.fc29                group_leapp-leapp-devel               6.8 k
 python2-leapp               noarch               0.5.0-0.201902261159Z.4c9040f.fedora_builds.fc29                group_leapp-leapp-devel               162 k
Installing dependencies:
 python2-six                 noarch               1.11.0-6.fc29                                                   fedora                                 34 k

Transaction Summary
==============================================================================================================================================================
Install  3 Packages

Total download size: 203 k
Installed size: 711 k
Is this ok [y/N]: n 
Operation aborted.
```
```
$ sudo dnf install python3-leapp*fedora_builds* leapp-deps*fedora_builds*
Last metadata expiration check: 0:01:02 ago on Tue 26 Feb 2019 03:16:10 PM CET.
Dependencies resolved.
==============================================================================================================================================================
 Package                     Arch                 Version                                                         Repository                             Size
==============================================================================================================================================================
Installing:
 leapp-deps                  noarch               0.5.0-0.201902261159Z.4c9040f.fedora_builds.fc29                group_leapp-leapp-devel               6.8 k
 python3-leapp               noarch               0.5.0-0.201902261159Z.4c9040f.fedora_builds.fc29                group_leapp-leapp-devel               167 k
Installing dependencies:
 python2-six                 noarch               1.11.0-6.fc29                                                   fedora                                 34 k

Transaction Summary
==============================================================================================================================================================
Install  3 Packages

Total download size: 208 k
Installed size: 654 k
```
```
$ sudo dnf install python3-leapp*fedora_builds*
Last metadata expiration check: 0:01:16 ago on Tue 26 Feb 2019 03:16:10 PM CET.
Dependencies resolved.
==============================================================================================================================================================
 Package                     Arch                 Version                                                         Repository                             Size
==============================================================================================================================================================
Installing:
 python3-leapp               noarch               0.5.0-0.201902261159Z.4c9040f.fedora_builds.fc29                group_leapp-leapp-devel               167 k

Transaction Summary
==============================================================================================================================================================
Install  1 Package

Total download size: 167 k
Installed size: 552 k
```
```
$ rpm -q python3-six
python3-six-1.11.0-6.fc29.noarch
$ sudo dnf install snactor*fedora_builds*
Last metadata expiration check: 0:01:52 ago on Tue 26 Feb 2019 03:16:10 PM CET.
Dependencies resolved.
==============================================================================================================================================================
 Package                     Arch                 Version                                                         Repository                             Size
==============================================================================================================================================================
Installing:
 snactor                     noarch               0.5.0-0.201902261159Z.4c9040f.fedora_builds.fc29                group_leapp-leapp-devel                13 k
Installing dependencies:
 python3-leapp               noarch               0.5.0-0.201902261159Z.4c9040f.fedora_builds.fc29                group_leapp-leapp-devel               167 k

Transaction Summary
==============================================================================================================================================================
Install  2 Packages

Total download size: 180 k
Installed size: 565 k
```

packaging/leapp.spec
```diff
@@ -25,7 +25,8 @@ Source0:    https://github.com/leapp-to/%{name}/archive/%{gittag}/%{name}-%{vers
 BuildArch:  noarch
 %if %{with python3}
 Requires: python3-%{name} = %{version}-%{release}
-%else
+%endif
+%if %{with python2}
```
pirat89: Hmm. Missing check for %fedora here. As this package will not be created on fedora now, it should be told, that this part is not relevant  for fedora `!0%{?fedora}` ... Another thing is that you are telling that the RPM can be ok with Py3 and Py2 but snactor can by build only for Py2 or Py3.

Another thing is, that you are telling that in case you build this for Python3 and Python2 - both has to be installed. Make the dependencies exclusive or let's discuss what are options. I am thinking here more about weak deps or rich dependencies to resolve the problem otherwise.

vinzenz: This change indeed is a mistake. It should have been done for snactor.
/bin/leapp and /bin/snactor actually would be python3 because they are built last. 

I wanted to have both, actually since you need a way to run the tests for both. Python3 and Python2 and then you'd want to have both frameworks.

vinzenz: I made the python2 / python3 part now a bit more obvious by creating snactor2 and snactor3 just in case.

packaging/leapp.spec
```diff
@@ -138,7 +136,9 @@ BuildRequires:  python3-setuptools
 BuildRequires:  python3-pytest-cov
 %endif
 
-Requires: leapp-framework-dependencies = 1
+Requires: python3-setuptools
```
vinzenz: Just to make it clear - We don't need the deps here because those are already defined as Python3, the whole problem we originally had was python-something or something-python vs python2-something this does not account for python3-something since they already used the right naming.

pirat89: @vinzenz I don't agree. That'S part of the problem. Another issue - that is not there know, but can be in future - is about changes of dependencies between original & target system. Nowhere is written, that you will not need additional dependency in the target system, e.g. because the python3-somethin package will be split into several other packages. Another point is, that this is against our documentation (not sure whether already both  parts of doc are merged) as well. We declared, that all outside dependencies should be defined instide the deps metapackage. I think we really should not care about whether the problem is relevant now for Python3. Just keep the logic as it was supposed to be used originally - or improve it. But this is basically wrong change. - the same for findutils and the others. These are *outer dependencies*.

pirat89: @vinzenz in case you want another discussion, I prefer call and then we can write here result of the discussion :)

pirat89:
leapp-ci build

packaging/leapp.spec
```diff
@@ -21,15 +21,16 @@ Summary:    OS & Application modernization framework
 License:    ASL 2.0
 URL:        https://leapp-to.github.io
 Source0:    https://github.com/leapp-to/%{name}/archive/%{gittag}/%{name}-%{version}.tar.gz
-
 BuildArch:  noarch
+
+%if !0%{fedora}
```
pirat89: Eror: /tmp/temp/packaging/leapp.spec:26: bad %if condition

```suggestion
%if !0%{?fedora}
```

packaging/leapp.spec
```diff
@@ -169,17 +170,21 @@ Python 3 leapp framework libraries.
 # Install
 ##################################################
 %install
+
+install -m 0755 -d %{buildroot}%{_mandir}/man1
+install -m 0644 -p man/snactor.1 %{buildroot}%{_mandir}/man1/
+
+%if !0%{fedora}
```
pirat89:
```suggestion
%if !0%{?fedora}
```

packaging/leapp.spec
```diff
@@ -189,9 +194,16 @@ install -m 0644 -p man/snactor.1 %{buildroot}%{_mandir}/man1/
 %py3_install
 %endif
 
+%if 0%{fedora}
```
pirat89:
```suggestion
%if 0%{?fedora}
```

packaging/leapp.spec
```diff
@@ -189,9 +194,16 @@ install -m 0644 -p man/snactor.1 %{buildroot}%{_mandir}/man1/
...
+rm -f %{buildroot}/%{_bindir}/leapp
+%endif
+
+
 ##################################################
 # leapp files
 ##################################################
+
+%if !0%{fedora}
```
pirat89:
```suggestion
%if !0%{?fedora}
```

PR #443, opened 2019-02-18
Fix indentation and add "--debug" to show output

Rezney:


docs/source/first-actor.md
```diff
@@ -114,14 +114,14 @@ from leapp.actors import Actor
 
 
 class HostnameScanner(Actor):
-     name = 'hostname_scanner'
-     description = 'No description has been provided for the hostname_scanner actor.'
-     consumes = ()
-     produces = ()
-     tags = ()
+    name = 'hostname_scanner'
+    description = 'No description has been provided for the hostname_scanner actor.'
```
pirat89: We want to use dosctrings nowadays

pirat89: We want to use docstring nowadays.

PR #445, opened 2019-02-19
New reporting

Rezney:
This is WIP PR for introducing new reporting models and report command.

Comments are highly appreciated.

Require Repo PR: oamg/leapp-repository/pull/86

### DIFF
--- a/leapp/cli/__init__.py
+++ b/leapp/cli/__init__.py
@@ -2,7 +2,11 @@ import os
 import socket
 
 from leapp.utils.clicmd import command, command_opt
-import leapp.cli.upgrade
+from leapp.cli import upgrade
+try:
+    from leapp.cli import report
+except ImportError:
+    report = None
 from leapp import VERSION
 
 
@@ -19,5 +23,8 @@ def cli(args):
 
 def main():
     os.environ['LEAPP_HOSTNAME'] = socket.getfqdn()
-    cli.command.add_sub(leapp.cli.upgrade.upgrade.command)
+    if report:
+        cli.command.add_sub(report.report.command)
+    cli.command.add_sub(upgrade.list_runs.command)
+    cli.command.add_sub(upgrade.upgrade.command)
     cli.command.execute('leapp version {}'.format(VERSION))
--- a/leapp/cli/report/__init__.py
+++ b/leapp/cli/report/__init__.py
@@ -1,0 +1,42 @@
+"""
+This module is currently single workflow specific (upgrades). After introducing other use cases we can make it
+more robust
+"""
+
+from __future__ import print_function
+import os
+import sys
+
+from leapp.exceptions import CommandError
+from leapp.utils.clicmd import command, command_opt, command_arg
+from leapp.utils.report import fetch_upgrade_report_messages
+from leapp.cli.upgrade import fetch_last_upgrade_context
+
+
+@command('report', help='Create report for last upgrade execution (default) or specific execution id')
+@command_opt('--id', help='ID of particular run to print report for. See # leapp list-runs for a list of all runs')
+@command_opt('--format', help='Format report using particular renderers. By default: "plaintext"')
+def report(args):
+    if os.getuid():
+        raise CommandError('This command has to be run under the root user.')
+
+    id = args.id or fetch_last_upgrade_context()[0]
+
+    if not id:
+        raise CommandError(
+            'No previous Leapp upgrade run found. This command can only be run after "leapp upgrade" has been executed'
+        )
+
+    # FIXME: add proper default and choices parameters for the format option
+    if args.format not in ('html', 'plaintext'):
+        print('No / not supported renderer provided as --format (available renderers: "html", "plaintext"). '
+              'Defaulting to "plaintext"', file=sys.stdout, end='\n\n')
+        renderer = 'plaintext'
+    else:
+        renderer = args.format
+
+    messages = fetch_upgrade_report_messages(id, renderer)
+    if not messages:
+        raise CommandError('No upgrade report messages found for context {}'.format(id))
+
+    print(messages, file=sys.stdout)
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -1,0 +1,103 @@
+from leapp.models import Model, fields
+from leapp.topics import ReportTopic
+from leapp.libraries.stdlib.api import produce
+import json
+
+
+class Renderers(Model):
+    """
+    Model for describing renderers to be used when presenting reports.
+    """
+
+    topic = ReportTopic
+
+    html = fields.String()
+    """
+    Render report message as html
+    """
+    plaintext = fields.String()
+    """
+    Render report message as plaintext
+    """
+
+
+class Report(Model):
+    """
+    Framework model used for reporting and presentation (see "renderers" field) purposes. The report can also carry
+    a special meaning using "flags" field.
+    """
+
+    topic = ReportTopic
+
+    severity = fields.StringEnum(choices=['low', 'medium', 'high'])
+    """
+    Severity of the report entry
+    """
+
+    title = fields.String()
+    """
+    Title of the report entry
+    """
+
+    detail = fields.JSON()
+    """
+    Detail of the report entry as JSON data
+    """
+
+    renderers = fields.Model(Renderers)
+    """
+    :class:`Renderers` describe how to render this report entry
+    """
+
+    audience = fields.List(fields.StringEnum(choices=['developer', 'sysadmin']))
+    """
+    Who is the main audience of this report entry
+    """
+
+    flags = fields.List(fields.StringEnum(choices=['inhibitor']))
+    """
+    Flags can give a special meaning to each report entry, currently supported flags:
+
+        inhibitor - This report entry will inhibit the upgrade process
+    """
+
+
+def report(title=None, detail=None, renderers=None, audience=None, flags=None, severity=None):
+    """
+    Create and produce a report entry
+
+    For more information about the arguments, please refer to the :class:`Report` model
+
+    :param title: Title of the report message
+    :type title: str
+    :param detail: Report message details
+    :type detail: dict
+    :param renderers: Available report entry renderers (e.g. html / plaintext)
+    :type renderers: dict
+    :param audience: Target audience of the report
+    :type audience: list
+    :param flags: Functionality flags (e.g. inhibitor)
+    :type flags: list
+    :param severity: Report severity
+    :type severity: str
+    """
+
+    # set some healthy defaults
+    audience = audience or ['sysadmin']
+    severity = severity or 'medium'
+    flags = flags or []
+
+    renderers = Renderers(**renderers)
+    report_entry = {
+        'title': title,
+        'detail': detail,
+        'renderers': renderers,
+        'severity': severity,
+        'flags': flags,
+        'audience': audience,
+    }
+
+    if 'title' in report_entry['detail']:
+        raise ValueError('Key "title" cannot be present in the report "detail" structure')
+
+    produce(Report(**report_entry))
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -27,6 +27,8 @@ class Report(Model):
     a special meaning using "flags" field.
     """
 
+    __non_inheritable__ = True
+
     topic = ReportTopic
 
     severity = fields.StringEnum(choices=['low', 'medium', 'high'])
--- a/leapp/utils/report.py
+++ b/leapp/utils/report.py
@@ -1,0 +1,53 @@
+import json
+import six
+
+from jinja2 import Template
+
+from leapp.utils.audit import get_messages
+
+
+def _flatten_dict(dict_):
+    new_dict = {}
+    if type(dict_) is not dict:
+        return new_dict
+
+    for k, v in six.iteritems(dict_):
+        if type(v) == dict:
+            new_dict.update(_flatten_dict(v))
+        else:
+            new_dict[k] = v
+
+    return new_dict
+
+
+def fetch_upgrade_report_messages(context_id, renderer):
+    """
+    :param context_id: ID to identify the needed messages
+    :type context_id: str
+    :param renderer: How messages will be rendered (e.g html or plaintext)
+    :type context_id: str
+    :return: All upgrade messages of type "Report" withing the given context
+    """
+    report_msgs = get_messages(names=['Report'], context=context_id)
+
+    if report_msgs:
+        messages_to_print = []
+        for message in report_msgs:
+            topic = message.get('topic')
+            phase = message.get('phase')
+            data = json.loads(message.get('message').get('data'))
+            merged_data = {'topic': topic, 'phase': phase}
+            merged_data.update({
+                'audience': data['audience'],
+                'title': data['title'],
+                'flags': data['flags'],
+                'severity': data['severity']
+            })
+            detail = _flatten_dict(json.loads(data['detail']))
+            merged_data.update(detail)
+            template = Template(data['renderers'].get(renderer))
+            messages_to_print.append(template.render(merged_data))
+
+        return messages_to_print
+
+    return None
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import itertools
 import json
 import os
@@ -36,6 +37,19 @@ def fetch_last_upgrade_context():
         if row:
             return row[0], json.loads(row[2])
     return None, {}
+
+
+def fetch_all_upgrade_contexts():
+    """
+    :return: All upgrade execution contexts
+    """
+    with get_connection(None) as db:
+        cursor = db.execute(
+            "SELECT context, stamp, configuration FROM execution WHERE kind = 'upgrade' ORDER BY stamp DESC")
+        row = cursor.fetchall()
+        if row:
+            return row
+    return None
 
 
 def get_last_phase(context):
@@ -103,3 +117,14 @@ def upgrade(args):
         workflow.run(context=context, skip_phases_until=skip_phases_until)
 
     report_errors(workflow.errors)
+
+
+@command('list-runs', help='List previous Leapp upgrade executions')
+def list_runs(args):
+    contexts = fetch_all_upgrade_contexts()
+    if contexts:
+        for context in contexts:
+            print('Context ID: {} - time: {} - details: {}'.format(context[0], context[1], json.loads(context[2])),
+                  file=sys.stdout)
+    else:
+        raise CommandError('No previous run found!')
--- a/leapp/models/__init__.py
+++ b/leapp/models/__init__.py
@@ -42,7 +42,10 @@ class ModelMeta(type):
     """
     def __new__(mcs, name, bases, attrs):
         klass = super(ModelMeta, mcs).__new__(mcs, name, bases, attrs)
-
+        for base in bases:
+            if getattr(base, '__non_inheritable__', False):
+                raise TypeError('{cls} cannot inherit from {base} because {base} is not inheritable'.format(
+                    cls=name, base=base.__name__))
         model_ref_cls = globals().get('_ModelReference')
         if name == '_ModelReference' or (model_ref_cls and issubclass(klass, model_ref_cls)):
             setattr(sys.modules[mcs.__module__], name, klass)
--- a/leapp/topics/__init__.py
+++ b/leapp/topics/__init__.py
@@ -36,6 +36,13 @@ class ErrorTopic(Topic):
     name = 'errors'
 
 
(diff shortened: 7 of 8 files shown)


leapp/cli/report/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    :return: All upgrade messages of type "LeappReport" withing the given context
+    """
+    with get_connection(None) as db:
+        cursor = db.execute(
+            "SELECT message_data_hash FROM message WHERE type = 'CheckResult' and context = '%s'" % context_id)
+        reports = cursor.fetchall()
+        report_msg_ids = [report[0] for report in reports]
+        cursor = db.execute(
+            "SELECT data FROM message_data WHERE hash in (%s)" % ','.join('?'*len(report_msg_ids)), report_msg_ids)
```
Rezney: Need to find better way here as sqlite will accept only up to 999 hash parameters I guess

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,193 @@
+from leapp.models import Model, fields
+from leapp.topics import Topic
+
+
+class ReportTopic(Topic):
+    name = 'report_topic'
```
pirat89: do not forget to move this into the separate file in the `leapp/topics` directory.

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,193 @@
...
+
+"""
+Detail requirements:
+    summary : str
+
+Example:
+    {'summary': 'Some text here'}
+"""
+RenderGeneric = {
+    'html': '<h1 class="title">{ title }</h1><p class="summary">{ summary }</p>',
```
pirat89: I would use here probably h2 in all cases, because I guess that al those messages will be printed in one page that would contain `h1` header in the top of the page.

```suggestion
    'html': '<h2 class="title">{ title }</h2><p class="summary">{ summary }</p>',
```

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,193 @@
...
+"""
+Detail requirements:
+    summary : str
+    remediation : str
+
+Example:
+    {'summary': 'Here is a summary', 'remediation': 'and --here-is -a solution'}
+"""
+RenderWithRemediation = {
+    'html': '<h1 class="title">{ title }</h1><p class="summary">{ summary }</p><h2>Remediation:</h2><pre class="code">{ remediation }</pre>',
```
pirat89:
```suggestion
    'html': '<h2 class="title">{ title }</h2><p class="summary">{ summary }</p><h3>Remediation:</h3><pre class="code">{ remediation }</pre>',
```

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,199 @@
...
+class Renderers(Model):
+    topic = ReportTopic
+
+    html = fields.String()
+    plaintext = fields.String()
+
+
+class Report(Model):
+    
+    __metaclass__ = DoNotInherit
```
Rezney: @vinzenz What would be the correct way to forbid inheriting from Report model?

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+     'links': [{'href': 'https://some.important.link.com', 'title': 'Knowledge base article'}]}
+"""
+RenderWithLinks = {
+    'html': RenderGeneric['html'] +  # noqa: E131,W504
+             ('<ul class="report-list">{{% for link in links %}}'
+              '<li class="report-list-item">'
+              ' <a href="{{ link.href }}" target="_blank" class="report-link">{{ link.title }}</a>'
+              '</li>'
+              '{{% endfor %}}</ul>'),
+    'plaintext': '{{ title }}\n{{ summary }}\nLinks:\n{{% for link in links %}}{{ link.title }} - {{ link.href }}{{% endfor %}}'
```
shaded-enmity:
```suggestion
    'plaintext': ('{{ title }}\n'
                  '{{ summary }}\n'
                  'Links:\n'
                  '{{% for link in links %}}{{ link.title }} - {{ link.href }}{{% endfor %}}')
```

leapp/cli/__init__.py
```diff
@@ -3,6 +3,7 @@
 
 from leapp.utils.clicmd import command, command_opt
 import leapp.cli.upgrade
+import leapp.cli.report
```
vinzenz: since we talked about having this report optional, we'd have to surround this by try/except

leapp/cli/__init__.py
```diff
@@ -14,5 +15,6 @@ def cli(args):
 
 def main():
     os.environ['LEAPP_HOSTNAME'] = socket.getfqdn()
+    cli.command.add_sub(leapp.cli.report.report.command)
```
vinzenz: same, try/except need

leapp/cli/report/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+        template = Template(data['renderers'].get(renderer))
+        context = merged_data['detail'].copy()
+        context.update(merged_data['named'])
+
+        messages_to_print.append(template.render(context))
+
+    print(messages_to_print)
+
+
+@command('list-runs', help='List previous Leapp upgrade executions')
```
vinzenz: not registered in leapp/cli/__init__.py with cli.command.add_sub

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+class Report(Model):
+
+    __non_inheritable__ = True
+
+    topic = ReportTopic
+
+    """
+    Version of framework and actors as :class:`ReportVersions`
+    """
+    versions = fields.Model(ReportVersions)
```
vinzenz: No sense for having the version of the framework and actors included in each report. You rather should have one ReportVersions per execution. It's just a field of noise IMHO

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    detail = fields.String()
+
+    """
+    Named variables, represented as JSON data, that can be used during rendering
+
+    Example:
+        named = {'summary_with_exclamations': '!!!!! { summary } !!!!'}
+    """
+    named = fields.Nullable(fields.String())
```
vinzenz: I thought 'detail' was supposed to be just that? I'm a bit confused about this field.

shaded-enmity: I just realized that we're not processing the field correctly. Nevertheless, the main difference should be:
1) `detail` is not jinja2 processed
2) `named` is jinja2 processed
I.e. in `detail` I don't need to worry about escaping possible embedded jinja2 strings, it remains static.
I'm fine with removing `named` and reintroduce it when it's needed as this is not really crucial for the enablement.

leapp/cli/report/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+from leapp.utils.report import fetch_upgrade_report_messages
+from leapp.cli.upgrade import fetch_last_upgrade_context
+
+from jinja2 import Template
+import json
+
+
+@command('report', help='Create report for last upgrade execution (default) or specific execution id using --id ID')
+@command_opt('id')
+@command_opt('--format')
```
jmikovic: commands are a bit inconsistent, use --id

leapp/cli/report/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+    if not id:
+        raise CommandError('No previous Leapp upgrade run of the requested context {} found'.format(id))
+
+    messages = fetch_upgrade_report_messages(id)
+    if not messages:
+        raise CommandError('No upgrade report messages found for context {}'.format(id))
+
+    # FIXME: add proper default and choices parametrs for the format option
+    if args.format not in ('html', 'plaintext'):
+        print('No / not supported renderer provided as --format (available renderers: "html", "plaintext"). '
```
fernflower: Please use logging instead of print

vinzenz: @fernflower For starting to use logging, one would have to setup logging first. This command hasn't it set up just yet. So probably not too bad.

leapp/cli/report/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+
+        for key in ('detail', 'named'):
+            if key in merged_data:
+                raise ValueError('Duplicate key {} in report structure'.format(key))
+            merged_data[key] = json.loads(data.get(key))
+
+        if not merged_data['named']:
+            merged_data['named'] = {}
+
+        if 'title' in merged_data['named'] or 'title' in merged_data['detail']:
```
vinzenz: I don't know but this feels to be the wrong place to check for this. We should rather think of additional validators for the models instead of making a check like this here.

leapp/cli/report/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+        renderer = args.format
+
+    messages_to_print = []
+    for message in messages:
+        topic = message.get('topic')
+        phase = message.get('phase')
+        data = json.loads(message.get('message').get('data'))
+        merged_data = {'topic': topic, 'phase': phase}
+
+        for key in ('detail', 'named'):
```
vinzenz: I don't know but this feels to be the wrong place to check for this. We should rather think of additional validators for the models instead of making a check like this here.

vinzenz:
Just FYI those doc strings you are specifying in the Report model aren't correctly placed. Docstrings are supposed to follow after the item they are referring to.

Example: 
```python
class Foobar(object):
    attribute = fields.Something()
    """ Docstring follows here """
```

Your documentation is the other way around, it is before the statement that introduces it.

leapp/cli/report/__init__.py
shaded-enmity: We actually thought about having the `id` as positional argument. Makes most sense IMHO.

leapp/cli/report/__init__.py
shaded-enmity: While totally valid point, I think this is outside of the scope of this PR. Model data validation would be nice topic for cabal.

vinzenz: @shaded-enmity it was just a note, I am not blocking you on this.

vinzenz: Then thinking about it, it could be actually checked in the report function itself, no?

leapp/cli/__init__.py
Rezney: Added try except, optional part will be done in different PR...

leapp/reporting/__init__.py
Rezney: We got rid of versioning completely. Initially we were thinking about generating it in specfile but we would not have it upstream then...

leapp/cli/report/__init__.py
vinzenz: @shaded-enmity @Rezney 
positional arguments use  @command_arg not @command_opt, just FYI

leapp/cli/report/__init__.py
shaded-enmity: yup i like that more, we’ll move this (and the other) validation to the ‘report’ func

leapp/cli/report/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+import os
+import sys
+
+from leapp.exceptions import CommandError
+from leapp.utils.clicmd import command, command_opt, command_arg
+from leapp.utils.report import fetch_upgrade_report_messages
+from leapp.cli.upgrade import fetch_last_upgrade_context
+
+
+"""
```
vinzenz: A module doc string needs to be on the top of the module.

leapp/cli/__init__.py
```diff
@@ -3,6 +3,10 @@
 
 from leapp.utils.clicmd import command, command_opt
 import leapp.cli.upgrade
+try:
+    import leapp.cli.report
```
vinzenz:
```suggestion
    from leapp.cli import report
```

leapp/cli/__init__.py
```diff
@@ -14,5 +18,8 @@ def cli(args):
 
 def main():
     os.environ['LEAPP_HOSTNAME'] = socket.getfqdn()
+    if report:
+        cli.command.add_sub(leapp.cli.report.report.command)
```
vinzenz:
```suggestion
        cli.command.add_sub(report.report.command)
```

leapp/cli/__init__.py
```diff
@@ -14,5 +18,8 @@ def cli(args):
...
+        cli.command.add_sub(leapp.cli.report.list_runs.command)
```
vinzenz:
```suggestion
        cli.command.add_sub(report.list_runs.command)
```

leapp/cli/report/__init__.py
Rezney: Sure, got confused by: 

SyntaxError: from __future__ imports must occur at the beginning of the file

but then one see:

https://docs.python.org/2/reference/simple_stmts.html#future

leapp/cli/report/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+
+from leapp.exceptions import CommandError
+from leapp.utils.clicmd import command, command_opt, command_arg
+from leapp.utils.report import fetch_upgrade_report_messages
+from leapp.cli.upgrade import fetch_last_upgrade_context
+
+
+@command('report', help='Create report for last upgrade execution (default) or specific execution id')
+@command_arg('id')
+@command_opt('--format')
```
vinzenz: At least format, should be receiving some help string.

Rezney: Definitely. Thanks for pointing this out.

Rezney:
@vinzenz @pirat89 @shaded-enmity anything else we are missing for this first round? If not let's do the final testing. With this PR in, testing the repository PR will be much easier...

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+
+        inhibitor - This report entry will inhibit the upgrade process
+    """
+
+
+def report(title=None, detail=None, renderers=None, audience=None, flags=None, severity=None):
+    """
+    Create and produce a report entry
+
+    For more information about the arguments, please refer to the :class:`Report` model
```
vinzenz: I am missing the documentation for the parameters here
```
:param title: Yadda yadda
:type title: str
```

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    # set some healthy defaults
+    audience = audience or ['sysadmin']
+    severity = severity or 'medium'
+    flags = flags or []
+    # if there is an inhibitor flag, the severity is always high
+    if 'inhibitor' in flags and severity != 'high':
+        raise ValueError('Ihibitors must have high severity')
```
vinzenz:
```suggestion
        raise ValueError('Inhibitors must have high severity')
```

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+        'title': title,
+        'detail': detail,
+        'renderers': renderers,
+        'severity': severity,
+        'flags': flags,
+        'audience': audience,
+    }
+
+    if 'title' in report_entry['detail']:
+        raise ValueError('Key "{}" cannot be present in the "detail" structure'.format('title'))
```
vinzenz:
```suggestion
        raise ValueError('Key "title" cannot be present in the "detail" structure')
```

leapp/utils/report.py
```diff
@@ -0,0 +1,53 @@
+import json
+import six
+
+from jinja2 import Template
+
+from leapp.utils.audit import get_messages
+
+
+def flatten_dict(dict_):
```
vinzenz: document this or make it private

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+    Create and produce a report entry
+
+    For more information about the arguments, please refer to the :class:`Report` model
+    """
+    # set some healthy defaults
+    audience = audience or ['sysadmin']
+    severity = severity or 'medium'
+    flags = flags or []
+    # if there is an inhibitor flag, the severity is always high
+    if 'inhibitor' in flags and severity != 'high':
```
shaded-enmity: I don't think inhibition necessarily means high severity.

leapp/utils/report.py
```diff
@@ -0,0 +1,53 @@
...
+            merged_data.update({
+                'audience': data['audience'],
+                'title': data['title'],
+                'flags': data['flags'],
+                'severity': data['severity']
+            })
+            detail = flatten_dict(json.loads(data['detail']))
+            merged_data.update(detail)
+            template = Template(data['renderers'].get(renderer))
+        messages_to_print.append(template.render(merged_data))
```
vinzenz: Either this should have been within the loop, or messages_to_print is a useless variable.

```suggestion
            messages_to_print.append(template.render(merged_data))
```

leapp/reporting/__init__.py
Rezney: Hehe, of course :D

leapp/utils/report.py
Rezney: Mistake during refactoring. Thanks for pointing this out.

leapp/reporting/__init__.py
Rezney: We got rid of this completely...

Rezney:
@vinzenz all your comments should be hopefully addressed. @bocekm @jmikovic could you please test it? I just did and seems to be working fine. Let me know you need any details.

jmikovic:
can we set ID for leapp report to last run, leapp list-runs show only last one anyways

Rezney:
@jmikovic ahh yep, changing the command from opt to arg broke this. @vinzenz is it currently possible to provide nargs='?' to the parser?

Rezney:
@jmikovic @vinzenz @shaded-enmity 
Anyway I guess optional ID with last run as default can wait till other iteration. I will get rid of the defaulting as id has to provided anyway.

Apart from that I also moved "list-runs" command to upgrade __init__ as it fits better there , me thinks.

leapp/cli/report/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+import sys
+
+from leapp.exceptions import CommandError
+from leapp.utils.clicmd import command, command_opt, command_arg
+from leapp.utils.report import fetch_upgrade_report_messages
+from leapp.cli.upgrade import fetch_last_upgrade_context
+
+
+@command('report', help='Create report for last upgrade execution (default) or specific execution id')
+@command_arg('id', help='ID of particular execution to print report for. Can be obtain by #leapp list-runs')
```
shaded-enmity:
```suggestion
@command_arg('id', help='ID of particular run to print report for. See # leapp list-runs for a list of all runs')
```

vinzenz:
@Rezney Well no it's not possible, but then again, if you want it optional, then don't make it a positional  argument but an option that allows you to specify an ID, otherwise it will be last.

leapp/cli/report/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+@command_opt('--id', help='ID of particular run to print report for. See # leapp list-runs for a list of all runs')
+@command_opt('--format', help='Format report using particular renderers. By default: "plaintext"')
+def report(args):
+    if os.getuid():
+        raise CommandError('This command has to be run under the root user.')
+
+    id = args.id or fetch_last_upgrade_context()[0]
+
+    if not id:
+        raise CommandError('No previous Leapp upgrade run of the requested context {} found'.format(id))
```
vinzenz: That would give you the message:

> No previous Leapp upgrade run of the requested context None found

Maybe better something along these lines:
> No previous Leapp upgrade run found. This command can only be run after 'leapp upgrade' has been executed.

leapp/reporting/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class Renderers(Model):
+    topic = ReportTopic
+
+    html = fields.String()
+    plaintext = fields.String()
+
+
+class Report(Model):
```
vinzenz: For this and above, we would like now to use the docstrings too, to describe what the Model is used for and what it represents. I am fine with a follow up PR however please consider doing it.

leapp/cli/report/__init__.py
Rezney: you are right...

leapp/reporting/__init__.py
Rezney: Above is "just" html and plaintext fields. Did I get it right or am missing something?

vinzenz: I am talking about the model class docstrings - What the model is used for.

Rezney: done

Rezney:
Let's wait with the merge until we have python babel pkg in PES data.

Rezney:
@vinzenz are your requested changes still up-to-date?

pirat89:
As discussed with @Rezney, the blocking rq-changes from @vinzenz has been kept to ensure we will merge it after we the related PR in leapp-repository will be prepared for merge as well. That's it, so removing the request and going to merge that.

PR #447, opened 2019-02-19
Add link in spec file for the dependencies documentation

asilveir:


packaging/leapp.spec
```diff
@@ -1,3 +1,8 @@
+# IMPORTANT:
+# Please read the documentation how to deal with the new dependencies
```
bocekm:
```suggestion
# Please read the documentation on how to deal with dependencies before adding a new one.
```

PR #451, opened 2019-02-25
Print a warning msg when missing model in "produces" tuple

Rezney:
Recently we spent some time with debugging "issue" with missing a model in produces tuple with @jmikovic and @shaded-enmity. This should help to save some time of actor developers.

Am I missing anything in this fix @vinzenz ? It does not have to be "AssertionError" of course.


Rezney:
@jmikovic the message has been changed as required

Issue #456, opened 2019-03-04
[snactor] Trace upon creating topic outside repository


fernflower:
**Actual behavior**
If topic creation command is executed outside leapp repository a following stack trace is seen.
```Traceback (most recent call last):
  File ".venv/bin/snactor", line 10, in <module>
    sys.exit(main())
  File "/home/ivasilev/projects/leapp/.venv/lib/python3.7/site-packages/leapp/snactor/__init__.py", line 67, in main
    cli.command.execute(version=_('snactor version {}').format(VERSION))
  File "/home/ivasilev/projects/leapp/.venv/lib/python3.7/site-packages/leapp/utils/clicmd.py", line 80, in execute
    args.func(args)
  File "/home/ivasilev/projects/leapp/.venv/lib/python3.7/site-packages/leapp/utils/clicmd.py", line 102, in called
    self.target(args)
  File "/home/ivasilev/projects/leapp/.venv/lib/python3.7/site-packages/leapp/snactor/commands/new_topic.py", line 24, in cli
    basedir = os.path.join(basedir, 'topics')
  File "/home/ivasilev/projects/leapp/.venv/lib64/python3.7/posixpath.py", line 80, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

**To Reproduce**
1. from any non-repo dir call snactor new-topic NewTopic

vinzenz:
Yeah the command is missing the @requires_repository decorator

PR #457, opened 2019-03-04
snactor: fixed trace on topic creation

fernflower:
If a topic creation is attempted outside leapp repository
directory it will fail with a trace during os.path.join.
This fixes the issue by validating that command execution
takes place in the actual leapp repo directory.

Closes-Issue:#456

### DIFF
--- a/tests/scripts/test_snactor.py
+++ b/tests/scripts/test_snactor.py
@@ -1,8 +1,9 @@
 import json
 import os
-from subprocess import check_call, check_output, CalledProcessError
+from subprocess import check_call, check_output, CalledProcessError, STDOUT
 
 from helpers import repository_dir
+from leapp.exceptions import CommandError
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


tests/scripts/test_snactor.py
```diff
@@ -30,6 +31,16 @@ def test_discovery(repository_dir):
...
 
 
+def test_raises_create_outside_repository():
+    for cmd in ([['snactor', 'new-tag', 'winteriscoming'],
+                 ['snactor', 'new-topic', 'winteriscoming'],
+                 ['snactor', 'new-model', 'winteriscoming', '--topic', 'WinteriscomingTopic']]):
+        with pytest.raises(CalledProcessError) as err:
+            check_output(cmd, stderr=STDOUT)
+        assert ('This command must be executed from the repository directory' in
+                err.value.stdout.decode('utf-8'))
```
vinzenz: From the CI:
```AttributeError: 'CalledProcessError' object has no attribute 'stdout'```

This is due to python 3 introducing those fields and the code must be Python 2 compatible.

fernflower: my bad, should be fixed by using value.ouput

PR #458, opened 2019-03-06
model.fields: Introduction of the JSON field type

vinzenz:
The JSON field type, is a fields that allows to assign any value that is
encodable to JSON by the json library of the python standard library.
With the exception for None value as the value itself. This is to keep
consistency with the rest of the fields, that require the field to be
marked as nullable. None value nested within, however is not restricted
by this limitation.

Example usage:
```python
class JsonUsedModel(Model):
    topic = SomeTopic

    data = fields.JSON()
    nullable_data = fields.Nullable(fields.JSON())

m = JsonUsedModel()

m.data = {'valid': None} # valid value
m.data = None # invalid value for this field

m.nullable_data = None # valid
m.nullable_data = {'valid': None} # valid value
```

### DIFF
--- a/leapp/models/fields/__init__.py
+++ b/leapp/models/fields/__init__.py
@@ -1,5 +1,7 @@
 import copy
 import datetime
+import json
+
 import six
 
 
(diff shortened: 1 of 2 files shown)


artmello:
Would be possible/interesting to add tests for passing None to nullable/non-nullable models? Other than that LGTM too.

vinzenz:
@artmello actually there are checks for it :-) If you look for the as_nullable() usage in the serialization tests you can see the checks. I just added more field type tests here which was sufficient

PR #465, opened 2019-03-25
leapp: Fix debug/verbose restore on resume

vinzenz:
Previously, when starting leapp with debug or verbose, applied to
commandline and environment variable enablement, the information was no
longer respected after resuming.

The problem has been incorrect handling of variables checking. This
patch fixes the problem and moves out the check into a function.

### DIFF
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -58,6 +58,13 @@ def get_last_phase(context):
         return checkpoints[-1]['phase']
 
 
+def check_env_and_conf(env_var, conf_var, configuration):
+    """
+    Checks whether the given environment variable or the given configuration value are set to '1'
+    """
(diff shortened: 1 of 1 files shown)


leapp/cli/upgrade/__init__.py
```diff
@@ -58,6 +58,13 @@ def get_last_phase(context):
         return checkpoints[-1]['phase']
 
 
+def check_env_and_conf(env_var, conf_var, configuration):
+    """
+    Checks whether the given environment variable or the given configuration value are set to '1'
+    """
+    return os.getenv('env_var', '0') == '1' or configuration.get(conf_var, '0') == '1'
```
asilveir: env_var is surrounded commas.

jmikovic:
debug is working but verbose doesn't

PR #466, opened 2019-03-25
Change default loglevel to ERROR

asilveir:
When LEAPP_DEBUG is not 1, self.log.error() and self.log.critical() won't print anything to stderr.

This should be true for lower severities, but from error up, things should appear also on stderr.

### DIFF
--- a/leapp/logger/__init__.py
+++ b/leapp/logger/__init__.py
@@ -5,6 +5,7 @@ import os
 import time
 import sys
 
+from leapp.libraries.stdlib.config import is_debug, is_verbose
 from leapp.utils.audit import Audit
 _logger = None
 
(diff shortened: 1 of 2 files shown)


pirat89:
I am missing the reason for this change. It is not in the PR description, neither in commit msg.

asilveir:
> I am missing the reason for this change. It is not in the PR description, neither in commit msg.

He has talked about this in the previous Cabal, It's a bug 1408 in Jira.

AloisMahdal:
I just re-ran the downstream test suite with *leapp-0.5.0-0.201903261646Z.08a3824.loglevel_error.PR466.el7_6.noarch*.

I can confirm this does fix the bug.

leapp/logger/__init__.py
```diff
@@ -53,22 +54,29 @@ def configure_logger():
...
             logging.getLogger('leapp').addHandler(handler)
+            logging.StreamHandler().setLevel(logging.ERROR)
 
-        if os.getenv('LEAPP_VERBOSE', '0') == '1':
-            logging.getLogger('').setLevel(logging.DEBUG if os.getenv('LEAPP_DEBUG', '0') == '1' else logging.INFO)
-            handler = logging.StreamHandler(sys.stderr)
-            handler.setFormatter(logging.Formatter(fmt=log_format, datefmt=log_date_format))
-            logging.getLogger('').addHandler(handler)
+        if is_verbose():
+            for handler in logging.getLogger().handlers:
```
vinzenz: you set for all handlers the log level here, that means you do not respect the fact that the audit log still should contain DEBUG data as we discussed it on the cabal.

vinzenz:
@AloisMahdal unfortunately it should fail the matrix in the way that DEBUG is not enabled all the time for the leapp handler.

PR #467, opened 2019-03-27
stdlib: remove the call function

pirat89:
We want to get rid of the subprocess library and we found the call
function as expendable as we want from people to use just `run` for
everything.

### DIFF
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -75,30 +75,6 @@ class CalledProcessError(LeappError):
         return self._result.get('pid')
 
 
-def call(args, split=True):
-    """
-    Call an external program, capture and automatically utf-8 decode its output.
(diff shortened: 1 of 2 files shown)


pirat89:
I will wait with writting of the tests as we want to discuss this solution yet. Maybe we will get rid of the `call` completely.

leapp/libraries/stdlib/__init__.py
```diff
@@ -87,16 +87,8 @@ def call(args, split=True):
...
 
-    r = None
-    with open(os.devnull, mode='w') as err:
-        if six.PY3:
-            r = subprocess.check_output(args, stderr=err, encoding='utf-8')
-        else:
-            r = subprocess.check_output(args, stderr=err).decode('utf-8')
-    if split:
-        return r.splitlines()
-    return r
```
asilveir: Too many blank lines

PR #471, opened 2019-03-29
Show information in debug mode about external commands

asilveir:
As in debug mode we have a lot of messages the user needs to be informed which external command has generated certain output.

This PR wants to solve this by adding this output:

```
leapp.workflow.Download.prepare_upgrade_transaction: External command is started: [/usr/bin/dnf rhel-upgrade]
*** external command output ***
leapp.workflow.Download.prepare_upgrade_transaction: External command is finished: [/usr/bin/dnf rhel-upgrade]
```

### DIFF
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -15,6 +15,7 @@ from leapp.exceptions import LeappError
 from leapp.utils.audit import create_audit_entry
 from leapp.libraries.stdlib import api
 from leapp.libraries.stdlib.call import _call, STDOUT
+from leapp.libraries.stdlib.config import is_debug
 
 
 class CalledProcessError(LeappError):
@@ -111,7 +112,7 @@ def _logging_handler(fd_info, buffer):
     :type buffer: bytes array
     """
     (_unused, fd_type) = fd_info
-    if os.getenv('LEAPP_VERBOSE', '0') == '1':
+    if is_debug():
         if fd_type == STDOUT:
             sys.stdout.write(buffer)
         else:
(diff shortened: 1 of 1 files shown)


leapp/libraries/stdlib/__init__.py
```diff
@@ -133,6 +135,7 @@ def run(args, split=False, callback_raw=_logging_handler):
     :return: {'stdout' : stdout, 'stderr': stderr, 'signal': signal, 'exit_code': exit_code, 'pid': pid}
     :rtype: dict
     """
+    logging.getLogger('leapp.stdlib.Run').debug('External command is started: [%s]', ' '.join(args))
```
```suggestion
    api.current_logger().debug('External command is started: [%s]', ' '.join(args))
```

leapp/libraries/stdlib/__init__.py
```diff
@@ -152,4 +155,5 @@ def run(args, split=False, callback_raw=_logging_handler):
     finally:
         create_audit_entry('process-end', _id)
         create_audit_entry('process-result', {'id': _id, 'parameters': args, 'result': result})
+        logging.getLogger('leapp.stdlib.Run').debug('External command is finished: [%s]', ' '.join(args))
```
```suggestion
        api.current_logger().debug('External command is finished: [%s]', ' '.join(args))
```

leapp/libraries/stdlib/__init__.py
```diff
@@ -3,6 +3,7 @@
 represents a location for functions that otherwise would be defined multiple times across leapp actors
 and at the same time, they are really useful for other actors.
 """
+import logging
```
```suggestion
```

pirat89:
I afraid little about the readability. What about reword the solution like:
```
====== ext cmd started:
*** external command output ***
====== ext cmd finished:
````

Just idea that it is easy to find substring like `=====`. What do you think guys? @oamg/developers

PR #478, opened 2019-04-02
stdlib: _call: Ignore the linebuffer

pirat89:
We found that we are not able to handle linebuffer correctly now.
Sometimes it happens that the buffered string is broken because it
ends in the middle of utf8 character. This is temporary solution.


pirat89:
The related issue has been created to resolve the problem correctly later: https://github.com/oamg/leapp/pull/478

PR #480, opened 2019-04-02
Handle traceback when executed command is not found

vojtechsokol:
When command is executed using stdlib.run function and it fails due to a missing or mistyped command, traceback with CalledProcessError (A Leapp Command Error occurred.) and OSError ([Errno 2] No such file or directory) occurs.

However if even both of these exceptions are caught, there is still a traceback:
<details>
    <summary>Click for the traceback</summary>

```
  File "/usr/bin/leapp", line 9, in <module>
    load_entry_point('leapp==0.6.0', 'console_scripts', 'leapp')()
  File "/usr/lib/python2.7/site-packages/leapp/cli/__init__.py", line 30, in main
  File "/usr/lib/python2.7/site-packages/leapp/utils/clicmd.py", line 90, in execute
    args.func(args)
  File "/usr/lib/python2.7/site-packages/leapp/utils/clicmd.py", line 112, in called
    self.target(args)
  File "/usr/lib/python2.7/site-packages/leapp/cli/upgrade/__init__.py", line 124, in upgrade
    workflow.run(context=context, skip_phases_until=skip_phases_until)
  File "/usr/lib/python2.7/site-packages/leapp/workflows/__init__.py", line 210, in run
    if messaging.errors():
  File "/usr/lib/python2.7/site-packages/leapp/messaging/__init__.py", line 55, in errors
    return list(self._errors)
  File "<string>", line 2, in __len__
  File "/usr/lib64/python2.7/multiprocessing/managers.py", line 773, in _callmethod
    raise convert_to_error(kind, result)
multiprocessing.managers.RemoteError:
---------------------------------------------------------------------------
  File "/usr/lib64/python2.7/multiprocessing/managers.py", line 242, in serve_client
    obj, exposed, gettypeid = id_to_obj[ident]
KeyError: '7f9305620bd8'
---------------------------------------------------------------------------
```
</details>

The problem is in the [_call](https://github.com/oamg/leapp/blob/v0.6.0/leapp/libraries/stdlib/call.py#L81) function which executes the command in forked process.

This PR handles the traceback by raising a CalledProcessError exception (if the command is missing) before the command is executed.

Requires leapp-repository PR: oamg/leapp-repository#123

### DIFF
--- a/leapp/libraries/stdlib/call.py
+++ b/leapp/libraries/stdlib/call.py
@@ -1,8 +1,53 @@
 from __future__ import print_function
 
-import select
-import os
 import codecs
+import os
+import select
+
+
+try:
+    from shutil import which
+except ImportError:
+    def which(cmd, mode=os.F_OK | os.X_OK, path=None):
+        """Given a command, mode, and a PATH string, return the path which
+        conforms to the given mode on the PATH, or None if there is no such
+        file.
+
+        `mode` defaults to os.F_OK | os.X_OK. `path` defaults to the result
+        of os.environ.get("PATH"), or can be overridden with a custom search
+        path.
+
+        """
+        # Check that a given file can be accessed with the correct mode.
+        # Additionally check that `file` is not a directory, as on Windows
+        # directories pass the os.access check.
+        def _access_check(fn, mode):
+            return os.path.exists(fn) and os.access(fn, mode) and not os.path.isdir(fn)
+
+        # If we're given a path with a directory part, look it up directly rather
+        # than referring to PATH directories. This includes checking relative to the
+        # current directory, e.g. ./script
+        if os.path.dirname(cmd):
+            if _access_check(cmd, mode):
+                return cmd
+            return None
+
+        if path is None:
+            path = os.environ.get("PATH", os.defpath)
+        # DBG
+        if not path:
+            return None
+        path = path.split(os.pathsep)
+
+        seen = set()
+        for a_dir in path:
+            normdir = os.path.normcase(a_dir)
+            if normdir not in seen:
+                seen.add(normdir)
+                name = os.path.join(a_dir, cmd)
+                if _access_check(name, mode):
+                    return name
+        return None
 
 
 STDIN = 0
@@ -125,6 +170,10 @@ def _call(command, callback_raw=lambda fd, value: None, callback_linebuffered=la
         if not isinstance(env, dict):
             raise TypeError('env parameter has to be a dictionary')
         environ.update(env)
+
+    if which(command[0], path=environ['PATH']) is None:
+        return None
+
     # Create a separate pipe for stdout/stderr
     #
     # The parent process is going to use the read-end of the pipes for reading child's
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -9,10 +9,23 @@ import sys
 import uuid
 
 from leapp.exceptions import LeappError
-from leapp.utils.audit import create_audit_entry
 from leapp.libraries.stdlib import api
 from leapp.libraries.stdlib.call import _call, STDOUT
 from leapp.libraries.stdlib.config import is_debug
+from leapp.utils.audit import create_audit_entry
+
+
+class MissingCommandError(LeappError):
+    """
+    Raised when the called commend does not exist.
+    """
+    def __init__(self, command):
+        """
+        Initialize MissingCommandError Exception.
+        :param command: The command that is missing.
+        """
+        super(MissingCommandError, self).__init__('Missing command: \'{}\''.format(command))
+        self.command = command
 
 
 class CalledProcessError(LeappError):
@@ -171,6 +184,8 @@ def run(args, split=False, callback_raw=_console_logging_handler, callback_lineb
     try:
         create_audit_entry('process-start', {'id': _id, 'parameters': args, 'env': env})
         result = _call(args, callback_raw=callback_raw, callback_linebuffered=callback_linebuffered, env=env)
+        if checked and result is None:
+            raise MissingCommandError(command=args[0])
         if checked and result['exit_code'] != 0:
             raise CalledProcessError(
                 message="A Leapp Command Error occurred. ",
--- a/tests/scripts/test_stdlib.py
+++ b/tests/scripts/test_stdlib.py
@@ -1,7 +1,7 @@
 import os
 import pytest
 
-from leapp.libraries.stdlib import CalledProcessError, run
+from leapp.libraries.stdlib import CalledProcessError, MissingCommandError, run
 from leapp.libraries.stdlib.config import is_debug, is_verbose
 
 
@@ -31,9 +31,27 @@ def test_check_error():
         run(a_command, checked=True)
 
 
+def test_check_error_with_path(monkeypatch):
+    monkeypatch.setattr(os, 'environ', {'PATH': None})
+    a_command = ['true']
+    with pytest.raises(MissingCommandError):
+        run(a_command, checked=True)
+
+
 def test_check_error_no_checked():
     a_command = ['false']
     assert run(a_command, checked=False)['exit_code'] == 1
+
+
+def test_check_non_existing():
+    a_command = ['non-existing']
+    with pytest.raises(MissingCommandError):
+        run(a_command, checked=True)
+
+
+def test_check_non_existing_no_checked():
+    a_command = ['non-existing']
+    assert run(a_command, checked=False) is None
 
 
 def test_is_verbose(monkeypatch):


leapp/libraries/stdlib/call.py
```diff
@@ -102,6 +104,12 @@ def _call(command, callback_raw=lambda fd, value: None, callback_linebuffered=la
         raise ValueError('poll_timeout parameter has to be integer greater than zero')
     if not isinstance(read_buffer_size, int) or isinstance(read_buffer_size, bool) or read_buffer_size <= 0:
         raise ValueError('read_buffer_size parameter has to be integer greater than zero')
+    if find_executable(command[0]) is None:
```
vinzenz: Why are we doing this? I would like to veto this part - the error you are raising is not because the command actually failed but because it does not exist 
This is a different kind of error and if you are not happy with handling OSErrors like subprocess and all other to me known apis raise then replace it something that is distinguishable.

Besides if the path exist and it is not executable IIRC you would actually get a different kind of error.
You are trying to reinvent the wheel here for something that already is properly handled on a lower layer of this code. So rather handle and transform the exception than trying make your own checks that properly not even cover everything

vinzenz:
@shaded-enmity @pirat89 if you think that my opinion is wrong feel free to overrule my review.

shaded-enmity:
@vojtechsokol +1 on what @vinzenz said, the much cleaner solution here is to inspect the result of `_call` and raise accordingly.

Note that besides the already mentioned points, this wouldn't work as expected once #481 is merged since the check for executable is done in the parent process. So if I `_call`'d with custom `env={'PATH': ...}`, the `find_executable` check could fail because it'd be looking in the wrong `PATH`.

vinzenz:
Note _call actually still raises anythin that exec functions are raising and no error object is returned for this by _call

vojtechsokol:
@shaded-enmity
I don't think inspecting the result of `_call` would help here. How would you distinguish the case when command fails with exit code 1 from the case when command does not exist? In both cases the `result` dictionary is the same, except for things that are worthless:

- signal: 0 in both cases
- pid: worthless
- exit_code: 1 in both cases
- stdout: worthless
- stderr: worthless (string with traceback for non-existitng command)

And regarding #481 - yes, that was expected, but not unsolvable, therefeore the WIP label.

@vinzenz

Yes, I'm raising exception not because the command failed, but because it's missing, as can be seen from `message` of raised exception. This is the intention of this PR - differentiate the case when command fails from case when command does not exist.

Regarding the case when path exists but is not executable - yes, I'm aware of that, buth that was first shot, therefeore the WIP label.

So what would you recommend then? Taking any action in `stdlib.run` function is too late, what remains is the `_call` function, but what exception should be transformed there?

pirat89:
Just looking into the original `subprocess.call`, it's a little tricky that the function raises different exceptions, depending on the Python you are using. In case of Python2, it raises just `OSError` with various messages. In case of Python3, every specific problem has own exception (`PermissionError`, `FileNotFoundError`). Honestly I like the Python3 way, but the other exceptions should be derived from the `CalledProcessError`, so actors that already work with that, would be able to catch it as it is without change.

vinzenz:
> Just looking into the original `subprocess.call`, it's a little tricky that the function raises different exceptions, depending on the Python you are using. In case of Python2, it raises just `OSError` with various messages. In case of Python3, every specific problem has own exception (`PermissionError`, `FileNotFoundError`)

Which are all derived from OSError https://docs.python.org/3/library/exceptions.html#exception-hierarchy

 > Honestly I like the Python3 way, but the other exceptions should be derived from the `CalledProcessError`, so actors that already work with that, would be able to catch it as it is without change.

I actually prefer if the exception is separate, as one is representing the side of the child process and the other one all things before that.

shaded-enmity:
@vinzenz I just wonder if the parent/child separation is important enough to complicate the failure modes when `check=True`, I'd assume that most call sites of `run` don't really know/care about such distinction, but perhaps my assumption is wrong.

vinzenz:
> @vinzenz I just wonder if the parent/child separation is important enough to complicate the failure modes when `check=True`, I'd assume that most call sites of `run` don't really know/care about such distinction, but perhaps my assumption is wrong.

I mean I would be fine if we would go with something like CommandExecutionError and our CalledProcessError being a derived class for it. But whatever we do, we will have to change already all the actors that are checking for OSError and may actually use it. e..g I use the distinction between OSError and CalledProcessError to decide what I am going to report to the user. e.g. Check that the thing is actually installed vs hmm maybe it's about your network, storage or whatever.

```python
try:
    run()
except OSError as e:
    # report the user that the app is not there  or even if there's a workaround like trying to use a different program then go this route
except CalledProcessError as e:
    # Report stderr of the command (Or whatever)
```

If I would have a unified way I have to go with something like:
```python
try:
    run()
except CalledProcessError as e:
    if e.exit_code is None: # Or whatever the moniker will be for this
      # hmm ok what was going on now? OSError will be stored inside?
    else:
      # Report stderr of the command (Or whatever)
```

Well, I am not sure why we actually need to do anything about this exception, you can use OSError and should be covered (from what I have seen but I am ok to be proven wrong) 

For whose convenience is this? If you want to make a convenience you can make a tuple available that contains all the exceptions and people should catch that if they do not care about the specifics.

I don't want to make the live harder for anyone who does actually care about the difference.

```python

from leapp.libraries.stdlib import run, LEAPP_RUN_ERRORS
...
try:
    run()
except LEAPP_RUN_ERRORS:
   pass
```

And for everybody else, you can keep CalledProcessError and OSError this way you don't break anything for anyone and allow people to catch what they need.


And about check=False, since this was supposed to be about the exit_code, OSError could still be thrown check=False wouldn't make the execution exception free, at least that's what I wouldn't except. If that is wanted there should be rather a 'noexcept' or whatever.

pirat89:
@vinzenz 
> If we would have a unified way one would end up with something like:
> 
> ```python
> try:
>     run()
> except CalledProcessError as e:
>     if e.exit_code is None: # Or whatever the moniker will be for this
>       # hmm ok what was going on now? OSError will be stored inside?
>     else:
>       # Report stderr of the command (Or whatever)
> ```
It doesn't need to. It would end up with something like:
```
try:
    run()
except FileNotFoundError as e:
   # do what you want
except CalledProcessError as e:
   # do what you want
```

And in case they do not care about difference between those two errors, they still can use just
```
try:
    run()
except CalledProcessError as e:
    # to catch 'everywhing'

**Comment by pirat89** [comment]
> Well, I am not sure why we actually need to do anything about this exception, you can use OSError and should be covered (from what I have seen but I am ok to be proven wrong)
> 
> For whose convenience is this? If you want to make a convenience you can make a tuple available that contains all the exceptions and people should catch that if they do not care about the specifics.
> 
> I don't want to make the life harder for anyone who does actually care about the difference.
> 
> ```python
> from leapp.libraries.stdlib import run, LEAPP_RUN_ERRORS
> ...
> try:
>     run()
> except LEAPP_RUN_ERRORS:
>    pass
> ```
> 
> And for everybody else, you can keep CalledProcessError and OSError this way you don't break anything for anyone and allow people to catch what they need.
> 

That would be solution as well. Even when I would like to go more with Python3 exceptions. But that's just preference. I have not so big problem with that solution as well. But in such case we will have to refactor all caches in leapp-repository. In the unified way, we do not have to do any changes and it's quite easy to work with that as well, without problems.

**Comment by pirat89** [comment]
@shaded-enmity just to keep you on the board :)

**Comment by vinzenz** [comment]
> > Well, I am not sure why we actually need to do anything about this exception, you can use OSError and should be covered (from what I have seen but I am ok to be proven wrong)
> > For whose convenience is this? If you want to make a convenience you can make a tuple available that contains all the exceptions and people should catch that if they do not care about the specifics.
> > I don't want to make the life harder for anyone who does actually care about the difference.
> > ```python
> > from leapp.libraries.stdlib import run, LEAPP_RUN_ERRORS
> > ...
> > try:
> >     run()
> > except LEAPP_RUN_ERRORS:
> >    pass
> > ```
> > 
> > 
> > And for everybody else, you can keep CalledProcessError and OSError this way you don't break anything for anyone and allow people to catch what they need.
> 
> That would be solution as well. Even when I would like to go more with Python3 exceptions. But that's just preference. I have not so big problem with that solution as well. But in such case we will have to refactor all caches in leapp-repository.

The point is that you don't have to refactor anything at this point, it's a can be updated by someone touching the actor. If you start introducing new Exceptions however that's a different story. Then you have to refactor everything also it's not so easy to make our exception + a builtin exception. I don't really see why you feel like we have to do anything at all. The only thing you can introduce is that tuple that has OSError and CalledProcessError and it should catch all the possible exceptions from run() - Theres no difference right now. Already everyone should be catching OSError at this point, if they don't then their code didn't handle the exceptions properly already.

> In the unified way, we do not have to do any changes and it's quite easy to work with that as well, without problems.

In the unified way you have to refactor every code, that actually handles them separately already.
While if you do just introduce a tuple to catch everything, only new code should use it.

**Comment by vojtechsokol** [comment]
@vinzenz 
As stated in the PR description - catching `OSError` and `CalledProcessError` is not enough, because this would lead to a traceback with `KeyError`.

Now the code should properly handle the cases when the path exists but is not executable, and when environment variable with `PATH` is used.

**Comment by pirat89** [comment]
@vinzenz man, I am not getting your answer. I see that in completely opossite way, when we will have to change something and when not. What are you writting, looks to me like you missed something in my answer or I am missing something in your answer. If you have:
```
class CalledProcessError(LeappError):
   ...

class FileNotFoundError(LeappError):
   ...

clas ...
```

Why I need to touch every call that catches `CalledProcessError`? The behaviour for them is same + catching errors when command doesn't exist. In case they care about the reason more, they can add catch of additional error. This means in >90% of cases it doesn't need any change from us. In your case, it's opposite as you need to change import and catch in >90%. Taking the point, that probably everyone who used catch of `CalledProcessError` expect that the only thing they need to do and don't know that there are other cases. (other cases generate nowadays raw, unhandled tracebacks in actors instead of clear error messages for users).

Additionally, generated OSErrors, etc. in this solution will point more to a bug in leapp stdlib instead of problem in command itself. That would be beneficial for us as well.

I think we are desynced here. Additional discussion through comments would be too slow, so I would prefer mtg to discuss it in case of need.

vinzenz:
The actual cause of the Traceback is handled now by https://github.com/oamg/leapp/pull/533

PR #481, opened 2019-04-03
modify _call to allow env variable injection

jmikovic:


jmikovic:
e2e tests

leapp/libraries/stdlib/__init__.py
```diff
@@ -132,7 +132,7 @@ def run(args, split=False, callback_raw=_logging_handler):
     result = None
     try:
         create_audit_entry('process-start', {'id': _id, 'parameters': args})
```
shaded-enmity: I missed this the first time:

```suggestion
        create_audit_entry('process-start', {'id': _id, 'parameters': args, 'env': env})
```

PR #482, opened 2019-04-05
[WIP] Enable profilling when env LEAPP_CPROFILE == '1'

pirat89:
Currently on some machines the run of the leapp / snactor is too
slow. Especially framework itself should be fast but sometimes the
time differences are in order of tens seconds. From this point,
it seems useful to be able to start profilling based on the env
variable when it is needed.

### DIFF
--- a/leapp/snactor/__init__.py
+++ b/leapp/snactor/__init__.py
@@ -88,7 +88,7 @@ def main():
     cli.command.execute(version=_('snactor version {}').format(VERSION))
     if profile_enabled:
         pr.disable()
-        s = StringIO.StringIO()
+        s = StringIO()
         sortby = 'cumulative'
         ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
         ps.print_stats()
--- a/leapp/snactor/__init__.py
+++ b/leapp/snactor/__init__.py
@@ -2,6 +2,17 @@ import os
 import pkgutil
 import socket
 
+# for profilling
+import cProfile
+import pstats
+try:
+    from StringIO import StringIO
+except ImportError:
+    # TODO: low possibility of the problem with encoding with Python3;
+    # # but it should not be so problematic in this case, so keeping now just
+    # # like that to keep it simple
+    from io import StringIO
+
 from leapp.utils.i18n import _
 from leapp.snactor import commands
 from leapp.snactor.commands import workflow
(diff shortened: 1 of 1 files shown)


leapp/snactor/__init__.py
```diff
@@ -68,6 +79,17 @@ def cli(args):
...
+    profile_enabled = os.environ.get('LEAPP_CPROFILE', '0') == '1'
+    if profile_enabled:
+        pr = cProfile.Profile()
+        pr.enable()
     os.environ['LEAPP_HOSTNAME'] = socket.getfqdn()
     load_commands()
     cli.command.execute(version=_('snactor version {}').format(VERSION))
+    if profile_enabled:
+        pr.disable()
+        s = StringIO.StringIO()
```
vinzenz: I am not totally sure, but this should be:
```suggestion
        s = StringIO()
```
right?

leapp/snactor/__init__.py
```diff
@@ -68,6 +79,17 @@ def cli(args):
...
+        sortby = 'cumulative'
+        ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
+        ps.print_stats()
+        print(s.getvalue())
```
vinzenz: I'd rather like to see this being printed to stderr

@AloisMahdal what's your take on this?

leapp/snactor/__init__.py
```diff
@@ -68,6 +79,17 @@ def cli(args):
 
 
 def main():
+    profile_enabled = os.environ.get('LEAPP_CPROFILE', '0') == '1'
```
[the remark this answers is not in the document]
vinzenz: While I +1 the idea, I was just thinking about this and I am wondering - Is it relevant that you know it's on for others that this? This is the only use case right now, right?

leapp/snactor/__init__.py
pirat89: yes!

leapp/snactor/__init__.py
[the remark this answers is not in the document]
vojtechsokol: This place is the only use case so far. I was thinking if it's worth it to move it to stdlib and I'm kind of indecisive. Maybe it could be moved there if another use case appears. So feel free to dismiss this sugeestion.

pirat89:
Closing in favor of #676

PR #490, opened 2019-04-10
Fix _call unicode issues

### LINKED ISSUE #488: Fix leapp.stdlib.call
Problem is around the `callback_raw` and `callback_linebuffered` functions. Documentation in the `_call` function tells (a), implementation inside `_multiplex` tells (b) (and in advance, the implementation is broken) and our callback functions in `leapp.stdlib` are kinda logic (c).

tl;dr - the current state is not systematic. Someone should sit above this and should fix the current state so the logic, doc, naming, ... will be clean and reliable.

PLUS: we need tests! Whoever will take this, please think about the implementation and design before you implement this. Keep in mind that some `callback_raw` functions are implemented in some actors as well, so when signature of the `callback_raw` function is changed, the modification will have to appear in such actors as well.

Hard requirement: __all related functions are really covered by tests!!__

shaded-enmity:
* Input data are now decoded using incremental decoder
* Docstring now reflects parameters better
* Add tests for various language pangrams

Fixes #488

### DIFF
--- a/leapp/libraries/stdlib/call.py
+++ b/leapp/libraries/stdlib/call.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 
 import select
 import os
+import codecs
 
 
 STDIN = 0
@@ -13,10 +14,6 @@ def _multiplex(ep, read_fds, callback_raw, callback_linebuffered,
                encoding='utf-8', write=None, timeout=1, buffer_size=80):
     # Register the file descriptors (stdout + stderr) with the epoll object
     # so that we'll get notifications when data are ready to read
-    # FIXME: the callback_linebuffered buffer is ignored now as the input stream
-    # # is not handled properly, all realted FIXME parts in the functions
-    # # are marked by (**LINE_BUFFER**) comment
-
     for fd in read_fds:
         ep.register(fd, select.EPOLLIN | select.EPOLLPRI)
 
@@ -35,8 +32,9 @@ def _multiplex(ep, read_fds, callback_raw, callback_linebuffered,
     num_expected = len(read_fds) + (1 if write else 0)
     # Set up file-descriptor specific buffers where we'll buffer the output
     buf = {fd: bytes() for fd in read_fds}
-    # FIXME: (**LINE_BUFFER**)
-    # linebufs = {fd: '' for fd in read_fds}
+    if encoding:
+        linebufs = {fd: '' for fd in read_fds}
+        decoders = {fd: codecs.getincrementaldecoder(encoding)() for fd in read_fds}
 
     while not ep.closed and len(hupped) != num_expected:
         events = ep.poll(timeout)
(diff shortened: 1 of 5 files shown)


shaded-enmity:
Note that #481 should be merged first then I'll rebase. I've accidentally squashed Jozef's commit :angel:

bocekm:
ping @drehak - you can test if this fixes also the returned non-unicode stuff you've dealt with today.

leapp/libraries/stdlib/call.py
```diff
@@ -68,21 +65,19 @@ def _multiplex(ep, read_fds, callback_raw, callback_linebuffered,
...
-    #         # [stdout, stderr] is relayed, stdout=1 a stderr=2
-    #         # as the field starting indexed is 0, so the +1 needs to be added
-    #         callback_linebuffered(read_fds.index(fd) + 1, lb)
+    # Process leftovers from line buffering
+    if encoding:
+        for (fd, lb) in linebufs.items():
+            if lb:
+                # [stdout, stderr] is relayed, stdout=1 a stderr=2
+                # as the field starting indexed is 0, so the +1 needs to be added
+                callback_linebuffered(read_fds.index(fd) + 1, lb)
```
pirat89: Still broken :)

```suggestion
                callback_linebuffered((fd, read_fds.index(fd) + 1), lb)
```

pirat89: Annyway, `red_fds.index(fd) + 1` is several times (2x seeing now). From the point that this should provide fd_type, what about rather write nested function `_get_fd_type`, like:
```
        def _get_fd_type(fd):
            return read_fds.index(fd) + 1
```

tests/scripts/test_stdlib_call.py
```diff
@@ -29,14 +31,14 @@ def test_stdin_fd():
...
-#
-#     def callback(fd, data):
-#         buffered.append(data)
-#     _call(('bash', '-c', 'echo 1; echo 2; echo 3'), callback_linebuffered=callback)
-#     assert buffered == ['1', '2', '3']
+
+def test_linebuffer_callback():
+    buffered = []
+
+    def callback(fd, data):
```
pirat89: just nitpick, fd_info would be better, but I am ok with fd as well as it is not used.

pirat89:
close the PR and create new one with "correct" branch name :) - or add fix to Makefile
 - our sed doesn't like `/`

shaded-enmity:
Even after updating `.flake8` to ignore E117 it still reports E117. At this point I'd appreciate someone looking into our CI setup. I don't think a 2h game of whack-a-mole with misconfigured linter is worth anyones time.

leapp/libraries/stdlib/call.py
```diff
@@ -94,25 +97,35 @@ def _call(command, callback_raw=lambda fd, value: None, callback_linebuffered=la
...
         raise TypeError('command parameter has to be a list or tuple')
-    if not callable(callback_raw) or callback_raw.__code__.co_argcount != 2:
-        raise TypeError('callback_raw parameter has to be callable accepting 2 parameters')
-    if not callable(callback_linebuffered) or callback_linebuffered.__code__.co_argcount != 2:
+    if not callable(callback_raw) or\
+            (getattr(callback_raw, '__code__', None) and callback_raw.__code__.co_argcount != 2):
+            raise TypeError('callback_raw parameter has to be callable accepting 2 parameters')
+    if not callable(callback_linebuffered)\
+            or (getattr(callback_linebuffered, '__code__', None)
+                and callback_linebuffered.__code__.co_argcount != 2):
```
jmikovic:
```suggestion
                and callback_linebuffered.__code__.co_argcount != 2): # noqa
```

should ignore flake8 error

leapp/libraries/stdlib/call.py
```diff
@@ -94,25 +97,35 @@ def _call(command, callback_raw=lambda fd, value: None, callback_linebuffered=la
...
         :rtype: dict
     """
     if not isinstance(command, (list, tuple)):
         raise TypeError('command parameter has to be a list or tuple')
-    if not callable(callback_raw) or callback_raw.__code__.co_argcount != 2:
-        raise TypeError('callback_raw parameter has to be callable accepting 2 parameters')
-    if not callable(callback_linebuffered) or callback_linebuffered.__code__.co_argcount != 2:
+    if not callable(callback_raw) or\
+            (getattr(callback_raw, '__code__', None) and callback_raw.__code__.co_argcount != 2):
+            raise TypeError('callback_raw parameter has to be callable accepting 2 parameters')
```
jmikovic:
```suggestion
        raise TypeError('callback_raw parameter has to be callable accepting 2 parameters')
```

fixes over-indent error

shaded-enmity:
Thanks to @jmikovic this PR now passes all the flake8 hurdles!

PR #493, opened 2019-04-16
Changes to archive and create logs structure

asilveir:
This PR  will archive old logs in `/var/log/leapp/archive` and create the logs directory structure.

```
/var/log/leapp/dnf-debugdata/
/var/log/leapp/leapp-upgrade.log
/var/log/leapp/leapp-report.txt
/var/log/leapp/dnf-plugin-data.txt
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -10,6 +10,7 @@ A clear and concise description of what the bug is.
 
 **To Reproduce**
 Steps to reproduce the behavior
+
 1. install '...'
 2. setup '....'
 3. run '....'
@@ -20,25 +21,32 @@ Steps to reproduce the behavior
 A clear and concise description of what you expected to happen.
 
 **System information (please complete the following information):**
- - OS and version: (e.g. Fedora 29 or `$ cat /etc/system-release`)
- - `# uname -a`
- - `# rpm -qa "*leapp*"` (or shorthashes of commits in case of manual installation):
+
+- OS and version: (e.g. Fedora 29 or `$ cat /etc/system-release`)
+- `# uname -a`
+- `# rpm -qa "*leapp*"` (or shorthashes of commits in case of manual installation):
+
 ```
 # rpm -qa "leapp*"
 leapp-0.4.0-1.201812151505Z.87a7b74.master.el7_5.noarch
 leapp-repository-data-0.4.0-1.201812131256Z.f25d14e.master.el7_5.noarch
 leapp-repository-0.4.0-1.201812131256Z.f25d14e.master.el7_5.noarch
 ```
- - attach (or provide link to) log files if applicable (optional - may contain confidential information):
-   - **_/var/log/upgrade.log_**
-   - */tmp/leapp-report.txt*
-   - `# tar -czf dnf-debug.tgz /tmp/download-debugdata`
-   - */var/lib/leapp/leapp.db*
-   - *journalctl*
- - anything else you would like to provide (e.g. storage info):
+
+Attach (or provide link to) log files if applicable (optional - may contain confidential information):
+
+- *All files in /var/log/leapp/*
+- */var/lib/leapp/leapp.db*
+- *journalctl*
+- If you want, you can optionally send anything else would you like to provide (e.g. storage info)
+
+**For your convenience you can pack all logs with this command:**
+
+`# tar -czf leapp-logs.tgz /var/log/leapp /var/lib/leapp/leapp.db`
+
+Then you may attach only the `leapp-logs.tgz` file.
 
 ****
 
 **Additional context**
 Add any other context about the problem here.
-
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -2,8 +2,11 @@ from __future__ import print_function
 import itertools
 import json
 import os
+import shutil
 import sys
+import tarfile
 import uuid
+from datetime import datetime
 
 from leapp.config import get_config
 from leapp.exceptions import CommandError, LeappError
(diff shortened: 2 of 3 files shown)


leapp/cli/upgrade/__init__.py
```diff
@@ -14,6 +17,46 @@
...
+                    os.remove(file)
+                except OSError:
+                    pass
+            if os.path.isfile(leapp_db):
+                tar.add(leapp_db)
+
+    if os.getenv('LEAPP_DEBUG', '0') == '1':
+        logs_dir = os.path.join(logs_dir, 'dnf-debugdata')
+    if not os.path.isdir(logs_dir):
+        os.makedirs(logs_dir)
```
bocekm: Do not create the dir here, but instead put it to spec file:
[line 186](https://github.com/oamg/leapp/blob/master/packaging/leapp.spec#L186): `install -m 0755 -d %{buildroot}%{_localstatedir}/log/leapp`
[line 220](https://github.com/oamg/leapp/blob/master/packaging/leapp.spec#L220): `%dir %{_localstatedir}/log/leapp`

leapp/cli/upgrade/__init__.py
```diff
@@ -14,6 +17,40 @@
 from leapp.utils.output import report_errors, beautify_actor_exception
 
 
+def logs_handler():
```
bocekm: I'd rename it to logfile_handler as this function does not handle all the logging (inlcuding logging to audit db, stdout/stderr, etc.), but it handles log files only.

bocekm: Well, as I'm thinking about it, now it does nothing else then archiving logs. So would name it `archive_logfiles()`.

asilveir: It makes sense!

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -20,25 +21,34 @@ Steps to reproduce the behavior
...
-   - **_/var/log/upgrade.log_**
-   - */tmp/leapp-report.txt*
-   - `# tar -czf dnf-debug.tgz /tmp/download-debugdata`
-   - */var/lib/leapp/leapp.db*
-   - *journalctl*
- - anything else you would like to provide (e.g. storage info):
+
+Attach (or provide link to) log files if applicable (optional - may contain confidential information):
+
+- *All files in /var/log/leapp*
```
pirat89:
```suggestion
- *All files in /var/log/leapp/*
```

leapp/cli/upgrade/__init__.py
```diff
@@ -14,6 +17,40 @@
...
+        '/var/log/leapp/leapp-upgrade.log',
+        '/var/log/leapp/leapp-report.txt',
+        '/var/log/leapp/dnf-plugin-data.txt'
+    ]
+    files_to_archive = [file for file in log_files if os.path.isfile(file)]
+
+    if os.path.isdir(debug_dir) and os.listdir(debug_dir):
+        files_to_archive.append(debug_dir)
+
+    if not os.path.isdir(archive_dir):
```
bocekm: Move the creation of the folder under the condition `if files_to_archive:`

leapp/cli/upgrade/__init__.py
```diff
@@ -14,6 +17,40 @@
...
+        os.makedirs(archive_dir)
+
+    if files_to_archive:
+        archive_file = os.path.join(archive_dir, 'leapp-{}-logs.tar.gz'.format(int(time.time())))
```
bocekm: change the time.time() to something else. Now it returns number of second since 1.1.1970.
E.g. `leapp-1555445048-logs.tar.gz`

bocekm: Take a look at https://github.com/oamg/leapp/pull/494

leapp/cli/upgrade/__init__.py
```diff
@@ -14,6 +17,40 @@
...
+
+        with tarfile.open(archive_file, "w:gz") as tar:
+            for file in files_to_archive:
```
bocekm: Do not use `file` because it's a python builtin.

leapp/cli/upgrade/__init__.py
```diff
@@ -14,6 +17,40 @@
...
+    debug_dir = '/var/log/leapp/dnf-debugdata/'
+    leapp_db = '/var/lib/leapp/leapp.db'
+    log_files = [
+        '/var/log/leapp/leapp-upgrade.log',
+        '/var/log/leapp/leapp-report.txt',
+        '/var/log/leapp/dnf-plugin-data.txt'
+    ]
+    files_to_archive = [file for file in log_files if os.path.isfile(file)]
+
+    if os.path.isdir(debug_dir) and os.listdir(debug_dir):
```
bocekm:
```suggestion
    if os.path.isdir(debug_dir):
```
This is enough. Because if it is empty, it's not added to the _files_to_archive_ and then it's not deleted.

leapp/cli/upgrade/__init__.py
```diff
@@ -14,6 +17,40 @@
...
+        with tarfile.open(archive_file, "w:gz") as tar:
+            for file in files_to_archive:
+                tar.add(file)
+                if os.path.isdir(file):
+                    shutil.rmtree(file, ignore_errors=True)
+                try:
+                    os.remove(file)
+                except OSError:
+                    pass
+            if os.path.isfile(leapp_db):
```
bocekm:
```suggestion
            # leapp_db is not in files_to_archive to not have it removed
            if os.path.isfile(leapp_db):
```

leapp/cli/upgrade/__init__.py
```diff
@@ -14,6 +17,40 @@
 from leapp.utils.output import report_errors, beautify_actor_exception
 
 
+def archive_logfiles():
+    """ Archive log files from a previous run of Leapp """
```
bocekm:
```suggestion
    """ Archive log files from a previous run of Leapp """
    # FIXME: The files in /var/log/leapp/ needs to be defined on one place. They are defined also in actors.
```

PR #496, opened 2019-04-17
Fix spec file

vojtechsokol:


packaging/leapp.spec
```diff
@@ -23,8 +23,8 @@ Release:    1%{?dist}
 Summary:    OS & Application modernization framework
 
 License:    ASL 2.0
-URL:        https://leapp-to.github.io
-Source0:    https://github.com/leapp-to/%{name}/archive/%{gittag}/%{name}-%{version}.tar.gz
+URL:        https://oamg.github.io/leapp/
+Source0:    https://github.com/oamg/%{name}/archive/%{gittag}/%{name}-%{version}.tar.gz
```
pirat89: https://github.com/oamg/leapp/archive/v%{version}.tar.gz#/%{name}-%{version}.tar.gz

PR #497, opened 2019-04-18
Make clear what show_message() from actor API does

bocekm:
With the current explanation of the function, it's not clear to me where the message will be shown to the user.
All it does is `print` only: https://github.com/oamg/leapp/blob/master/leapp/dialogs/renderer.py#L142

### DIFF
--- a/leapp/actors/__init__.py
+++ b/leapp/actors/__init__.py
@@ -104,7 +104,9 @@ class Actor(object):
 
     def show_message(self, message):
         """
-        Used to display messages to the user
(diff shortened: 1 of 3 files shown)


AloisMahdal:
Is the `without storing it in audit database` really relevant?

vojtechsokol:
leapp-ci build

bocekm:
@vinzenz how does it look like now?

pirat89:
IMO it is good now. But waiting for the others.

PR #498, opened 2019-04-24
stdlib: add option to run() to ignore exit code

mlichvar:
This will be useful in the chrony check actor (https://github.com/oamg/leapp-repository/pull/139).

### DIFF
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -119,7 +119,7 @@ def _logging_handler(fd_info, buffer):
         __write_raw(fd_info, buffer)
 
 
-def run(args, split=False, callback_raw=_logging_handler, env=None):
+def run(args, split=False, callback_raw=_logging_handler, env=None, checked=True):
     """
     Run a command and return its result as a dict.
 
@@ -131,6 +131,8 @@ def run(args, split=False, callback_raw=_logging_handler, env=None):
     :type split: bool
     :param callback_raw: Optional custom callback executed on raw data to print in console
     :type callback_raw: (fd: int, buffer: bytes) -> None
+    :param checked: Raise an exception on a non-zero exit code, default True
+    :type checked: bool
     :return: {'stdout' : stdout, 'stderr': stderr, 'signal': signal, 'exit_code': exit_code, 'pid': pid}
     :rtype: dict
     """
@@ -140,7 +142,7 @@ def run(args, split=False, callback_raw=_logging_handler, env=None):
     try:
         create_audit_entry('process-start', {'id': _id, 'parameters': args, 'env': env})
         result = _call(args, callback_raw=callback_raw, env=env)
-        if result['exit_code'] != 0:
+        if checked and result['exit_code'] != 0:
             raise CalledProcessError(
                 message="A Leapp Command Error occurred. ",
                 command=args,
--- a/tests/scripts/test_stdlib.py
+++ b/tests/scripts/test_stdlib.py
@@ -1,6 +1,7 @@
 import os
+import pytest
 
-from leapp.libraries.stdlib import run
+from leapp.libraries.stdlib import CalledProcessError, run
 from leapp.libraries.stdlib.config import is_debug, is_verbose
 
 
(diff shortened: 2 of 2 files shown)


leapp/libraries/stdlib/__init__.py
```diff
@@ -140,7 +142,7 @@ def run(args, split=False, callback_raw=_logging_handler, env=None):
     try:
         create_audit_entry('process-start', {'id': _id, 'parameters': args, 'env': env})
         result = _call(args, callback_raw=callback_raw, env=env)
-        if result['exit_code'] != 0:
+        if not ignore_exit_code and result['exit_code'] != 0:
```
fernflower: Maybe add warning at logging level? At least it won't be completely silent then

fernflower: On second thought - probably not

fernflower:
Can you please add a test for it?

leapp/libraries/stdlib/__init__.py
```diff
@@ -119,7 +119,7 @@ def _logging_handler(fd_info, buffer):
         __write_raw(fd_info, buffer)
 
 
-def run(args, split=False, callback_raw=_logging_handler, env=None):
+def run(args, split=False, callback_raw=_logging_handler, env=None, ignore_exit_code=False):
```
```suggestion
def run(args, split=False, callback_raw=_logging_handler, env=None, checked=True):
```

leapp/libraries/stdlib/__init__.py
```diff
@@ -131,6 +131,8 @@ def run(args, split=False, callback_raw=_logging_handler, env=None):
     :type split: bool
     :param callback_raw: Optional custom callback executed on raw data to print in console
     :type callback_raw: (fd: int, buffer: bytes) -> None
+    :param ignore_exit_code: Don't raise exception on a non-zero exit code
```
```suggestion
    :param checked: Raise an exception on a non-zero exit code, default True
```

leapp/libraries/stdlib/__init__.py
```diff
@@ -131,6 +131,8 @@ def run(args, split=False, callback_raw=_logging_handler, env=None):
...
+    :type ignore_exit_code: bool
```
```suggestion
    :type checked: bool
```

leapp/libraries/stdlib/__init__.py
```suggestion
        if checked and result['exit_code'] != 0:
```

mlichvar: Ok. Makes sense.

mlichvar:
> Can you please add a test for it?

Ok, I added two tests and changed the code to use the ``checked`` parameters

tests/scripts/test_stdlib.py
```diff
@@ -24,6 +24,20 @@ def test_check_multiline_output_no_split():
     assert run(a_command, split=False)['stdout'] == u'This a multi-\nline No Split test!\n'
 
 
+def test_check_error():
+    a_command = ['false']
+    try:
```
fernflower: Please use pytest.raises for testing exception is raised, smth like 

```
with pytest.raises(CalledProcessError) as err:                                                  
    run(a_command, checked=True) 
```

https://docs.pytest.org/en/latest/assert.html

tests/scripts/test_stdlib.py
```diff
@@ -24,6 +24,20 @@ def test_check_multiline_output_no_split():
     assert run(a_command, split=False)['stdout'] == u'This a multi-\nline No Split test!\n'
 
 
+def test_check_error():
+    a_command = ['false']
```
fernflower: I'd prefer not 'false' but something self-speaking like 'nosuchcommand'

tests/scripts/test_stdlib.py
```diff
@@ -24,6 +24,20 @@ def test_check_multiline_output_no_split():
...
+    try:
+        run(a_command, checked=True)
+        assert False
+    except CalledProcessError:
+        pass
+
+
+def test_check_error_no_checked():
+    a_command = ['false']
```
fernflower: Same thing here

tests/scripts/test_stdlib.py
mlichvar: I think we should be testing exit codes of existing commands here.

tests/scripts/test_stdlib.py
mlichvar: Ok. Changed in the current patch.

leapp/libraries/stdlib/__init__.py
```diff
@@ -131,6 +131,8 @@ def run(args, split=False, callback_raw=_logging_handler, env=None):
     :type split: bool
     :param callback_raw: Optional custom callback executed on raw data to print in console
     :type callback_raw: (fd: int, buffer: bytes) -> None
+    :param checked: Raise exception on a non-zero exit code, default True
```
```suggestion
    :param checked: Raise an exception on a non-zero exit code, default True
```

mlichvar: Thanks.

tests/scripts/test_stdlib.py
fernflower: makes sense

PR #501, opened 2019-04-30
Add a simple logger to snactor discover command

asilveir:
It's a fix to snactor to avoid spurious output to console:
`No handlers could be found for logger "leapp.linter"`

Now the `snactor discover` will print message like this:
`WARNING:leapp.linter:Actor <class 'actor.Biosdevname'> field produces should be a tuple of Models`

### DIFF
--- a/leapp/snactor/commands/discover.py
+++ b/leapp/snactor/commands/discover.py
@@ -1,4 +1,5 @@
 import json as json_mod
+import logging
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


leapp/snactor/commands/discover.py
```diff
@@ -89,6 +90,7 @@ def _get_model_details(model):
 @command_opt('safe', is_flag=True, help='Analyze actor files statically to work around runtime errors')
 @requires_repository
 def cli(args):
+    logging.basicConfig(level=logging.WARNING, stream=sys.stderr)
```
vinzenz: I wonder if it wouldn't be easier to think about putting configure_logger from leapp.logger at the end of def cli in leapp/snactor/__init__.py and remove the configure_logger calls from `snactor workflow run` and `snactor run`

PR #502, opened 2019-05-03
api: Add consume_first for convenience

vinzenz:
A lot of actors are expecting only one type and are using next to
retrieve the value. This patch introduces a convenience function that is
simplifying the usage for actors that have a need for this.

Additionally it covers some gotchas that are mostly not handled, mainly
if there is no message available then an empty tuple is returned that is
not iterable by next.

Usage example:
--------------

```
from leapp.models import SomeModel

def some_function_previously():
    value = next(api.consume(SomeModel), None)
    value_other = next(api.consume(SomeModel), SomeModel())

def some_function_now():
    value = api.consume_first(SomeModel)
    value_other api.consume_first(SomeModel, default=SomeModel())
```


shaded-enmity:
Wondering if it makes sense to check if there's actually more messages and maybe emit a log message if there is, i.e. `Discarded 3 MyModel messages  in somerepo.Someactor`?

vinzenz:
> Wondering if it makes sense to check if there's actually more messages and maybe emit a log message if there is, i.e. `Discarded 3 MyModel messages in somerepo.Someactor`?

That'd be something like expect_one or so, no?

shaded-enmity:
Yeah, for me `consume_first` entails `expect_one` as well - what I mean, thinking from a long term maintenance perspective, if I say `consume_first` and at some point in time another actor is added which produces that message, suddenly there are two messages for me to consume, however I have no knowledge of the contents / semantics of the second message, and even if I did, I probably don't have control over which message will be produced first.
It is also more explicit in terms of expectancies - if I see actor use `expect_one` I can be pretty sure that adding a second message will not play along nicely with the logic the actor implements.

pirat89:
+1 for `expect_one` or something like that. But it creates impression in my mind, that one msg is always *expected*. And there doesn't have to be any. So maybe another naming would be better yet.

fernflower:
That's a nice to have thing, maybe we could revive it? 
FWIW, `consume_one` naming makes more sense to me :)

PR #508, opened 2019-05-16
Redirect from github.io to RTD

asilveir:
All documentation will be consolidate in Read the docs.


index.html
```diff
@@ -0,0 +1,3 @@
+<head>
+    <meta http-equiv="refresh" content="0; URL='https://leapp.readthedocs.io/en/latest/'" />
+  </head>
```
GAZDOWN: This is unfortunately not HTML compliant and it may not work in some browsers at all. The code needs to be extended by minimal html structure (example here: https://www.sitepoint.com/a-minimal-html-document-html5-edition/) and while doing so, I would add a link in the body, so user can click on that in case the automatic redirect fails.

pirat89:
Works as expected.

Issue #510, opened 2019-05-21
Importing actor's library from another library of the same actor depends on names of the libraries


olysonek:
**Actual behavior**
Say I have an actor `A` that has two private libraries: `libraryA` and `libraryB`. If `libraryA` imports `libraryB`, the Leapp's import system cannot deal with it - it will fail with `ImportError: cannot import name libraryB`. I believe this is because at the time when Leapp is trying to import `libraryA`, `libraryB` has not yet been imported. I.e., Leapp tries to import one library at a time in a given order and does not recurse.

**To Reproduce**
You can try this out on the existing VsftpdConfigRead actor. The actor has two private libraries: `library` and `config_parser`. `library` imports `config_parser`.
https://github.com/oamg/leapp-repository/tree/master/repos/system_upgrade/el7toel8/actors/vsftpdconfigread

If you run the tests of the actor as it is now, it will work just fine. However if you rename the actor's private library `config_parser` to something that lexicographically comes after `library`, such as `moo`, Leapp will fail in the import phase. You can reproduce the problem with the following patch:
```
diff --git a/repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/config_parser.py b/repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/moo.py
similarity index 100%
rename from repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/config_parser.py
rename to repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/moo.py
diff --git a/repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py b/repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
index cb3c688..ec17bdf 100644
--- a/repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
@@ -1,7 +1,7 @@
 import errno
 import os
 
-from leapp.libraries.actor import config_parser
+from leapp.libraries.actor import moo
 from leapp.libraries.common.vsftpdutils import get_config_contents, get_default_config_hash, \
                                                read_file, \
                                                VSFTPD_CONFIG_DIR, \
@@ -12,9 +12,9 @@ from leapp.models import VsftpdConfig, VsftpdFacts
 
 def _parse_config(path, content):
     try:
[...]

PR #511, opened 2019-05-22
Actor library loading improvement

vinzenz:
Previously we had hand crafted code for library loading. There were even
two different implementations for leapp.libraries.common libraries and
leapp.libraries.actor libraries. This patch introduces a sys.meta_path
based hooking approach and reuse as much as possible code.
With this change we are fixing additionally the problem of imports
between libraries. Previously for those imports to succeed, the names
had to be in lexicographical order, otherwise it would just fail.
This patch also fixes this problem.

Fixes https://github.com/oamg/leapp/issues/510


vinzenz:
leapp-ci build

leapp/repository/actor_definition.py
```diff
@@ -242,19 +244,10 @@ def injected_context(self):
         if self.tools:
             os.environ['LEAPP_TOOLS'] = os.path.join(self._repo_dir, self._directory, self.tools[0])
 
-        # We make a snapshot of the symbols in the module
```
vinzenz: For anyone reviewing - The library snapshotting code was actually never really useful and didn't help - The new approach is much more resilient due to python handling most of the problematic cases :-)

leapp/repository/actor_definition.py
```diff
@@ -276,19 +269,6 @@ def injected_context(self):
             else:
                 os.environ.pop('LEAPP_TOOLS', None)
 
-            # Remove all symbols in the actor lib before the execution
```
vinzenz: That's the reversal of the snapshotting

vinzenz:
e2e tests

PR #513, opened 2019-05-22
Locate and fix bug with model inheritance

fernflower:
https://github.com/oamg/leapp-repository/issues/203

Turns out base class fields aren't properly copied.
Current implementation of a fix makes eyes bleed, used purely
for bug localization.

### DIFF
--- a/leapp/models/fields/__init__.py
+++ b/leapp/models/fields/__init__.py
@@ -108,7 +108,9 @@ class Field(object):
         :type target: Instance of a Model derived class
         :return: None
         """
-        source_value = source.get(name, self._default)
+        # if base class has a field like mutable collection -> copy it, otherwise
+        # the base class and the child class may share the same object
+        source_value = copy.copy(source.get(name, self._default))
         self._validate_model_value(value=source_value, name=name)
         setattr(target, name, source_value)
 
@@ -124,7 +126,9 @@ class Field(object):
         :type target: Instance of a Model derived class
         :return: None
         """
-        source_value = source.get(name, self._default)
+        # if base class has a field like mutable collection -> copy it, otherwise
+        # the base class and the child class may share the same object
+        source_value = copy.copy(source.get(name, self._default))
         if source_value is not None:
             source_value = self._convert_to_model(source_value, name=name)
         self._validate_model_value(value=source_value, name=name)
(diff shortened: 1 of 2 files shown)


tests/scripts/test_models.py
```diff
@@ -70,7 +75,15 @@ def test_inheritance():
...
-    assert len(InheritedUnitTestModel.fields) == 2
+    assert len(InheritedUnitTestModel.fields) == 3
     assert isinstance(InheritedUnitTestModel.fields['integer'], leapp.models.fields.Integer)
     assert isinstance(InheritedUnitTestModel.fields['strings'], leapp.models.fields.List)
     assert InheritedUnitTestModel.fields is not UnitTestModel.fields
+
+
+def test_inheritance_list_field_not_shared():
+    model1 = InheritedUnitTestModel()
+    model2 = AnotherInheritedUnitTestModel()
```
vinzenz: I'd also extend the tests to ensure that they test for identity of the items fields:
```suggestion
    model2 = AnotherInheritedUnitTestModel()
    model3 = InheritedUnitTestModel()
    model4 = AnotherInheritedUnitTestModel()
    assert model3.items is not model1.items
    assert model4.items is not model2.items
    assert model1.items is not model2.items
    assert model3.items is not model4.items
```

leapp/models/fields/__init__.py
```diff
@@ -108,7 +108,9 @@ def from_initialization(self, source, name, target):
         :type target: Instance of a Model derived class
         :return: None
         """
-        source_value = source.get(name, self._default)
+        # if base class has a field like mutable collection -> copy it, otherwise
+        # the base class and the child class may share the same object
```
vinzenz: This comment is just wrong - This is about the instantiation of the Model during the runtime. This problem with the inheritance just showed more drastically a problem with the model defaults.

```python
class SomeModel(Model):
    topic = WhateverTopic
    items = fields.List(fields.String, default=[])
```
has the same problem

Try it like this:
```python
model1 = SomeModel()
model2 = SomeModel()
model1.items.append('woot')
assert model1.items is model2.items
```
The code you see here is the initialization of the runtime values. This has nothing to do anymore with inheritance.

PR #515, opened 2019-05-27
Add preupgrade subcommand + refactor config, paths and output colors

drehak:
Co-authored by: Inessa Vasilevskaya <ivasilev@redhat.com>

Breakdown of changes:
- Added `leapp preupgrade`, which stops after Reports phase and displays info about generated report
- Moved all hardcoded paths to leapp config
- Refactored colorized decorative output
- Removed redundant direct writes to stderr
- Removed redundant dashes from option definitions

Thanks to Inessa for a working proof of concept for the preupgrade.

### DIFF
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -14,36 +14,29 @@ from leapp.logger import configure_logger
 from leapp.repository.scan import find_and_scan_repositories
 from leapp.utils.audit import Execution, get_connection, get_checkpoints
 from leapp.utils.clicmd import command, command_opt
-from leapp.utils.output import report_errors, beautify_actor_exception
+from leapp.utils.output import report_errors, report_info, beautify_actor_exception
 
 
 def archive_logfiles():
     """ Archive log files from a previous run of Leapp """
-    # FIXME: The files in /var/log/leapp/ needs to be defined on one place. They are defined also in actors.
-    archive_dir = '/var/log/leapp/archive/'
-    logs_directory = '/var/log/leapp'
-    debug_dir = '/var/log/leapp/dnf-debugdata/'
-    leapp_db = '/var/lib/leapp/leapp.db'
-    log_files = [
-        '/var/log/leapp/leapp-upgrade.log',
-        '/var/log/leapp/leapp-report.txt',
-        '/var/log/leapp/dnf-plugin-data.txt'
-    ]
+    cfg = get_config()
 
-    if not os.path.isdir(logs_directory):
-        os.makedirs(logs_directory)
+    if not os.path.isdir(cfg.get('logs', 'dir')):
+        os.makedirs(cfg.get('logs', 'dir'))
 
-    files_to_archive = [f for f in log_files if os.path.isfile(f)]
+    files_to_archive = [os.path.join(cfg.get('logs', 'dir'), f)
+                        for f in cfg.get('logs', 'files').split(':')
+                        if os.path.isfile(os.path.join(cfg.get('logs', 'dir'), f))]
 
-    if not os.path.isdir(archive_dir):
-        os.makedirs(archive_dir)
+    if not os.path.isdir(cfg.get('archive', 'dir')):
+        os.makedirs(cfg.get('archive', 'dir'))
 
     if files_to_archive:
-        if os.path.isdir(debug_dir):
-            files_to_archive.append(debug_dir)
+        if os.path.isdir(cfg.get('debug', 'dir')):
+            files_to_archive.append(cfg.get('debug', 'dir'))
 
         now = datetime.now().strftime('%Y%m%d%H%M%S')
-        archive_file = os.path.join(archive_dir, 'leapp-{}-logs.tar.gz'.format(now))
+        archive_file = os.path.join(cfg.get('archive', 'dir'), 'leapp-{}-logs.tar.gz'.format(now))
 
         with tarfile.open(archive_file, "w:gz") as tar:
             for file_to_add in files_to_archive:
@@ -55,8 +48,8 @@ def archive_logfiles():
                 except OSError:
                     pass
             # leapp_db is not in files_to_archive to not have it removed
-            if os.path.isfile(leapp_db):
-                tar.add(leapp_db)
+            if os.path.isfile(cfg.get('database', 'path')):
+                tar.add(cfg.get('database', 'path'))
 
 
 def load_repositories_from(name, repo_path, manager=None):
@@ -113,7 +106,7 @@ def check_env_and_conf(env_var, conf_var, configuration):
 @command('upgrade', help='Upgrades the current system to the next available major version.')
 @command_opt('resume', is_flag=True, help='Continue the last execution after it was stopped (e.g. after reboot)')
 @command_opt('reboot', is_flag=True, help='Automatically performs reboot when requested.')
-@command_opt('--whitelist-experimental', action='append', metavar='ActorName',
+@command_opt('whitelist-experimental', action='append', metavar='ActorName',
              help='Enables experimental actors')
 def upgrade(args):
     if os.getuid():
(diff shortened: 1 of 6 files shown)


leapp/cli/upgrade/__init__.py
```diff
@@ -149,6 +150,11 @@ def upgrade(args):
         logger = configure_logger('leapp-upgrade.log')
     os.environ['LEAPP_EXECUTION_ID'] = context
 
+    until_phase = None
+    if args.pre_upgrade:
+        until_phase = "DownloadPhase"
```
pirat89: I propose to set the dedault to the report phase. As initial transaction checks will be part of the check phase. Additionally, there would be additional option like `--include-download` that would continue with shift the end until the download phase.

drehak: This is a good topic to discuss at Cabal, since others already gave me (conflicting) opinions on this. Adding to agenda.

drehak: (Just for context's sake, the solution we've decided to use is the one proposed here.)

pirat89:
I removed the WIP flag as I guess it has been just forgotten there.

pirat89:
@drehak  @fernflower : Shouldn't be supported `--whitelist-experimental` option as well, when it is supported for the `upgrade` command?

leapp/cli/upgrade/__init__.py
```diff
@@ -175,6 +175,40 @@ def upgrade(args):
...
+@command('preupgrade', help='Generate preupgrade report')
+def preupgrade(args):
+    if os.getuid():
+        raise CommandError('This command has to be run under the root user.')
+
+    context = str(uuid.uuid4())
+    configuration = {
+        'debug': os.getenv('LEAPP_DEBUG', '0'),
+        'verbose': os.getenv('LEAPP_VERBOSE', '0'),
+    }
```
pirat89: I meant, to have whitelist_experimental here.

drehak: Good point, I'll add it.

drehak: Now that you mention it, would it make sense to implement `--resume` as well or not?

vinzenz: Well not really, because resume was used for continuation after reboot or planned interruption - I don't see why that would be needed for preupgrade

leapp/cli/upgrade/__init__.py
```diff
@@ -175,6 +175,43 @@ def upgrade(args):
         sys.exit(1)
 
 
+@command('preupgrade', help='Generate preupgrade report')
+@command_opt('--whitelist-experimental', action='append', metavar='ActorName',
```
vinzenz: just FYI -- is not necessary command_opt is only for -- options - Not that it hurts anyone, it strips them away first anyway and then readds them :D

leapp/cli/upgrade/__init__.py
```diff
@@ -175,6 +175,43 @@ def upgrade(args):
...
+    archive_logfiles()
+    logger = configure_logger('leapp-preupgrade.log')
+    os.environ['LEAPP_EXECUTION_ID'] = context
+
+    try:
+        repositories = load_repositories()
+    except LeappError as exc:
+        sys.stderr.write(exc.message)
+        sys.stderr.write('\n')
+        raise CommandError(exc.message)
```
vinzenz: I don't see why you need to print the message when CommandError handling does print it as well

leapp/cli/upgrade/__init__.py
drehak: Nice to know. I just copied it from the upgrade command as closely as possible.

leapp/cli/upgrade/__init__.py
```diff
@@ -175,6 +175,43 @@ def upgrade(args):
...
+    workflow = repositories.lookup_workflow('IPUWorkflow')()
+    with beautify_actor_exception():
```
vinzenz: Beware that the lookup may return None if leapp-repository is broken or not installed. The latter however shouldn't be the case though

pirat89:
One thing that we discussed with @fernflower. Currently the phase when the preupgrade should ended is hardcoded to the specific name of the phase. I don't remember whether we had agreement to do it like that or not, just would like to ask whether to keep it like that or set up another flag (preupgrade_end?...) that would specifically mark the phase when end should happen. What do you thing? @drehak, @vinzenz , @oamg/developers? I am ok with hardcoded name even when it put requirement on specific names of phases. Just would like to ensure we haven't missed that.

leapp/cli/upgrade/__init__.py
drehak: Gone now.

drehak:
@pirat89 IMO it's perfectly fine to hardcode it.

drehak:
Yeah, I accidentally left out some chunks of the command. That should be fixed now.

How would you want the command to be tested?

drehak:
@pirat89 I've decided for displaying the message from inside the VerifyCheckResults actor, since that's the only part of our code that "sees" the report path.
Relevant: https://github.com/oamg/leapp-repository/pull/236
I still need to test these two PRs together, so don't merge yet.

leapp/utils/output.py
```diff
@@ -28,9 +28,23 @@ def report_errors(errors):
 ============================================================{reset}\n\n""".format(red=red, reset=reset))
 
 
+def report_report(path):
```
fernflower: Can you please rename the function to something more self-speaking?
report_info?

leapp/cli/upgrade/__init__.py
```diff
@@ -175,6 +173,52 @@ def upgrade(args):
...
+            msg = 'No such Actor: {}'.format(actor_name)
+            logger.error(msg)
+            raise CommandError(msg)
+    with beautify_actor_exception():
+        until_phase = 'ReportsPhase'
+        logger.info('Executing workflow until phase: %s', until_phase)
+        workflow.run(context=context, until_phase=until_phase)
+
+    report_errors(workflow.errors)
+    report_report('/var/log/leapp/leapp-report.txt')
```
fernflower: Please do not make leapp-report path a magic constant

leapp/utils/output.py
drehak: Changed to report_info.

leapp/cli/upgrade/__init__.py
drehak: It's in the config now.

pirat89:
It looks perfect from my POV when everything is ok, but when error happens, I am not sure it's good:
```
                        ERRORS

2019-06-10 22:43:49.983508 [ERROR] Actor: verify_check_results Message: Unsupported vsftpd configuration
2019-06-10 22:43:49.997420 [ERROR] Actor: verify_check_results Message: TCP wrappers support removed in the next major version
2019-06-10 22:43:50.006318 [ERROR] Actor: verify_check_results Message: Ending process due to errors found during checks, see /var/log/leapp/leapp-report.txt for detailed report.

                     END OF ERRORS


                        REPORT

A report has been generated at /var/log/leapp/leapp-report.txt

                        REPORT

```

Additional info: the REPORT separators are always green, in case of error.

drehak:
@pirat89 
> It looks perfect from my POV when everything is ok. But when error happens, I am not sure it's good:

It's just the way I intended. Why do you think it's a problem if the report was still generated?

> Additional info: the REPORT separators are always green, in case of error.

Same thing: how is that a problem? I'm no UX expert, so I don't get your point but I think the red error block still gives away that the command failed.

vinzenz:
@drehak just keep it white

drehak:
Time to wrangle the colors a bit, don't merge please.

drehak:
Squashed down to a single commit.

leapp/utils/output.py
```diff
@@ -8,37 +8,44 @@
...
-        for error in errors:
-            print_error(error)
-        sys.stdout.write("""{red}
-============================================================
-                     END OF ERRORS
-============================================================{reset}\n\n""".format(red=red, reset=reset))
+def pretty_block(string, color=Color.bold, width=60):
+    return (color + "=" * width + "\n"
+            + string.center(width) + "\n"  # noqa: W503
+            + "=" * width + Color.reset + "\n\n")  # noqa: W503
```
pirat89: As we suggest to everyone to not use concatenation of strings in this way, we should not do that as well. Anyway, this is much readable for me:

```
return "{color}{separator}\n{text}\n{separator}{reset}\n\n".format(
    color=color,
    separator="="*width,
    reset=reset,
    text=string.center(width)
```

drehak: Replaced.

PR #517, opened 2019-05-31
Validation of input parameters

drehak:


leapp/workflows/__init__.py
```diff
@@ -178,6 +184,11 @@ def run(self, context=None, until_phase=None, until_actor=None, skip_phases_unti
 
         self._errors = get_errors(context)
 
+        for phase in skip_phases_until, needle_phase:
+            if not is_valid_phase(phase):
+                self.log.error('Phase {phase} does not exist in the workflow'.format(phase=phase))
+                return
```
vinzenz: well you should then ensure that we exit with non 0 status code.

leapp/workflows/__init__.py
```diff
@@ -27,6 +27,14 @@ def _get_phases_sorted(attrs):
     return tuple(sorted([attr for attr in attrs.values() if _is_phase(attr)], key=_phase_sorter_key))
 
 
+def actor_names(actor=None):
+    return (actor.name.lower(), actor.class_name.lower()) if actor else (None)
+
+
+def phase_names(phase=None):
+    return (phase[0].__name__.lower(), phase[0].name.lower()) if phase else (None)
```
vinzenz:
```suggestion
    return (phase[0].__name__.lower(), phase[0].name.lower()) if phase else ()
```

leapp/workflows/__init__.py
```diff
@@ -27,6 +27,14 @@ def _get_phases_sorted(attrs):
     return tuple(sorted([attr for attr in attrs.values() if _is_phase(attr)], key=_phase_sorter_key))
 
 
+def actor_names(actor=None):
+    return (actor.name.lower(), actor.class_name.lower()) if actor else (None)
```
vinzenz:
```suggestion
    return (actor.name.lower(), actor.class_name.lower()) if actor else ()
```

leapp/workflows/__init__.py
```diff
@@ -141,6 +149,12 @@ def serialize(cls):
             'phases': [phase.serialize() for phase in cls.phases],
         }
 
+
+    def is_valid_phase(self, phase=None):
+        if phase:
+            return phase in [phase_names(phase) for phase in self._phase_actors]
```
vinzenz: you Need to flatten the list of tuples to be able to search in it

drehak: Should be fine now.

drehak:
Since the related issue is about general validation - rather than just workflow phase names - are there any other input parameters that might need similar validation?

vinzenz:
I am not sure if --whitelist-experimental is already validated - Please check, if not then we gotta add it too. But again It's actor names

leapp/workflows/__init__.py
vinzenz: AFAIR yes

drehak: Okay then, replaced this with a CommandError.

drehak:
Turns out it's validated. I've nuked my latest commits, hopefully nothing breaks this way.

leapp/workflows/__init__.py
```diff
@@ -178,11 +191,15 @@ def run(self, context=None, until_phase=None, until_actor=None, skip_phases_unti
 
         self._errors = get_errors(context)
 
+        for phase in skip_phases_until, needle_phase:
+            if phase and not self.is_valid_phase(phase):
```
vinzenz: Is valid phase returns None if the phase is None or empty so the is unnecessary

vinzenz: Never mind it is necessary I just realized

PR #519, opened 2019-06-05
[WIP] lint: Add some more linter checks

fernflower:


leapp/snactor/commands/repo/__init__.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
vojtechsokol: Shouldn't there be a blank line between imports from `__future__` and other imports?

Issue #474, opened 2019-03-29
when no args, help text is printed to stdout, not stderr


AloisMahdal:
**Actual behavior**
If called without arguments, `leapp` and `snactor` will print help text to the stdout instead of stderr.

**To Reproduce**
Steps to reproduce the behavior
1. install leapp
2. run `leapp 1>1 2>2`
3. See error

**Expected behavior**
Error *and* help goes to std*err*.  stdout stays empty.

**System information (please complete the following information):
 *  *leapp-0.5.0-1.201903291447Z.fcdb4f0.master.el7_6.noarch*

****

**Additional context**

A good rule od thumb: stdout should always stay empty unless it contains the particular output that was asked for.  (Stuff that are "about" execution such as errors or infos should always be in stderr.)

One use case when this could arise is if someone composes the command in a script with intention to parse the output (this is harder to imagine for leapp, but easier for eg. `snactor discover`) .  If for some reason the composition fails and help/error is returned, they must not go to stdout or they could be parsed, possibly snowballing the issue further down the line.

Madeyro:
> @drehak drehak referenced this issue on Jun 5
Merged
Print usage to stderr #521 

I think you can close this issue :)

PR #521, opened 2019-06-05
Print usage to stderr

drehak:
Related to https://github.com/oamg/leapp/issues/474.


drehak:
This should cover all the cases where usage is printed.

PR #523, opened 2019-06-05
Use jekyll redirect plugin to redirect pages.

asilveir:
- This approach is better because all pages can be redirected without code duplication


pirat89:
good work :)

PR #526, opened 2019-06-10
Add links to resources not included in RTD

vojtechsokol:


docs/source/index.rst
```diff
@@ -1,6 +1,15 @@
...
-   `Leapp <https://oamg.github.io/leapp/>`_ is an OS and application modernization framework
+=============================================
+`Leapp <https://oamg.github.io/leapp/>`_ is an OS and application modernization framework.
+
+Resources not included in this RTD:
+ - RPM dependencies for `leapp-repository <https://github.com/oamg/leapp-repository/blob/master/docs/dependencies.md>`_.
+ - Upgrade inhibition by producing a `inhibiting report message <https://github.com/oamg/leapp-repository/blob/master/docs/howto-inhibit-upgrade.md>`_.
+ - Type of tests used for `testing the actors <https://github.com/oamg/leapp-repository/blob/master/docs/tests.md>`_.
+ - NVR schema used for for leapp-related `rpm packages <https://github.com/oamg/leapp-guidelines/blob/master/build-schema.md>`_.
+ - Overview (slightly outdated) of phases used in `Inplace upgrades Workflow <https://raw.githubusercontent.com/leapp-to/leapp-to.github.io/master/img/phases.png>`_.
```
jmikovic: I wouldn't include the outdated list of phases

docs/source/index.rst
```diff
@@ -1,6 +1,15 @@
 Welcome to developer documentation for Leapp!
-=================================
-   `Leapp <https://oamg.github.io/leapp/>`_ is an OS and application modernization framework
+=============================================
+`Leapp <https://oamg.github.io/leapp/>`_ is an OS and application modernization framework.
+
+Resources not included in this RTD:
+ - RPM dependencies for `leapp-repository <https://github.com/oamg/leapp-repository/blob/master/docs/dependencies.md>`_.
+ - Upgrade inhibition by producing a `inhibiting report message <https://github.com/oamg/leapp-repository/blob/master/docs/howto-inhibit-upgrade.md>`_.
```
jmikovic: This should be turned into proper tutorial

docs/source/index.rst
```diff
@@ -1,6 +1,15 @@
...
+ - Type of tests used for `testing the actors <https://github.com/oamg/leapp-repository/blob/master/docs/tests.md>`_.
+ - NVR schema used for for leapp-related `rpm packages <https://github.com/oamg/leapp-guidelines/blob/master/build-schema.md>`_.
+ - Overview (slightly outdated) of phases used in `Inplace upgrades Workflow <https://raw.githubusercontent.com/leapp-to/leapp-to.github.io/master/img/phases.png>`_.
+   For up-to-date state, consult the `source of IPU workflow <https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py>`_.
```
jmikovic: I think this is already included in one of the tutorials

docs/source/index.rst
examon: I agree with @jmikovic , contents from the both links should be moved into RTD. I don't see the reason why they are buried inside leapp-repository.

docs/source/index.rst
```diff
@@ -1,6 +1,15 @@
...
+ - Architecture overview of `leapp and snactor tools <https://github.com/leapp-to/leapp-to.github.io/blob/master/architecture.md>`_.
```
examon: This link points to the archived repo. I think we should move the content directly to the RTD.

docs/source/index.rst
```diff
@@ -1,6 +1,15 @@
 Welcome to developer documentation for Leapp!
-=================================
-   `Leapp <https://oamg.github.io/leapp/>`_ is an OS and application modernization framework
+=============================================
+`Leapp <https://oamg.github.io/leapp/>`_ is an OS and application modernization framework.
+
+Resources not included in this RTD:
+ - RPM dependencies for `leapp-repository <https://github.com/oamg/leapp-repository/blob/master/docs/dependencies.md>`_.
+ - Upgrade inhibition by producing a `inhibiting report message <https://github.com/oamg/leapp-repository/blob/master/docs/howto-inhibit-upgrade.md>`_.
+ - Type of tests used for `testing the actors <https://github.com/oamg/leapp-repository/blob/master/docs/tests.md>`_.
```
examon: I think this should be moved/edited to the RTD testing page.

vojtechsokol:
Regarding the comments that most of the linked content should be moved to RTD - I agree, I was planning to write it to a comment, but you were faster than me. Anyway, the task this PR is part of should only involve adding links with content that is not part of the RTD. There are also other issues (e.g. dependencies.md is not linked from anywhere in RTD and is accessible only by using search function), but it should be solved in separate PR.

docs/source/index.rst
vojtechsokol: removed

docs/source/index.rst
vojtechsokol: removed

PR #527, opened 2019-06-14
WIP: Ensure non-zero exit code in case of actor runtime errors

drehak:
Right now, when an internal actor error occurs during workflow execution, leapp exits with 0. This PR is meant to change that, currently by catching the exception and adding it to errors.

### DIFF
--- a/leapp/actors/__init__.py
+++ b/leapp/actors/__init__.py
@@ -6,7 +6,7 @@ import sys
 from leapp.compat import string_types
 from leapp.dialogs import Dialog
 from leapp.exceptions import MissingActorAttributeError, WrongAttributeTypeError, StopActorExecutionError, \
-    StopActorExecution, StandardError
+    StopActorExecution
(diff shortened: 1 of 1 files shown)


leapp/actors/__init__.py
```diff
@@ -303,7 +303,7 @@ def run(self, *args):
             pass
         except StopActorExecutionError as err:
             self.report_error(err.message, err.severity, err.details)
-        except StandardError as err:
+        except LeappRuntimeError as err:
             self.report_error(err.message, ErrorSeverity.FATAL)
```
vinzenz: I wonder if we would want the stack trace as a detail, but you don't need to care about this here

vinzenz:
Handled now by #536

pirat89:
@vinzenz Just nitpick that #536 is not merged yet. Better would be to keep closing on automated process, that close this one automatically when PR is accepted (merged)

PR #529, opened 2019-06-19
Introduce json report generation

fernflower:
That is an ugly hack to make leapp preupgrade produce
a json report alongwith default text one.
Json report will be present in the same directory as the
text report.

### DIFF
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -15,6 +15,7 @@ from leapp.repository.scan import find_and_scan_repositories
 from leapp.utils.audit import Execution, get_connection, get_checkpoints
 from leapp.utils.clicmd import command, command_opt
 from leapp.utils.output import report_errors, report_info, beautify_actor_exception
+from leapp.utils.report import fetch_upgrade_report_raw
 
 
 def archive_logfiles():
@@ -25,7 +26,7 @@ def archive_logfiles():
         os.makedirs(cfg.get('logs', 'dir'))
 
     files_to_archive = [os.path.join(cfg.get('logs', 'dir'), f)
-                        for f in cfg.get('logs', 'files').split(':')
+                        for f in cfg.get('logs', 'files').split(',')
                         if os.path.isfile(os.path.join(cfg.get('logs', 'dir'), f))]
 
     if not os.path.isdir(cfg.get('archive', 'dir')):
@@ -207,10 +208,15 @@ def preupgrade(args):
 
     report_errors(workflow.errors)
 
+    report_txt, report_json = [os.path.join(get_config().get('report', 'dir'),
+                                            'leapp-report.{}'.format(f)) for f in ['txt', 'json']]
+    report_info([report_txt, report_json], fail=workflow.errors)
+    # fetch all report messages as a list of dicts
+    messages = fetch_upgrade_report_raw(context, renderers=False)
+    with open(report_json, 'w+') as f:
+        json.dump({'entries': messages}, f, indent=2)
     if workflow.errors:
-        report_info(get_config().get('report', 'path'), fail=True)
         sys.exit(1)
-    report_info(get_config().get('report', 'path'))
 
 
 @command('list-runs', help='List previous Leapp upgrade executions')
--- a/leapp/utils/report.py
+++ b/leapp/utils/report.py
@@ -20,34 +20,53 @@ def _flatten_dict(dict_):
     return new_dict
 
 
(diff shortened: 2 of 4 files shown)


leapp/cli/upgrade/__init__.py
```diff
@@ -211,6 +212,12 @@ def preupgrade(args):
         report_info(get_config().get('report', 'path'), fail=True)
         sys.exit(1)
     report_info(get_config().get('report', 'path'))
+    # fetch all report messages as a list of dicts
+    messages = fetch_upgrade_report_raw(context, renderers=False)
+    report_json = get_config().get('report', 'path').replace('.txt', '.json')
+    # XXX FIXME use leapp way of outputting things?
+    with open(report_json, 'w+') as f:
+        f.write(json.dumps({'entries': messages}))
```
vinzenz: json.dump? I don’t see why you need to create a temporary string - I know WIP and POC etc 
Just saying:-)

leapp/utils/report.py
```diff
@@ -20,34 +20,57 @@ def _flatten_dict(dict_):
...
+    :return: A list of upgrade messages of type "Report" withing the given context,
+             each message is a dict of raw data.
+    """
+    report_msgs = get_messages(names=['Report'], context=context_id)
+    if not report_msgs:
+        # return empty list to be consistent?
+        return
+
+    messages = []
+    for message in report_msgs:
```
vinzenz:
```python
    for message in report_msgs:
        data = json.loads(message.get('message').get('data'))
        merged_data = {
            'topic': message.get('topic'),
            'phase': message.get('phase'),
            'audience': data['audience'],
            'title': data['title'],
            'flags': data['flags'],
            'severity': data['severity']
        }
        if renderers:
            merged_data['renderers'] = data['renderers']
        merged_data.update(_flatten_dict(json.loads(data['detail'])))
        messages.append(merged_data)
```

fernflower: yeah, will refactor, thanks

leapp/cli/upgrade/__init__.py
```diff
@@ -211,6 +212,11 @@ def preupgrade(args):
         report_info(get_config().get('report', 'path'), fail=True)
         sys.exit(1)
     report_info(get_config().get('report', 'path'))
+    # fetch all report messages as a list of dicts
+    messages = fetch_upgrade_report_raw(context, renderers=False)
+    report_json = get_config().get('report', 'path').replace('.txt', '.json')
```
pirat89: would about
```suggestion
    report_json = get_config().get('report', 'path_json')
```
(require update of 'config' file as well)

fernflower: Honestly I see no point in defining separate paths for txt and json report, defining a 'leapp report dir' should be enough control over output. 
But replace('.txt', '.json') can be improved

drehak:
Looks good to me personally.

drehak:
Before we merge, shouldn't the json report be mentioned in the output too? Or will we make a separate PR for that?

leapp/cli/upgrade/__init__.py
```diff
@@ -207,10 +208,16 @@ def preupgrade(args):
 
     report_errors(workflow.errors)
 
+    report_txt, report_json = [os.path.join(get_config().get('report', 'dir'),
+                                            'leapp-report.{}'.format(f)) for f in ['txt', 'json']]
+    report_info(report_txt, fail=workflow.errors)
+    # fetch all report messages as a list of dicts
+    messages = fetch_upgrade_report_raw(context, renderers=False)
+    report_json = os.path.join(get_config().get('report', 'dir'), 'leapp-report.json')
```
pirat89: @fernflower  probably this line should be removed

pirat89:
@fernflower Additionally, when we are doing the change, I would like to see change in leapp-repository as well. One of reasons to have the paths defined was to have just one source of true. See the [line](https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/verifycheckresults/actor.py#L26). When you do change, I expect there is a way to get the correct path so it will not be hardcoded here. Or the actor will be modified so the path will not be needed.

pirat89:
but still I am unhappy that path is again hardcoded (and not sure how to provide the info to actors). even the dir doesn't look to me like something that's needed as it is the same one used for logs. I think that better solution would be to provide paths `report_txt` and `report_json` as with that the info doesn't have to be hardcoded anywhere else.

fernflower:
@pirat89 I get you, I created a bug/enhancement https://github.com/oamg/leapp/issues/531

PR #536, opened 2019-07-10
Exit non zero on unhandled exceptions and errors

vinzenz:
Previously on unhandled exceptions non of the applications did return
with a non zero exit code. This patch will introduce changes to the
framework and endpoint scripts to handle the situation and end with a
non zero exit code.


pirat89:
Btw (where are tests? :P)

AloisMahdal:
Looks like this fixes #536 so it should be closed.  (Downstream tests started to pass again.)

PR #538, opened 2019-07-17
Stop adding 'process-end' audit entry

Rezney:
We already have "process-start" and "process-result" so
we do not need "process-end" as it does not add any
additional value.

### DIFF
--- a/leapp/utils/audit/__init__.py
+++ b/leapp/utils/audit/__init__.py
@@ -261,7 +261,7 @@ def create_audit_entry(event, data, message=None):
     """
     Create an audit entry
 
-    :param event: Type of this event e.g. process-start or process-end but can be anything
+    :param event: Event type identifier
     :param data: Data related to Type of the event, e.g. a command and its arguments
     :param message: An optional message.
     :return:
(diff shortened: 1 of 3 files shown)


Rezney:
Original reason for digging into this was that when trying to analyse DB content using "jq" tool, it failed as "process-end" was not JSON. 

So we can either remove it or change to JSON. What do you think @vinzenz ?

vinzenz:
@AloisMahdal heads up on this, if you rely on it somehow

@Rezney I agree

leapp/utils/audit/__init__.py
```diff
@@ -283,7 +283,7 @@ def get_audit_entry(event, context):
 
     :param context: The execution context
     :type context: str
-    :param event: Type of this event e.g. process-start or process-end but can be anything
+    :param event: Type of this event e.g. process-start or process-result but can be anything
```
vinzenz: Now that I see this, this is weird wording. Type of this event ... can be anything. It sounds like it can be of any type, which it can't. Identifier for this type of event, might be more clear.

vinzenz:
Other than the documentation rewording this looks good to me.

leapp/utils/audit/__init__.py
Rezney: Makes sense. Fixed.

Rezney:
@vinzenz thanks for checking this. Also good idea to ask @AloisMahdal so we do not break anything. Let's wait for his ack...

AloisMahdal:
Sorry, my github notifications reading morale was bad lately.

Thanks for asking, but we don't query the db too much yet, and only the "messages" part and we just print it, so I don't think this should break tests.

PR #543, opened 2019-07-29
Introduce report composability, remove renderers

Rezney:
Introduce report composability, remove renderers
    
    In order to easily extend report with more fields and
    change the structure, we switch to using classes
    (primitives) with configurable paths. This commit
    also removes renderers as a standalone html report
    should substitute it.

kudos to @shaded-enmity

### DIFF
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -5,102 +5,258 @@ from leapp.topics import ReportTopic
 from leapp.libraries.stdlib.api import produce
 
 
-class Renderers(Model):
-    """
-    Model for describing renderers to be used when presenting reports.
-    """
-
-    topic = ReportTopic
-
-    html = fields.String()
-    """
-    Render report message as html
-    """
-    plaintext = fields.String()
-    """
-    Render report message as plaintext
-    """
-
-
 class Report(Model):
     """
-    Framework model used for reporting and presentation (see "renderers" field) purposes. The report can also carry
-    a special meaning using "flags" field.
+    Framework model used for reporting
     """
 
     __non_inheritable__ = True
 
     topic = ReportTopic
 
-    severity = fields.StringEnum(choices=['low', 'medium', 'high'])
-    """
-    Severity of the report entry
-    """
-
-    title = fields.String()
-    """
-    Title of the report entry
-    """
-
-    detail = fields.JSON()
-    """
-    Detail of the report entry as JSON data
-    """
-
-    renderers = fields.Model(Renderers)
-    """
-    :class:`Renderers` describe how to render this report entry
-    """
-
-    audience = fields.List(fields.StringEnum(choices=['developer', 'sysadmin']))
-    """
-    Who is the main audience of this report entry
-    """
-
-    flags = fields.List(fields.StringEnum(choices=['inhibitor']))
-    """
-    Flags can give a special meaning to each report entry, currently supported flags:
-
-        inhibitor - This report entry will inhibit the upgrade process
-    """
+    report = fields.JSON()
 
 
-def report(title=None, detail=None, renderers=None, audience=None, flags=None, severity=None):
+def _add_to_dict(data, path, value, leaf_list=False):
+    for p in path[:-1]:
+        if p not in data:
+            data[p] = {}
+        data = data[p]
+
+    if leaf_list:
+        data[path[-1]] = data.get(path[-1], []) + [value]
+    else:
+        if path[-1] in data:
+            raise ValueError('Path {} is already taken'.format('.'.join(path)))
+        data[path[-1]] = value
+
+
+class BasePrimitive(object):
     """
-    Create and produce a report entry
+    Report primitive (field) base class for dict targets (implies unique path)
+    """
+    # deepest nested key in resulting report dict
+    name = ''
 
-    For more information about the arguments, please refer to the :class:`Report` model
+    def __init__(self, value=None):
+        if not isinstance(value, str):
+            raise TypeError('Value of "{}" must be a string'.format(self.__class__.__name__))
+        self._value = value
 
-    :param title: Title of the report message
-    :type title: str
-    :param detail: Report message details
-    :type detail: dict
-    :param renderers: Available report entry renderers (e.g. html / plaintext)
-    :type renderers: dict
-    :param audience: Target audience of the report
-    :type audience: list
-    :param flags: Functionality flags (e.g. inhibitor)
-    :type flags: list
-    :param severity: Report severity
-    :type severity: str
+    @property
+    def value(self):
+        """Return report field value"""
+        return self._value
+
+    @property
+    def path(self):
+        """Return report field destination path (nested dict description)"""
+        return (self.name, )
+
+    @property
+    def to_dict(self):
+        return {self.name: self._value}
+
+    def apply(self, report):
+        """Add report entry to the report dict based on provided path"""
+        _add_to_dict(report, self.path, self.value, leaf_list=False)
+
+
+class BaseListPrimitive(BasePrimitive):
+    """
+    Report primitive (field) base class for list targets (we append value to a list)
+    """
+    def apply(self, report):
+        """Add report entry to the report dict based on provided path"""
+        _add_to_dict(report, self.path, self.value, leaf_list=True)
+
+
+class Title(BasePrimitive):
+    """Report title"""
+    name = 'title'
+
+
+class Summary(BasePrimitive):
+    """Report summary"""
+    name = 'summary'
+
+
+class Severity(BasePrimitive):
+    """Report severity"""
+    name = 'severity'
+
+    INFO = 'info'
+    LOW = 'low'
+    MEDIUM = 'medium'
+    HIGH = 'high'
+
+    def __init__(self, value=None):
+        severities = (self.INFO, self.LOW, self.MEDIUM, self.HIGH)
+        if value not in severities:
+            raise ValueError('Severity should be one of: "{}"'.format(' '.join(severities)))
+        self._value = value
+
+
+class Audience(BasePrimitive):
+    """Target audience of the report"""
+    name = 'audience'
+
+    def __init__(self, value=None):
+        if not isinstance(value, str):
+            raise TypeError('Value of "Audience" must be a string')
+        audiences = ('sysadmin', 'developer')
+        if value not in audiences:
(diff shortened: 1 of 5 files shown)


leapp/reporting/__init__.py
```diff
@@ -1,106 +1,249 @@
...
+
+class ExternalLink(BasePrimitive):
+    """External link report detail field"""
+    def __init__(self, url=None, title=None):
+        if not all(isinstance(v, str) for v in (url, title)):
+            raise TypeError('Values "url" and "title" of "{}" must be a string'.format(self.__class__.__name__))
+        self._value = {'url': url, 'title': title}
+
+    @property
+    def kind(self):
```
fernflower: It's not so intuitive to guess what kind 0 and kind 1 means. Can you rename this to something like is_somebooleanprop (is_list? is_dict?), so that the name would actually correspond to the value?

leapp/utils/report.py
```diff
@@ -1,72 +1,24 @@
...
-def fetch_upgrade_report_raw(context_id, renderers=True):
+def fetch_upgrade_report_messages(context_id):
     """
     :param context_id: ID to identify the needed messages
     :type context_id: str
-    :param renderers: whether copy or skip renderers field of original message
-    :type boolean
-    :return: A list of upgrade messages of type "Report" withing the given context,
-             each message is a dict of raw data.
+    :param renderer: How messages will be rendered (e.g html or plaintext)
```
fernflower: renderer param to be removed?

leapp/utils/report.py
```diff
@@ -1,72 +1,24 @@
...
+    :type renderer: str
+    :return: All upgrade messages of type "Report" withing the given context
     """
     report_msgs = get_messages(names=['Report'], context=context_id) or []
+    if not report_msgs:
```
fernflower: That's not needed iiuc, report_msgs will be an empty list and cycle at L19-L22 will be skipped

leapp/reporting/__init__.py
```diff
@@ -1,106 +1,260 @@
...
-    a special meaning using "flags" field.
+    Boolean to set whether the primitive leaf is supposed to be list or dict
     """
+    scalar = 0
+    vector = 1
 
-    __non_inheritable__ = True
 
-    topic = ReportTopic
+def _add(data, path, value, leaf_list=False):
```
fernflower: I really miss some basic unit tests for that. At least to make out on the spot what arrives as path.

leapp/reporting/__init__.py
Rezney: fixed

leapp/utils/report.py
Rezney: forgotten stuff

leapp/utils/report.py
Rezney: fixed

leapp/reporting/__init__.py
Rezney: tests added

leapp/reporting/__init__.py
```diff
@@ -5,102 +5,256 @@
...
+
+    INFO = 'info'
+    LOW = 'low'
+    MEDIUM = 'medium'
+    HIGH = 'high'
+
+    def __init__(self, value=None):
+        if not isinstance(value, str):
+            raise TypeError('Value of "{}" must be a string'.format(self.__class__.__name__))
+        severities = ('info', 'low', 'medium', 'high')
```
fernflower: Reuse the defined constants at L83-L86?

leapp/reporting/__init__.py
```diff
@@ -5,102 +5,256 @@
...
+class Severity(BasePrimitive):
+    """Report severity"""
+
+    INFO = 'info'
+    LOW = 'low'
+    MEDIUM = 'medium'
+    HIGH = 'high'
+
+    def __init__(self, value=None):
+        if not isinstance(value, str):
```
fernflower: Wondering if this check is necessary at all - no value other that one of 4 strings in severities will be allowed anyway. Do we really need such strict type checking (raising TypeError in one case and ValueError in another), won't check at L93 be enough?

leapp/reporting/__init__.py
```diff
@@ -5,102 +5,256 @@
...
+
+
+class ExternalLink(BasePrimitive):
+    """External link report detail field"""
+    def __init__(self, url=None, title=None):
+        if not all(isinstance(v, str) for v in (url, title)):
+            raise TypeError('Values "url" and "title" of "{}" must be a string'.format(self.__class__.__name__))
+        self._value = {'url': url, 'title': title}
+
+    @property
```
fernflower: Open question: introduce class BaseListPrimitive?

leapp/utils/report.py
```diff
@@ -1,72 +1,20 @@
...
-            'title': data['title'],
-            'flags': data['flags'],
-            'severity': data['severity']
-        }
-        if renderers:
-            merged_data['renderers'] = data['renderers']
-        detail = _flatten_dict(json.loads(data['detail']))
-        merged_data.update(detail)
-        messages.append(merged_data)
+        data = message.get('message').get('data')
```
fernflower: Open question: do we sanitize message data somewhere? In case like message = {} this will raise AttributeError.

leapp/utils/report.py
```diff
@@ -1,72 +1,20 @@
...
+        report = json.loads(json.loads(data).get('report'))
```
fernflower: Same question about sanitizing - in case there is no 'report' in data a TypeError is raised by the external json.loads

leapp/reporting/__init__.py
```diff
@@ -5,102 +5,256 @@
...
+    def path(self):
+        return ('flags',)
+
+
+class Tags(BasePrimitive):
+    """Report tags"""
+
+    ACCESSIBILITY = 'accessibility'
```
fernflower: How will those tags be used? And who will be responsible for adding a new tag if necessary?

Rezney: https://github.com/oamg/leapp-repository/blob/3c8c66eaef327c5c5eaa4e28d60aa07931d6caee/repos/system_upgrade/el7toel8/actors/authselectapply/actor.py#L40

Rezney: Mostly reviewers I would say but we will have to document it properly.

leapp/reporting/__init__.py
Rezney: @shaded-enmity what do you think. Shall wee need any extension soon? I am not against but probably in the next iteration...

leapp/reporting/__init__.py
Rezney: sure

leapp/reporting/__init__.py
Rezney: removed

leapp/reporting/__init__.py
vinzenz: For validation purposes I would change this to something else, but this is _if_ you do NOT want to support other values than listed here. Which is I feel like it is the idea, no?

```python
class Tags(BasePrimitive):
    """Report tags"""
    class _Value(object):
        def __init__(self, value):
            self.value = value

    ACCESSIBILITY = Tags._Value('accessibility')
    ....
    def __init__(self, value=None):
        if not isinstance(value, list):
            raise TypeError('Value of "Tags" must be a list')
        if not all([isinstance(v, Tags._Value) for v in value]):
            raise TypeError('Unsupported tag value passed for Report Tags.')
        self._value = value
```

leapp/reporting/__init__.py
```diff
@@ -5,102 +5,254 @@
...
+    SECURITY = 'security'
+    SELINUX = 'selinux'
+    SERVICES = 'services'
+    TIME_MANAGEMENT = 'time management'
+    TOOLS = 'tools'
+    UPGRADE_PROCESS = 'upgrade process'
+
+    def __init__(self, value=None):
+        if not isinstance(value, list):
+            raise TypeError('Value of "{}" must be a list'.format(self.__class__.__name__))
```
vinzenz: unnecessary format

leapp/reporting/__init__.py
Rezney: Wonderful. Yep, that is what I wanted.

You probably meant:

`ACCESSIBILITY = _Value('accessibility')`

and not 

`ACCESSIBILITY = Tags._Value('accessibility')`

right?

vinzenz: TBH i wasn't sure if just _Value works, but it should IIRC

leapp/utils/report.py
Rezney: I see https://github.com/oamg/leapp/blob/04e6041faf0e7852c37eff069fe6468ce23d8fae/leapp/utils/audit/__init__.py#L393

but not sure how much this covers us. @vinzenz what do you think?

vinzenz: This particular case is covered now - one shouldn't be able to get a message without data member. If so, the problem lays deeper. Actually for the intend one should probably make this:
```python
data = message['message']['data']
```
which would rather give you a KeyError instead.

leapp/utils/report.py
vinzenz: Then the message is not of type Report

leapp/reporting/__init__.py
fernflower: Imho overriding is_list method in all list-based classes that follow isn't as clear as introducing another BaseList type and using inheritance where necessary.

Rezney: @fernflower you are free to submit a PR to this PR :)

fernflower: But why have 2 if this code isn't merged yet? :)

vinzenz: Well you make a PR to his branch on his fork, he merges in his branch then. It's not 2 PRs to this repo.

Rezney: applied...

leapp/utils/report.py
Rezney: changed it to `data = message['message']['data']` apart from that, as discussed, we consider this ok...

leapp/reporting/__init__.py
Rezney: implemented

leapp/reporting/__init__.py
Rezney: I left the format only in the base class...

fernflower:
Tested, upgrade passed. Merging as it is good enough for initial and will unblock further reporting work.

PR #544, opened 2019-07-30
Fix display of generated reports

drehak:
Changes:
- report blocks aren't green in case of thrown exceptions
- only report files generated in the current preupgrade run are displayed
  - currently comparison of modification time with stamp inside workflow's Execution is used
- slight reorder of newlines in output

### DIFF
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -209,13 +209,15 @@ def preupgrade(args):
 
     report_errors(workflow.errors)
 
-    report_txt, report_json = [os.path.join(get_config().get('report', 'dir'),
-                                            'leapp-report.{}'.format(f)) for f in ['txt', 'json']]
-    report_info([report_txt, report_json], fail=workflow.errors)
+    cfg = get_config()
+    report_json = os.path.join(cfg.get('report', 'dir'), 'leapp-report.json')
     # fetch all report messages as a list of dicts
     messages = fetch_upgrade_report_raw(context, renderers=False)
     with open(report_json, 'w+') as f:
         json.dump({'entries': messages}, f, indent=2)
+    report_files = [os.path.join(cfg.get('report', 'dir'), r) for r in cfg.get('report', 'files').split(',')]
+    report_info([rf for rf in report_files if os.path.isfile(rf)], fail=workflow.failure)
+
     if workflow.failure:
         sys.exit(1)
 
--- a/leapp/config.py
+++ b/leapp/config.py
@@ -8,6 +8,20 @@ from leapp.utils.repository import find_repository_basedir
 
 
 _LEAPP_CONFIG = None
+# files that will go into the leapp-logs.tar.gz archive and get deleted each time a run is started
+_FILES_TO_ARCHIVE = [
+    'dnf-plugin-data.txt',
+    'leapp-report.json',
+    'leapp-report.txt',
+    'leapp-preupgrade.log',
+    'leapp-upgrade.log',
+]
+# files that will get reported at the end of a preupgrade run if they were created/modified during it
+_FILES_TO_REPORT = [
+    'leapp-report.json',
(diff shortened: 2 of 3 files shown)


leapp/cli/upgrade/__init__.py
```diff
@@ -211,11 +215,14 @@ def preupgrade(args):
 
     report_txt, report_json = [os.path.join(get_config().get('report', 'dir'),
                                             'leapp-report.{}'.format(f)) for f in ['txt', 'json']]
-    report_info([report_txt, report_json], fail=workflow.errors)
     # fetch all report messages as a list of dicts
     messages = fetch_upgrade_report_raw(context, renderers=False)
     with open(report_json, 'w+') as f:
         json.dump({'entries': messages}, f, indent=2)
+    report_files = [r for r in [report_txt, report_json]
+                    if os.path.isfile(r) and unix_timestamp_to_iso(os.path.getmtime(r)) > e.stamp]
```
jmikovic: since old reports are archived, there is no need to check timestamp as you will always have only reports created in the current run

```suggestion
                    if os.path.isfile(r)]
```

jmikovic:
leapp-preupgrade.log is being generated but never reported

drehak:
The timestamps will be needed, trust me. I will elaborate once I get out of this coding frenzy.

leapp/cli/upgrade/__init__.py
```diff
@@ -209,13 +213,18 @@ def preupgrade(args):
...
+    cfg = get_config()
+    report_txt, report_json = [os.path.join(cfg.get('report', 'dir'),
                                             'leapp-report.{}'.format(f)) for f in ['txt', 'json']]
-    report_info([report_txt, report_json], fail=workflow.errors)
     # fetch all report messages as a list of dicts
     messages = fetch_upgrade_report_raw(context, renderers=False)
     with open(report_json, 'w+') as f:
         json.dump({'entries': messages}, f, indent=2)
+    report_files = [os.path.join(cfg.get('report', 'dir'), r) for r in cfg.get('report', 'files').split(',')]
+    modified_report_files = [rf for rf in report_files
```
fernflower: Currently modified_report_files is [] and report_info doesn't report anything

fernflower: You'll need to reconsider timestamp comparison, the way it is now the list will remain empty:
```ipdb> [unix_timestamp_to_iso(os.path.getmtime(r)) for r in report_files]
['2019-08-08T16:22:01.372725Z', '2019-08-08T16:22:01.346725Z', '2019-08-08T16:22:01.364725Z']
ipdb> e.stamp
'2019-08-08T20:19:31.968004Z'
```

drehak: Yeah, looks like a timezone issue. I'll try to dig into that, thanks.

drehak: Okay, now it's always comparing UTC to UTC.
Tested on 3 VMs with timezone set to UTC, New York and Prague respectively - all OK.

leapp/cli/upgrade/__init__.py
fernflower: I have the same concern, why do we need timestamp filtering if reports are archived and the directory contains only the recent data?

drehak:
> But I don't quite understand why do we need timestamp filtering in first place if previous reports are archived?

My rationale behind time-based filtering was that if archiving goes wrong for some reason and residuals remain, those would get wrongly displayed as files generated in this run. That was before I found out archiving is more robust than I originally assumed, though.
Moreover, the current assumption is that reported files are always a perfect subset of archived files - this might not hold true in the future, although I can't think of a real world example now. But since both sets of files are configurable, they could be overridden... in which case, the above mentioned problem could ensue.

So overall it's more or less just an attempt to make the display of generated files more robust.

drehak:
Squashed and removed the stamp filtering.

PR #546, opened 2019-08-07
move report file generation to library

fernflower:
- json report generation moved to leapp.utils.report;
- txt report is generated as part of leapp upgrade/
  leapp preupgrade command;
- txt report generation in VerifyCheckActor will be
  made obsolete and removed in leapp-repository PR as
  well as actor's library;
- remove leapp report command (might be reverted).


leapp/cli/upgrade/__init__.py
```diff
@@ -211,11 +210,12 @@ def preupgrade(args):
...
     report_txt, report_json = [os.path.join(get_config().get('report', 'dir'),
                                             'leapp-report.{}'.format(f)) for f in ['txt', 'json']]
-    report_info([report_txt, report_json], fail=workflow.errors)
     # fetch all report messages as a list of dicts
-    messages = fetch_upgrade_report_raw(context, renderers=False)
-    with open(report_json, 'w+') as f:
-        json.dump({'entries': messages}, f, indent=2)
+    messages = fetch_upgrade_report_messages(context)
+    generate_report_file(messages, report_json)
+    # XXX FIXME original actor filtered all non-high messages leaving high/indicator only. Should we do the same?
```
Rezney: Definitely not. This has been a mistake which we need to fix...

leapp/utils/report.py
```diff
@@ -1,7 +1,6 @@
 import json
-import six
 
-from jinja2 import Template
```
Rezney: After this we should also remove jinja2 dependencies from spec but not in this PR as it any change has to be synced with leapp-repository.

leapp/utils/report.py
```diff
@@ -42,31 +39,23 @@ def fetch_upgrade_report_raw(context_id, renderers=True):
...
-    for msg_data in messages_raw:
-        template = Template(msg_data.pop('renderers').get(renderer))
-        messages_to_print.append(template.render(msg_data))
-
-    return messages_to_print
+def generate_report_file(messages_to_report, path):
+    if path.endswith(".txt"):
+        with open(path, 'w') as f:
+            for message in messages_to_report:
+                f.write('Severity: {}\n'.format(message['severity']))
```
Rezney: I am wondering what about flags, audiences, tags, external links. Some of those does not need to be in the plaintext report, e.g. tags, but links should definitely be there.

Question is, how to best render those. Currently it seems that:

`f.write('Remediation: {}\n'.format(message['detail']['remediation']))`

will be more JSON-like as ['remediation'] will return list of remediations and remediations may contain another list of commands.

We discussed this briefly with @shaded-enmity, we probably does not need to take care of formatting commands but should avoid some deep structures in our plaintext report. 

But it is matter of discussion of course.

leapp/cli/upgrade/__init__.py
fernflower: Thanks, will remove the XXX FIXME then. Now 2 files (txt and json) finally match

leapp/utils/report.py
fernflower: Sure, I'll add anything we consider necessary, for now as the initial commit I intended to keep the basic functionality as it is.

drehak:
Just wondering, should we merge this before #544 or after?

fernflower:
@drehak #544 should definitely be merged first imho

leapp/utils/report.py
```diff
@@ -18,3 +18,18 @@ def fetch_upgrade_report_messages(context_id):
...
+
+
+def generate_report_file(messages_to_report, path):
+    if path.endswith(".txt"):
+        with open(path, 'w') as f:
+            for message in messages_to_report:
+                f.write('Severity: {}\n'.format(message['severity']))
+                f.write('Title: {}\n'.format(message['title']))
+                f.write('Summary: {}\n'.format(message['summary']))
+                if message.get('detail', {}).get('remediation'):
```
shaded-enmity:
```suggestion
                if message.get('detail', {}).get('remediations'):
```

https://github.com/oamg/leapp/blob/master/leapp/reporting/__init__.py#L237

leapp/utils/report.py
```diff
@@ -18,3 +18,18 @@ def fetch_upgrade_report_messages(context_id):
...
+                    f.write('Remediation: {}\n'.format(message['detail']['remediation']))
```
shaded-enmity: Ditto from above, and apart from that `remediations` is a list of possible remediations in `{type: ..., value: ...}` format so we probably want to do something like this here
```
['{type}\n  {value}'.format(**r) for r in message['detail']['remediations']]
```

leapp/utils/report.py
```diff
@@ -18,3 +18,19 @@ def fetch_upgrade_report_messages(context_id):
...
+
+def generate_report_file(messages_to_report, path):
+    if path.endswith(".txt"):
+        with open(path, 'w') as f:
+            for message in messages_to_report:
+                f.write('Severity: {}\n'.format(message['severity']))
+                f.write('Title: {}\n'.format(message['title']))
+                f.write('Summary: {}\n'.format(message['summary']))
+                if message.get('detail', {}).get('remediations'):
+                    f.write('Remediation: {}\n'.format(','.join(["{}: {}".format(r['type'], r['context'])
```
shaded-enmity: Ahh, so when `type == command` we get this:
```
[u'yum', u'remove', u'leapp leapp-deps leapp-pes-data leapp-repository leapp-repository-deps leapp-repository-sos-plugin python2-leapp']
```
So I guess that we want something like `' '.join(...)` -- I think that we want to move this to a separate function like `_write_remediations()` for readability

leapp/utils/report.py
```diff
@@ -18,3 +18,19 @@ def fetch_upgrade_report_messages(context_id):
...
+                                                                 for r in message['detail']['remediations']])))
+                f.write('-' * 40 + '\n')
+    elif path.endswith(".json"):
+        with open(path, 'w') as f:
+            json.dump({'entries': messages_to_report}, f, indent=2)
```
shaded-enmity: Can we also get here the UUID of the run as stored in DB?

```suggestion
            json.dump({'entries': messages_to_report, 'leapp_run_id': id}, f, indent=2)
```

leapp/utils/report.py
fernflower: I'd rather have a class method like repr() for each Remediation type than one function, but you are right

PR #548, opened 2019-08-14
Fix _get_actor

Rezney:
return valid actor when there is another actor which name
is a substring of the actor in context.

E.g:
"persistentnetnames" vs. "persistentnetnamesdisable"


drehak:
LGTM.

PR #549, opened 2019-08-23
Handle string types in compatible way

shaded-enmity:
i.e. make sure that we can pass `unicode` in Py2 context

### DIFF
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -3,6 +3,7 @@ import json
 from leapp.models import fields, Model
 from leapp.topics import ReportTopic
 from leapp.libraries.stdlib.api import produce
+from leapp.compat import string_types
 
 
 class Report(Model):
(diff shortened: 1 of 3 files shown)


pirat89:
@shaded-enmity It looks that at least these lines should be fixed as well:
```
leapp/utils/audit/__init__.py:        if not isinstance(configuration, str):
leapp/libraries/stdlib/call.py:    elif isinstance(stdin, str):
```

leapp/reporting/__init__.py
```diff
@@ -3,6 +3,7 @@
 from leapp.models import fields, Model
 from leapp.topics import ReportTopic
 from leapp.libraries.stdlib.api import produce
+from leapp.compat import string_types
```
fernflower: Please rearrange imports in alphabetical order. 

Hmm, wondering why didn't linter catch that?

fernflower:
leapp-ci build

Rezney:
Looks good to me. Running some tests...

PR #559, opened 2019-09-06
Allow testing actors with config models

Rezney:
Currently, when tested actor runs and tries to access workflow
config, it fails with WorkflowConfigNotAvailable. Apart from
feeding the actor a mock config model, we need to also make
it aware it should use the config.

### DIFF
--- a/leapp/snactor/fixture.py
+++ b/leapp/snactor/fixture.py
@@ -95,13 +95,21 @@ class ActorContext(object):
         for model in models:
             self._messaging.feed(model, self)
 
-    def run(self):
+    def run(self, config_model=None):
         """
         Execute the current actor.
 
+        :param config_model: Config model for the actor to consume.
+        :type config_model: Config model instance derived from :py:class:`leapp.models.Model`
         :return: None
         """
-        self._actor(messaging=self._messaging).run()
+        config_model_cls = config_model.__class__ if config_model else None
+        if config_model:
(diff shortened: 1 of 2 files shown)


Rezney:
Strange, it looked like it was working for me during my POC but now it is not. Setting WIP then...

leapp/snactor/fixture.py
```diff
@@ -95,13 +95,16 @@ def feed(self, *models):
         for model in models:
             self._messaging.feed(model, self)
 
-    def run(self):
+    def run(self, config_model=None):
         """
         Execute the current actor.
```
vinzenz: Once you're done, just don't forget to update the docstring here :-)

Rezney: @vinzenz  Sure thing. I changed the code a little in order to not force people to "feed()" the config. 

So in the test it would just be:

```
current_actor_context.run(mock_configs.CONFIG_ALL_SIGNED)
```

Could you please take a look at it whether it is not too hacky (probably is)?

vinzenz: seems fine to me

Rezney:
@vinzenz  yep I will confirm once it is in order...

Rezney:
@vinzenz should be hopefully ok now...

Rezney:
https://github.com/oamg/leapp-repository/pull/303 as an usage example...

PR #552, opened 2019-08-30
Improve failed command error message

bocekm:
Most of the actors catching the `CalledProcessError as e` print `str(e)` translating to a suboptimal and non-descriptive message _"A Leapp Command Error occurred."_
Real-world example:
> Removing mount directory /var/lib/leapp/scratch/mounts/source_overlay/var/cache/dnf failed with: A Leapp Command Error occurred. 

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1746956

Also changing the way of printing the called command (`' '.join(args)` -> `str(args)`) as it has caused some confusion about missing or not missing quotation marks: https://bugzilla.redhat.com/show_bug.cgi?id=1718161

PR #563, opened 2019-09-11
Log external command errors to audit

AloisMahdal:
There's no reason why external command errors should not be logged in
audit.  In fact, having the error messages there may help understand the
context in which the command has failed.

### DIFF
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -166,7 +166,9 @@ def run(args, split=False, callback_raw=_console_logging_handler, callback_lineb
     :rtype: dict
     """
     if not args:
-        raise ValueError('Command to call is missing.')
+        message = 'Command to call is missing.'
+        api.current_logger().error(message)
+        raise ValueError(message)
     api.current_logger().debug('External command has started: {0}'.format(str(args)))
     _id = str(uuid.uuid4())
     result = None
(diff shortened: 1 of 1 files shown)


AloisMahdal:
Note that builds on #552

AloisMahdal:
e2e tests

leapp/libraries/stdlib/__init__.py
```diff
@@ -166,16 +166,20 @@ def run(args, split=False, callback_raw=_console_logging_handler, callback_lineb
...
+        raise ValueError(message)
     api.current_logger().debug('External command has started: {0}'.format(str(args)))
     _id = str(uuid.uuid4())
     result = None
     try:
         create_audit_entry('process-start', {'id': _id, 'parameters': args, 'env': env})
         result = _call(args, callback_raw=callback_raw, callback_linebuffered=callback_linebuffered, env=env)
         if checked and result['exit_code'] != 0:
+            message = 'Command {0} failed with exit code {1}.'.format(str(args), result.get('exit_code'))
+            api.current_logger().error(message)
```
vinzenz: This is not necessarily an error and should not be logged as such. Maximum debug as well, as the caller needs to handle the exception and deal with it how to treat it.

AloisMahdal: It's already being raised as `CalledProcessError`; I'm just adding the logging part which is simply consistent with that notion.

AloisMahdal: (Come to think about it, the error message could be useful if the CPE is _not_ caught, ie. somewhere back in leapp.  That would be PR for another day though...)

AloisMahdal:
BTW I'm also testing this with updated version of the downstream integration test suite.  Passed as of d77e012.

PR #564, opened 2019-09-13
Add snactor sanity-check command

Rezney:
Currently, in the command, we check whether we do not attempt
to consume a message before it is produced.

This will be added to our CI to perform a check on every PR.

### DIFF
--- a/leapp/snactor/commands/workflow/sanity_check.py
+++ b/leapp/snactor/commands/workflow/sanity_check.py
@@ -1,0 +1,43 @@
+from __future__ import print_function
+import sys
+
+from leapp.exceptions import LeappError, CommandError
+from leapp.logger import configure_logger
+from leapp.repository.scan import find_and_scan_repositories
+from leapp.snactor.commands.workflow import workflow
+from leapp.utils.clicmd import command_arg
+from leapp.utils.repository import requires_repository, find_repository_basedir
+
+_DESCRIPTION = 'The following messages are attempted to be consumed before they are produced: {}'
+_LONG_DESCRIPTION = '''
+Perform workflow sanity checks
+
+- check whether there is a message in the given workflow which is attempted to be consumed before it was produced
+
+For more information please consider reading the documentation at:
+https://red.ht/leapp-docs
+'''
+
+
+@workflow.command('sanity-check', help='Perform workflow sanity checks', description=_LONG_DESCRIPTION)
+@command_arg('name')
+@requires_repository
+def cli(params):
+    configure_logger()
+    repository = find_and_scan_repositories(find_repository_basedir('.'), include_locals=True)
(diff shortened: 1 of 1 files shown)


leapp/snactor/commands/workflow/sanity_check.py
```diff
@@ -0,0 +1,43 @@
...
+            sys.exit(1)
+
+        wf = repository.lookup_workflow(params.name)
+        if not wf:
+            raise CommandError('Could not find any workflow named "{}"'.format(params.name))
+
+        instance = wf()
+        produced_late = set(instance.initial).intersection(set(instance.produced))
+        if produced_late:
+            sys.stderr.write(_DESCRIPTION.format(' '.join(produced_late)))
```
vinzenz: produced_late are models classes, so you would want to produce a list of names here instead i guess.
Something like:
```suggestion
            sys.stderr.write(_DESCRIPTION.format(' '.join([m.__name__ for m in produced_late])))
```

leapp/snactor/commands/workflow/sanity_check.py
```diff
@@ -0,0 +1,43 @@
...
+For more information please consider reading the documentation at:
+https://red.ht/leapp-docs
+'''
+
+
+@workflow.command('sanity-check', help='Perform workflow sanity checks', description=_LONG_DESCRIPTION)
+@command_arg('name')
+@requires_repository
+def cli(params):
+    def impl(context=None):
```
vinzenz: you gotta get rid of this
```suggestion
```

leapp/snactor/commands/workflow/sanity_check.py
Rezney: Sure thing, was about to test it (sorry, did not put WIP)

leapp/snactor/commands/workflow/sanity_check.py
Rezney: thanks, will do

Rezney:
@vinzenz should be ok now. Just switched sys.stderr for print to provide a new line more conveniently...

leapp/snactor/commands/workflow/sanity_check.py
```diff
@@ -0,0 +1,43 @@
...
+        sys.stderr.write(exc.message)
+        sys.stderr.write('\n')
+        sys.exit(1)
+
+    wf = repository.lookup_workflow(params.name)
+    if not wf:
+        raise CommandError('Could not find any workflow named "{}"'.format(params.name))
+
+    instance = wf()
+    produced_late = set(instance.initial).intersection(set(instance.produces))
```
fernflower: Nit: set(instance.initial) & set(instance.produces)

PR #566, opened 2019-09-23
Dynamic phase skipping support

vinzenz:
This patch introduces support for actors being able to let leapp skip
phases.
Phases can be skipped by an actor using leapp.messaging.commands.SkipPhasesUntilCommand
like this within the process method:
```python
self._messaging.command(SkipPhasesUntilCommand(until_phase='report'))
```

This will make leapp skip phases until it hits the 'report' phase.

### DIFF
--- a/leapp/messaging/__init__.py
+++ b/leapp/messaging/__init__.py
@@ -12,6 +12,7 @@ from leapp.dialogs.renderer import CommandlineRenderer
 from leapp.messaging.answerstore import AnswerStore
 from leapp.exceptions import CannotConsumeErrorMessages
 from leapp.models import ErrorModel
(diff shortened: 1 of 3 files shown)


bocekm:
Wouldn't it be cleaner to introduce a separate module Commands instead of adding it to the Messaging module?

pirat89:
@bocekm We discussed better solutions, but this is the one we will use now. It's not public solution: see the underscore in the name `_command`. So it will not be even documented. So we can use it now and we will improve it in future when we have time.

pirat89:
Tested. Only issue is that it will skip even the mentioned phase. But we can use it in the repository as it is.

PR #567, opened 2019-10-02
Fix doc reference to StopActorExecutionError

bocekm:
The issue can be seen at https://leapp.readthedocs.io/en/latest/best-practices.html#use-convenience-exceptions-to-stop-the-actor-s-execution. The link is broken.


pirat89:
leapp-ci build

bocekm:
Created RHELLEAPP-1867 to deal with the Travis CI issue.

pirat89:
Broken travis CI for python 3 because of deps on six 1.12 (but only 1.11 is available). I am going to merg this anyway as it is unreslated to the issue. Reported issue: https://github.com/oamg/leapp/issues/568

PR #571, opened 2019-10-22
Packaging: Add /var/log/leapp into the leapp rpm

pirat89:
We've got feedback that this file is needed for purposes of possible frontends (e.g. in cockpit). To not put obstacles here, adding the directory into the rpm, so it will exist always.

NOTE: It's not common practice to own anything in /var/log in the rpm, but as the path is documented (and there are other existing examples of rpms that are doing that, e.g. systemd) I think it could be ok to add it inside leapp.


pirat89:
@vinzenz Obviously the master is broken (pylint fail). As the failed CI is not relevant to this PR, I suggest to merge it

PR #574, opened 2019-11-12
Change verbosity arguments from the base command to the upgrade related commands

alexandrepossebom:
Change verbosity arguments from the base command to the upgrade related commands

The verbose and debug options it's now only available in commands where make sense to
increase verbosity.

### DIFF
--- a/leapp/cli/__init__.py
+++ b/leapp/cli/__init__.py
@@ -7,14 +7,8 @@ from leapp import VERSION
 
 
 @command('')
-@command_opt('debug', is_flag=True, help='Enable debug mode', inherit=True)
-@command_opt('verbose', is_flag=True, help='Enable verbose logging', inherit=True)
-def cli(args):
-    os.environ['LEAPP_DEBUG'] = '1' if args.debug else os.environ.get('LEAPP_DEBUG', '0')
-    if os.environ['LEAPP_DEBUG'] == '1' or args.verbose:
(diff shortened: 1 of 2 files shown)


bocekm:
Tested it and it works fine. I'll take a look at the code later.

leapp/cli/__init__.py
```diff
@@ -7,14 +7,8 @@
...
-@command_opt('debug', is_flag=True, help='Enable debug mode', inherit=True)
-@command_opt('verbose', is_flag=True, help='Enable verbose logging', inherit=True)
-def cli(args):
-    os.environ['LEAPP_DEBUG'] = '1' if args.debug else os.environ.get('LEAPP_DEBUG', '0')
-    if os.environ['LEAPP_DEBUG'] == '1' or args.verbose:
-        os.environ['LEAPP_VERBOSE'] = '1'
-    else:
-        os.environ['LEAPP_VERBOSE'] = os.environ.get('LEAPP_VERBOSE', '0')
+def cli(args): # noqa; pylint: disable=unused-argument
+    pass
```
bocekm: @vinzenz, do you know how to get rid of this function here? I haven't studied the cli code much but it seems to me it has no meaning here now.

bocekm:
Thanks, Alex. We've created a new PR #590 for us to be able to finish it.

Issue #575, opened 2019-11-14
preupgrade report doesn't contain workflow.errors


fernflower:
In case something pretty severe is found (like absence of proper subscription) it is not reflected in the report

PR #576, opened 2019-11-15
Make error messages part of preupgrade report

fernflower:
Previously workflow errors weren't shown in the
preupgrade report. This patch attempts to solve the
issue by introducing naive one way convertion between
ErrorModel and Report during fetch_report_messages
stage at the end of the workflow run. Database won't be
altered, no new objects added after that convertion.

A follow up may be suggested to refactor existing
errors with "steps needed\hints" information into proper
Report messages.

Closes-Bug: #575

### DIFF
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -1,7 +1,7 @@
 import json
 
 from leapp.compat import string_types
-from leapp.models import fields, Model
+from leapp.models import fields, Model, ErrorModel
 from leapp.topics import ReportTopic
 from leapp.libraries.stdlib.api import produce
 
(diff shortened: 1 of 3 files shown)


vinzenz:
I am wondering if it is not easier just to make the conversion of to the report here: 
https://github.com/oamg/leapp/blob/a83c46fa8ad7f623c06366960101a0db7fd3b88a/leapp/messaging/__init__.py#L107

fernflower:
> I am wondering if it is not easier just to make the conversion of to the report here:
> 
> https://github.com/oamg/leapp/blob/a83c46fa8ad7f623c06366960101a0db7fd3b88a/leapp/messaging/__init__.py#L107
This would result in 2 messages with the same content being produced - Error and Report, and I wanted to be minimally invasive. But I can look into that for sure

drehak:
Looks good to me, though I'm not feeling so confident about that. It would be best if another dev looked at this PR.

leapp/reporting/__init__.py
```diff
@@ -269,15 +269,28 @@ def _sanitize_entries(entries):
...
+    """
+    produce(_create_report_object(entries))
+
+
+def create_report_from_error(error_dict):
+    """
+    Convert error json to report json
+    """
+    error = ErrorModel.create(error_dict)
+    entries = [Title(error.message), Summary(error.details or ""), Severity('high'), Audience('sysadmin')]
```
drehak: Even though Summary itself is older code - why is it Summary here and Detail in the error display at the end of a leapp run?

drehak: Not to mention that the dict labeled as `Detail` often contains a `details` key.

PR #577, opened 2019-11-18
Display a warning when upgrade is unsupported

drehak:
A complement of:
https://github.com/oamg/leapp-repository/pull/404
If the inhibition of unsupported upgrade is bypassed through `LEAPP_UNSUPPORTED`, a giant yellow warning appears.

### DIFF
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -14,7 +14,7 @@ from leapp.logger import configure_logger
 from leapp.repository.scan import find_and_scan_repositories
 from leapp.utils.audit import Execution, get_connection, get_checkpoints
 from leapp.utils.clicmd import command, command_opt
-from leapp.utils.output import report_errors, report_info, beautify_actor_exception
+from leapp.utils.output import report_errors, report_info, beautify_actor_exception, report_unsupported
 from leapp.utils.report import fetch_upgrade_report_messages, generate_report_file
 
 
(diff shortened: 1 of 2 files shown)


leapp/cli/upgrade/__init__.py
```diff
@@ -117,6 +117,21 @@ def generate_report_files(context):
     generate_report_file(messages, context, report_txt)
 
 
+def warn_if_unsupported():
+    env = os.environ
+    devel_vars = {k: env[k] for k in env if k.startswith('LEAPP_DEVEL_')}
```
fernflower: Hmmm iirc the final agreement was to have some special variable that would control if the upgrade is supported, not to check LEAPP_DEVEL_* vars. @vinzenz @shaded-enmity ?

leapp/utils/output.py
```diff
@@ -35,7 +36,7 @@ def print_error(error):
 def report_errors(errors):
     if errors:
         sys.stdout.write(pretty_block("ERRORS", color=Color.red))
-        sys.stderr.write("\n")
+        sys.stdout.write("\n")
```
fernflower: Again, why that change, stderr->stdout?

leapp/cli/upgrade/__init__.py
vinzenz: Checking for those is ok, but it should not inhibit when the "LEAPP_I_UNDERSTAND_THE_SYSTEM_IS_UNSUPPORTED" (or something like this) is set to 1

leapp/utils/output.py
drehak: Because everything else in this file is going to stdout (see line above).

leapp/cli/upgrade/__init__.py
drehak: It's like that now.

pirat89:
Now I am pretty confused why the piece of code is here when it should be in leapp-repository?

drehak:
@pirat89 No, that's different. This one prints a warning at the start, while the leapp-repository one produces the actual inhibitors.
As far as I know, printing from actors is a terrible practice and I haven't found a good way to gather this info in just leapp or leapp-repository.

pirat89:
@drehak The point is, when you raise inhibitor, it's already documented - they should open the report file anyway, so why to add additional prints into the leapp? This was the original idea before we agreed on inhibition of IPU. Now, when upgrade is inhibit when LEAPP_DEVEL_* or --whitelist-experimental is used, it doesn'T make sense to put additional warning into the leapp. WDYT guys? (@shaded-enmity, @vinzenz, @bocekm, @fernflower, @Rezney, @oamg/developers ).

drehak:
@fernflower 
> 
> I'm wondering if we should also show have a report entry about unsupported upgrade.

Yeah, it won't hurt. I'll add it to the other. Just check the specifics (like description, tags and severity) when the commit is there, because I'm not that sure what to put there.

PR #578, opened 2019-11-19
lint: fix py3 job

fernflower:
Travis ci py3 check job was failing because of
some violations (supposedly merged in pre-release hassle
when ci wasn't reliable together with py3 check enablement).

### DIFF
--- a/leapp/snactor/commands/discover.py
+++ b/leapp/snactor/commands/discover.py
@@ -111,7 +111,7 @@ def cli(args):
         sys.stderr.write('\n')
         sys.exit(1)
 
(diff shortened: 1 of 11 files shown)


leapp/snactor/commands/discover.py
```diff
@@ -111,7 +111,7 @@ def cli(args):
         sys.stderr.write('\n')
         sys.exit(1)
 
-    actors = [actor for actor in repository.actors]
+    actors = repository.actors
```
pirat89: This is cute :D

PR #582, opened 2019-11-29
Introduce skip_dialogs functionality

fernflower:
In order to make leapp preupgrade non-interactive by
default the workflow and actor being executed need to
be aware of the mode they are processed in (preupgrade or
upgrade workflow) and skip dialogs accordingly.

This is controlled by the skip_dialogs method variable which
if set to False by default and is propagated into every
workflow.run and actor initialization method.

The preupgrade cli command now sets skip_dialogs=True for
workflow.run which makes all dialogs present to be skipped and
answerfile with all options commented out to be generated while
the upgrade command keeps the old behavior allowing questions to
be asked interactively if no option has been chosen and saved in
answerfile.

### DIFF
--- a/tests/scripts/test_workflow.py
+++ b/tests/scripts/test_workflow.py
@@ -26,7 +26,7 @@ def test_workflow(repository):
         assert workflow.phase_actors
         assert not workflow.consumes
         assert not workflow.produces
-        workflow.run()
+        workflow.run(skip_dialogs=True)
         test_log_file.seek(0)
         order = [json.loads(line.decode('utf-8'))['class_name'] for line in test_log_file]
         assert order.pop(0) == 'FirstActor'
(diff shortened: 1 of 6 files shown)


pirat89:
Heh. I gave it approaval instead of comment O:-) But I am keeping it opened.

tests/scripts/test_workflow.py
```diff
@@ -161,3 +161,12 @@ def test_workflow_error_policy_fail_phase(repository):
...
         assert workflow.errors and len(workflow.errors) == 2
+
+
+def test_workflow_render_dialogs(repository):
+    workflow = repository.lookup_workflow('UnitTest')()
+    with tempfile.NamedTemporaryFile(mode='w') as stdin_dialog:
+        stdin_dialog.write('Yes\n')
+        stdin_dialog.seek(0)
+        with mock.patch('sys.stdin.fileno', return_value=stdin_dialog.fileno()):
+            workflow.run(skip_dialogs=False)
```
pirat89: Just for the others, the PR is ok, we are just trying to figure out how to check the stdout.

Issue #579, opened 2019-11-20
`run` function in stdlib should expose stdin


shaded-enmity:
It's already supported by the lower-level `_call` function, we just need to add this parameter to the `run` interface.

PR #583, opened 2019-12-03
libraries.stdlib.run: expose stdin

pirat89:
Add possibility to setup stdin (string or fd) when run subcommand
using the libraries.stdlib.run function. That's needed e.g. when
secret data should be used as input for the command and we do not
want to expose those data anywhere.

Fixes: #579


pirat89:
@fernflower yy. That's why I didn't set review request yet :)

pirat89:
Done. I took the tests for `_call`. Just applying them to the `run` function so it should be clear that stdin is exposed. I have thought about parametrize original call tests instead of copy&paste, to apply those for both functions (_call, run), but regarding the current state, I will keep them separated.

PR #584, opened 2019-12-03
docs: reporting.Title best practices

fernflower:
Require Title to be generic for reports of the same
semantics.
Also some minor refactor of reporting section has been
done.

### DIFF
--- a/docs/source/actor-rhel7-to-rhel8.md
+++ b/docs/source/actor-rhel7-to-rhel8.md
@@ -15,7 +15,7 @@ The main tools you will use for the actor development are listed below.
 
 ### leapp
 
-The leapp framework uses provides the libraries required to be imported by any actor and also a binary tool used to control the execution of actors within a workflow.  
+The leapp framework provides the libraries required to be imported by any actor and also a binary tool used to control the execution of actors within a workflow.
 
 ### snactor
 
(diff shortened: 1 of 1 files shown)


docs/source/actor-rhel7-to-rhel8.md
```diff
@@ -15,7 +15,7 @@ The main tools you will use for the actor development are listed below.
 
 ### leapp
 
-The leapp framework uses provides the libraries required to be imported by any actor and also a binary tool used to control the execution of actors within a workflow.  
+The leapp framework uses provides the libraries required to be imported by any actor and also a binary tool used to control the execution of actors within a workflow.
```
Rezney: I know this was not done but you but we can fix it here.

> uses provides the libraries

Rezney: apart from this looks good to me...

docs/source/actor-rhel7-to-rhel8.md
```diff
@@ -161,20 +163,40 @@ class MyNewActor(Actor):
...
+create_report([
+    reporting.Title(
+-        'Upgrade process was interrupted because a deprecated module {0} is enabled'.format(module),
++        'Upgrade process was interrupted because a deprecated module is enabled',
+    reporting.Summary(
+        'Module {0} was surpassed by shiny-new-{0} and therefore it was '
+        'removed from RHEL-8. Keeping it in the configuration may '
+        'lock out the system thus it is necessary to disable it '
+        'before the upgrade process can continue.'.format(module)
+    ),
```
pirat89: I would go even further and skip generating of multiple reports for the same thing So summary would looks like:
```diff
-        'Module {0} was surpassed by shiny-new-{0} and therefore it was '
-        'removed from RHEL-8. Keeping it in the configuration may '
-        'lock out the system thus it is necessary to disable it '
-        'before the upgrade process can continue.'.format(module)
+        'Modules below have been surpassed by new modules and therefore these '
          ....
+        'before the upgrade process can continue. The list of affected modules:'
+        '    - modulA  -> Amod'
+        '    - modulB  -> Bmod'
```

pirat89: or it could start as well as 'We detected deprecated modules that are replaced....'

fernflower: I'd say we shouldn't restrict 1-type-1-report-entry-per-actor (imagine we have a remediation for each module, some are hints, some are commands, but they must be run in specific order), but I'll add your example as another way of fixing the trouble.

docs/source/actor-rhel7-to-rhel8.md
fernflower: `The leapp framework provides the libraries` will do?

docs/source/actor-rhel7-to-rhel8.md
pirat89: @fernflower But in this very specific case, it's just about names of modules. everythig else is same. I think we could later create list of examples, documenting several cases and explain why we choose to report them in this of another way.

docs/source/actor-rhel7-to-rhel8.md
pirat89: I guess the original idea was: `The leapp framework uses the provided libraries ...`.  But @fernflower seems better to me.

Rezney: @fernflower sounds good...

PR #586, opened 2019-12-10
Introduce workflow.dialogs

fernflower:
Previously answerfile was generated from all the dialogs
present in the actors of the workflow, regardless of
whether this or that dialog has actually been shown.
This change collects dialogs actually shown to the user
and populates answerfile accordingly.

### DIFF
--- a/leapp/messaging/__init__.py
+++ b/leapp/messaging/__init__.py
@@ -31,6 +31,7 @@ class BaseMessaging(object):
         self._errors = self._manager.list()
         self._stored = stored
         self._config_models = (config_model,) if config_model else ()
+        self._dialogs = self._manager.list()
 
     def load_answers(self, answer_file, workflow):
         """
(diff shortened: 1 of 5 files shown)


leapp/messaging/__init__.py
```diff
@@ -58,6 +59,13 @@ def stored(self):
         """
         return self._stored
 
+    def dialogs(self):
+        """
+        Gets all dialogs actually encountered during workflow run
```
pirat89:
```suggestion
        Gets all dialogs actually encountered during workflow run
        
```

Please keep the required format of docstring:
```
One line title

N lines of whatever after
the empty line.
```

fernflower: You mean the newline missing between description\return? Ok, will fix that everywhere in the file.

pirat89: Yes, thanks.

PR #590, opened 2020-01-06
Change verbosity arguments from the base command to the upgrade related commands

drehak:
Change verbosity arguments from the base command to the upgrade related commands

The verbose and debug options it's now only available in commands where make sense to
increase verbosity.

A re-request of https://github.com/oamg/leapp/pull/574 by @alexandrepossebom, since he left.

### DIFF
--- a/leapp/cli/__init__.py
+++ b/leapp/cli/__init__.py
@@ -7,14 +7,8 @@ from leapp import VERSION
 
 
 @command('')
-@command_opt('debug', is_flag=True, help='Enable debug mode', inherit=True)
-@command_opt('verbose', is_flag=True, help='Enable verbose logging', inherit=True)
-def cli(args):
-    os.environ['LEAPP_DEBUG'] = '1' if args.debug else os.environ.get('LEAPP_DEBUG', '0')
-    if os.environ['LEAPP_DEBUG'] == '1' or args.verbose:
-        os.environ['LEAPP_VERBOSE'] = '1'
-    else:
-        os.environ['LEAPP_VERBOSE'] = os.environ.get('LEAPP_VERBOSE', '0')
+def cli(args): # noqa; pylint: disable=unused-argument
+    pass
 
 
 def main():
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -124,15 +124,28 @@ def warn_if_unsupported(configuration):
         report_unsupported(devel_vars, configuration["whitelist_experimental"])
 
 
+def handle_output_level(args):
+    """
+    Set environment variables following command line arguments.
+    """
+    os.environ['LEAPP_DEBUG'] = '1' if args.debug else os.environ.get('LEAPP_DEBUG', '0')
+    if os.environ['LEAPP_DEBUG'] == '1' or args.verbose:
+        os.environ['LEAPP_VERBOSE'] = '1'
+    else:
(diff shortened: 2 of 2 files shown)


leapp/cli/__init__.py
```diff
@@ -7,14 +7,8 @@
...
-@command_opt('debug', is_flag=True, help='Enable debug mode', inherit=True)
-@command_opt('verbose', is_flag=True, help='Enable verbose logging', inherit=True)
-def cli(args):
-    os.environ['LEAPP_DEBUG'] = '1' if args.debug else os.environ.get('LEAPP_DEBUG', '0')
-    if os.environ['LEAPP_DEBUG'] == '1' or args.verbose:
-        os.environ['LEAPP_VERBOSE'] = '1'
-    else:
-        os.environ['LEAPP_VERBOSE'] = os.environ.get('LEAPP_VERBOSE', '0')
+def cli(args): # noqa; pylint: disable=unused-argument
+    pass
```
pirat89: re-raising original post: @vinzenz, do you know how to get rid of this function here? I haven't studied the cli code much but it seems to me it has no meaning here now.

bocekm: @vinzenz, could you please add a comment explaining why this function needs to be here. Thanks. We can just leave this strange code here but it should be at least properly commented.

vinzenz: this function used to handle debug and verbose for all sub commands by setting the environment variable, so not every sub command had to implement this.

vinzenz: and i am not sure what is so strange about it, it was just a way of not having to do the same stuff every time

leapp/cli/upgrade/__init__.py
```diff
@@ -198,13 +210,14 @@ def upgrade(args):
...
-             help='Enables experimental actors')
+@command_opt('whitelist-experimental', action='append', metavar='ActorName', help='Enables experimental actors')
 @command_opt('save-answerfile', help='Path to save custom answerfile')
+@command_opt('debug', is_flag=True, help='Enable debug mode', inherit=False)
+@command_opt('verbose', is_flag=True, help='Enable verbose logging', inherit=False)
 def preupgrade(args):
     if os.getuid():
         raise CommandError('This command has to be run under the root user.')
-
+    parse_verbose(args)
```
vinzenz: you're not really parsing it, but handling it

vinzenz: also it handles debug too, so this function name is kind of misleading

leapp/cli/__init__.py
bocekm: I understand the original purpose, but now it has lost it. I and Alex were trying to get rid of it but with no success.

drehak: Alright, we'll keep it this way.

vinzenz: > ... with no success.

could you elaborate on that?

leapp/cli/upgrade/__init__.py
drehak: Good points, replaced with `handle_output_level`.

drehak:
Applied feedback and squashed.

leapp/cli/__init__.py
bocekm: > could you elaborate on that?
I don't remember what constructs I was trying but basically whatever change I tried to remove the function warranted a traceback. After I gave up for not wielding enough python powers.

PR #591, opened 2020-01-08
packaging: improve deps, polish spec and fix split of files

pirat89:
Originally we had troubles with dependencies and release processes
    as we have been strict about versions of leapp and repositories,
    specified in the leapp rpm. As well, we would like to ensure that
    currently we will install all rpms when leapp tool is installed.
    
    So we are going to keep dependency on leapp-repository, but
    without a version specification. We are going to left the
    reponsiblity for version check on repositories itself.
    
    As well, python?-leapp rpms will provide new leapp-framework
    capability informing about incompatible changes or new features that
    affects leapp repository (like changes in stdlib). This one will
    be more flexible than official versioning and can be updated per
    PR (see comment in the spec file for more details).
    
    As well, I added several new macros in the SPEC, so there is no need
    to update same values on several places anymore. Just update macros
    in the top of the SPEC.

There is missing solution for Py3 framework. See the comments for more details, and [this](https://serverfault.com/questions/411444/rpm-set-required-somepackage-0-5-0-and-somepackage-0-6-0). I commented out the capability for `python3-leapp` now, so we can resolve it even later.

--------
## Tasks:

- [x] update leapp.spec
- [ ] create documentation
- [x] prepare PR in leapp-repository

### DIFF
--- a/packaging/leapp.spec
+++ b/packaging/leapp.spec
@@ -158,7 +158,14 @@ BuildRequires:  python3-pytest-cov
 %endif
 
 Requires: leapp-framework-dependencies = %{framework_dependencies}
-Provides: leapp-framework = %{framework_version}
+
+# FIXME: avoiding problems with dependencies on virtual capabilities, see:
+#    https://serverfault.com/questions/411444/rpm-set-required-somepackage-0-5-0-and-somepackage-0-6-0
+# Currently, we do not use this rpm, so commenting the provides for now, until
+# we come up with reliable solution. E.g. avoid possibility to install both
+# version of frameworks - for Py2 and Py3 in the same time. Or rename the 
+# capability; e.g.: leapp-framework-py3
+# Provides: leapp-framework = %{framework_version}
 
 %description -n python3-%{name}
 Python 3 leapp framework libraries.
--- a/packaging/leapp.spec
+++ b/packaging/leapp.spec
@@ -237,6 +237,7 @@ rm -f %{buildroot}/%{_bindir}/leapp
 %{_bindir}/leapp
 %dir %{_sharedstatedir}/leapp
 %dir %{_localstatedir}/log/leapp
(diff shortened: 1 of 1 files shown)


pirat89:
@oamg/developers - the name of capability can be changed yet. I am waiting with writting of the documentation until we have agreement on the solution.

pirat89:
@oamg/developers Here is prepared tmp repository with several `leapp` and `leapp-repository` builds with various provides & requirements:
```
[copr:copr.fedorainfracloud.org:pstodulk:leapp-deps-test]
name=Copr repo for leapp-deps-test owned by pstodulk
baseurl=https://copr-be.cloud.fedoraproject.org/results/pstodulk/leapp-deps-test/epel-7-$basearch/
type=rpm-md
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://copr-be.cloud.fedoraproject.org/results/pstodulk/leapp-deps-test/pubkey.gpg
repo_gpgcheck=0
enabled=1
enabled_metadata=1
```

https://copr.fedorainfracloud.org/coprs/pstodulk/leapp-deps-test

drehak:
Given that this is a pretty big change in deps, shouldn't this be mentioned in the `%changelog`? Or do we not care about that section?

pirat89:
@drehak: we never care about %changelog in upstream. this will be kept unchanged forever. That's why we have changlog for releases. Downstream takes care of %changelog.

pirat89:
I will probably document that in the following PR as documentation needs more changes anyway and I am kind of sick now. keeping that opened, until @bocekm  finishes review of Beta/HTB, so we willl be able to merge both things together (as in both cases, other prs will have to be rebased).

pirat89:
It's broken travis ci - issues with pylint. That's unrelated to the changes in the PR. As well, leapp bot issue is unrelated to the PR.

PR #592, opened 2020-01-10
leapp answer command for answerfile management

fernflower:
Introduce a handy alternative to manual answerfile update to ease
the adoption of new preupgrade\answer\upgrade strategy.

This pr reflects only one part of the story - non-interactive
answerfile management. Prior to it the only option for
answerfile manipulation was to update it via some editor of choice;
now it is possible to register user choices for specific dialog options
as a single cli command:

leapp answer --section remove_pam_pkcs11_module_check.confirm=True \
             --section may_not_be_present_section.option=False

Some refactoring to existing cli commands will also be
carried out as part of this patch.

### DIFF
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -11,6 +11,7 @@ from datetime import datetime
 from leapp.config import get_config
 from leapp.exceptions import CommandError, LeappError
 from leapp.logger import configure_logger
+from leapp.messaging.answerstore import AnswerStore
 from leapp.repository.scan import find_and_scan_repositories
 from leapp.utils.audit import Execution, get_connection, get_checkpoints
 from leapp.utils.clicmd import command, command_opt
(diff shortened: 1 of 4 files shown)


leapp/cli/upgrade/__init__.py
```diff
@@ -270,6 +268,24 @@ def preupgrade(args):
...
+@command('answer', help='Manage answerfile')
+@command_opt('answerfile', help='Path to an answerfile to update')
+@command_opt('section', action='append', metavar='dialog_sections',
+             help='Record answer for specific section in answerfile')
+def answer(args):
+    """A command to manage answerfile. Updates answerfile with userchoices"""
+    cfg = get_config()
+    if args.section:
+        args.section = list(itertools.chain(*[i.split(',') for i in args.section]))
+    answerfile_path = args.answerfile or cfg.get('report', 'answerfile')
```
fernflower: A "file exists check" won't hurt, will be done in the next commit

fernflower: And no section chosen should result in an error, not a trace. Also will be fixed

vinzenz:
Please provide a context of this PR in the initial message and some reasoning for the existence of this PR in the shape it is in the commit message.
Also I would like to know how you imagine this to be used by a command line user.

Thanks

fernflower:
@vinzenz I'll present a nice description after I squash the commits.
As for the command line usage this PR will allow answerfile update via smth like 
`leapp answer --section remove_pam_pkcs11_module_check.confirm=True --section may_not_be_present_section.option=False`. Now the only option there is for answerfile manipulation is updating it via some editor of choice so this pr can be considered as a handy alternative which can allow easier adoption of new preupgrade\answer\upgrade strategy.

This pr reflects only one part of the story - non-interactive answerfile management. We will need one for sure as once authconfig->authselect prs become non-experimental a few scenarios (like upgrading official rhel7.6 openstack image) will start hitting the "Dialog choice required" and having a fast way to manage that should ease the frustration.

PR #601, opened 2020-01-31
Add --add flag to the leapp answer

fernflower:
Now by invoking leapp answer command with --allow-missing
flag it will be possible to register an option in the
answerfile.
This scenario may be useful for vagrant box setup and
doesn't affect the preupgrade\upgrade in any way.

leapp answer --section nosuchsection.key=value --allow-missing

### DIFF
--- a/leapp/messaging/answerstore.py
+++ b/leapp/messaging/answerstore.py
@@ -24,7 +24,7 @@ class AnswerStore(object):
         # So don't even bother updating this to some more 'pythonic' coding style
         self._storage[scope] = dialog_scope
 
-    def update(self, answer_file):
+    def update(self, answer_file, allow_missing=False):
         """
         Update answerfile with all answers from answerstore that have correspondent sections in the file.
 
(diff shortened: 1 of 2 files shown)


vinzenz:
How about calling it --add what it effectively is where the standard mode of operation is usually update, no?

fernflower:
> How about calling it --add what it effectively is where the standard mode of operation is usually update, no?

You mean no --allow-missing but --add instead?

vinzenz:
> > How about calling it --add what it effectively is where the standard mode of operation is usually update, no?
> 
> You mean no --allow-missing but --add instead?

Yes indeed

leapp/messaging/answerstore.py
```diff
@@ -24,7 +24,7 @@ def answer(self, scope, key, value):
         # So don't even bother updating this to some more 'pythonic' coding style
         self._storage[scope] = dialog_scope
 
-    def update(self, answer_file):
+    def update(self, answer_file, allow_missing=False):
```
fernflower: Somehow keeping allow_missing here makes more sense to me than renaming that to add (tribute to configparser's allow_no_value?).

vinzenz: Yeah I saw that and it's fine, it's just shorter to write --add than allow-missing on the command line if typed manually

PR #610, opened 2020-02-17
Sort text report entries by inhibitor status and risk factor

drehak:
Currently, the plaintext output entries are written in the order they were produced in. This makes it easy for an inhibitor to get lost, when it (at least in my opinion) should be at the very top.
This PR adds a simple `importance` function, which acts as a sort key. After inhibitors, entries with a high risk factor appear next. It can easily be added to the JSON report too, though in that one it makes sense to retain order by timestamp.

### DIFF
--- a/leapp/utils/report.py
+++ b/leapp/utils/report.py
@@ -2,7 +2,7 @@ import hashlib
 import json
 import os
 
-from leapp.reporting import Remediation, create_report_from_error, create_report_from_deprecation
+from leapp.reporting import Remediation, Severity, create_report_from_error, create_report_from_deprecation
 from leapp.utils.audit import get_messages, get_audit_entry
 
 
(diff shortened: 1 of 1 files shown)


leapp/utils/report.py
```diff
@@ -42,10 +42,24 @@ def fetch_upgrade_report_messages(context_id):
     return messages
 
 
+IMPORTANCES = {
```
bocekm: Call it SEVERITY_LEVELS. Importance does not go well with Severity.

drehak: Alright.

drehak:
e2e tests

drehak:
e2e tests

drehak:
e2e tests

drehak:
e2e tests

drehak:
e2e tests

drehak:
e2e tests

bocekm:
e2e tests

bocekm:
Please rebase and resolve the conflicts.

drehak:
@bocekm done

PR #613, opened 2020-02-21
Workflow API support

vinzenz:
This patch introduces initial necessary changes for the Workflow API
support. The Workflow API allows the actor to specify APIs to use and it
will not to have to specify anymore messages to consume or produce.

With this, it is possible to implement pure code API functions for actor
writers, where they do not have to work with messages but directly can
work with data.

This allows us to provide a stable API that is not changed, however we
can change the messages and system below.
Additionally we can use multiple versions of the API where we can allow
to change APIs in newer versions to make them more useful to consumers
of the API, however we could keep the old versions around to keep their
original code working.

### DIFF
--- a/leapp/actors/__init__.py
+++ b/leapp/actors/__init__.py
@@ -9,9 +9,11 @@ from leapp.exceptions import MissingActorAttributeError, WrongAttributeTypeError
     StopActorExecution, WorkflowConfigNotAvailable
 from leapp.models import DialogModel, Model
 from leapp.tags import Tag
+from leapp.utils import get_api_models
 from leapp.utils.i18n import install_translation_for_actor
 from leapp.utils.meta import get_flattened_subclasses
 from leapp.models.error_severity import ErrorSeverity
+from leapp.workflows.api import WorkflowAPI
 
 
 class Actor(object):
@@ -63,6 +65,12 @@ class Actor(object):
     Dialogs that are added to this list allow for persisting answers the user has given in the answer file storage.
     """
 
+    apis = ()
+    """
+    Tuple of :py:class:`leapp.workflow.api.WorkflowAPI` derived classes that implement Workflow APIs that are used by
+    an actor. Any models the apis produce or consume will be considered by the framework as if the actor defined them.
+    """
+
     text_domain = None
     """
     Using text domain allows to override the default gettext text domain, for custom localization support.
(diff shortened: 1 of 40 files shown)


leapp/actors/__init__.py
```diff
@@ -102,6 +110,8 @@ def __init__(self, messaging=None, logger=None, config_model=None, skip_dialogs=
         if config_model:
             self._configuration = next(self.consume(config_model), None)
 
+        type(self).produces = get_api_models(type(self), 'produces')
```
Rezney: Please comment in the code about why this is needed.

tests/data/workflow-api-tests/actors/apiv1test/actor.py
```diff
@@ -0,0 +1,26 @@
...
+
+    name = 'api_v1_test'
+    consumes = ()
+    produces = ()
+    apis = (TestAPI,)
+    tags = (FirstPhaseTag, WorkflowApiTestWorkflowTag)
+
+    def process(self):
+        api = TestAPI()
+        if api.order_change(1, 2, 3) != (1, 2, 3):
```
Rezney: Please document somewhere what this test wants to achieve.

leapp/snactor/__init__.py
```diff
@@ -44,6 +44,9 @@ def _load_commands_from(path):
 def cli(args):
     if args.logger_config and os.path.isfile(args.logger_config):
         os.environ['LEAPP_LOGGER_CONFIG'] = args.logger_config
+    elif find_repository_basedir('.') and os.path.isfile(os.path.join(find_repository_basedir('.'),
+                                                                      '.leapp/logger.conf')):
+        os.environ['LEAPP_LOGGER_CONFIG'] = os.path.join(find_repository_basedir('.'), '.leapp/logger.conf')
```
Rezney: Why this is needed?

leapp/messaging/__init__.py
```diff
@@ -241,7 +242,7 @@ def consume(self, actor, *types):
         """
         types = tuple((getattr(t, '_resolved', t) for t in types))
         messages = list(self._data) + list(self._new_data)
-        lookup = {model.__name__: model for model in type(actor).consumes + self._config_models}
+        lookup = {model.__name__: model for model in get_api_models(type(actor), 'consumes') + self._config_models}
```
Rezney: Please comment about using this helper function.

Rezney:
I would suggest to create a small tutorial in RTD but it does not necessarily need to be part of this PR...

Rezney:
@vinzenz thanks for the additional commentaries.

We do not have any code using this yet but tests should cover the functionality and we can adjust/improve it on the way. I am going to merge it if @pirat89 does not have any objection.

PR #614, opened 2020-03-09
Tupelize non tuples consumes, produces or apis

vinzenz:
Previously non tuples for consumes or produces fields did get
converted to a tuple. However with the introduction of the workflow apis
this behavior got reverted and caused some breakage. This patch adds
this behavior back and extends it even to apis fields.


Rezney:
Tested manually and it passed. @pirat89 shall we merge it?

PR #616, opened 2020-03-10
Add missing leapp logger to logger.conf

bocekm:
- framework is using the 'leapp' logger which was not being configured in the conf file
- it caused certain commands to get stuck on RHEL 7 when leapp package (with the
  conf file) was installed, e.g. `snactor workflow sanity-check ipu`


leapp/logger/__init__.py
```diff
@@ -52,39 +52,65 @@ def _remote_emit(self, log_data):
...
+
             logging.getLogger('urllib3').setLevel(logging.WARN)
-            handler = LeappAuditHandler()
-            handler.setFormatter(logging.Formatter(fmt=log_format, datefmt=log_date_format))
-            logging.getLogger('leapp').addHandler(handler)
+
+            audit_handler = LeappAuditHandler()
+            audit_handler.setFormatter(logging.Formatter(fmt=log_format, datefmt=log_date_format))
+            audit_handler.setLevel(logging.DEBUG)
+            root_logger.addHandler(audit_handler)
```
bocekm: @vinzenz, should the audit be attached to the root logger? The file handler is attached to the 'leapp' logger.

leapp/logger/__init__.py
```diff
@@ -52,39 +52,66 @@ def _remote_emit(self, log_data):
 
 
 def configure_logger(log_file=None):
-    global _logger
-    if not _logger:
+    """
+    Configure loggers as per the description below.
```
ZhukovGreen, line 56: Should be on the same line with """ (https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)

leapp/logger/__init__.py
```diff
@@ -52,39 +52,66 @@ def _remote_emit(self, log_data):
...
+    :return: The 'leapp' logger
+    """
+    global _leapp_logger
+    if not _leapp_logger:
+        _leapp_logger = logging.getLogger('leapp')
+
         log_format = '%(asctime)s.%(msecs)-3d %(levelname)-8s PID: %(process)d %(name)s: %(message)s'
         log_date_format = '%Y-%m-%d %H:%M:%S'
-        path = os.getenv('LEAPP_LOGGER_CONFIG', '/etc/leapp/logger.conf')
+        logging.Formatter.converter = time.gmtime
```
ZhukovGreen, line 79: Please don't do this. This is a global state of all loggers. What about the one from the cookbook? (https://docs.python.org/3/howto/logging-cookbook.html#formatting-times-using-utc-gmt-via-configuration)
this could be solved something similar to:
```diff
diff --git a/etc/leapp/logger.conf b/etc/leapp/logger.conf
index ee1cf91..66102c4 100644
--- a/etc/leapp/logger.conf
+++ b/etc/leapp/logger.conf
@@ -10,7 +10,7 @@ keys=leapp_audit,stream
 [formatter_leapp]
 format=%(asctime)s.%(msecs)-3d %(levelname)-8s PID: %(process)d %(name)s: %(message)s
 datefmt=%Y-%m-%d %H:%M:%S
-class=logging.Formatter
+class=leapp.logger.UTCFormatter
 
 [logger_urllib3]
 level=WARN
```

leapp/logger/__init__.py
```diff
@@ -52,39 +52,66 @@ def _remote_emit(self, log_data):
...
+        handler: FileHandler
+          level: DEBUG
+
+    :return: The 'leapp' logger
+    """
+    global _leapp_logger
+    if not _leapp_logger:
+        _leapp_logger = logging.getLogger('leapp')
+
         log_format = '%(asctime)s.%(msecs)-3d %(levelname)-8s PID: %(process)d %(name)s: %(message)s'
```
ZhukovGreen, line 77: I think fo ruser would be convininent to see that this time is UTC time
```suggestion
        log_format = '%(asctime)s.%(msecs)-3d UTC %(levelname)-8s PID: %(process)d %(name)s: %(message)s'
```

PR #618, opened 2020-03-11
Add tutorial for Workflow APIs

vinzenz:
This patch adds a basic tutorial for Workflow APIs how to use them, how
to define them and some best practises for them.

### DIFF
--- a/docs/source/workflow-apis.md
+++ b/docs/source/workflow-apis.md
@@ -1,0 +1,219 @@
+# Workflow APIs
+
+## Using Workflow APIs
+
+To start using a Workflow API you have to know which API you are going to be using.
+In this case we are assuming that the repository defines a v2 namespace with an API definition of MyExampleAPI.
+
+Now to use it, you have to import the API class object you want to use and specify the type of the class in the `apis`
+field.
+This is necessary so all the messages that are consumed and produced by the API can be picked up by the framework
+and the actor can depend on them.
+
+Here is an example of the usage:
+
+```python
+from leapp.actors import Actor
+from leapp.workflows.api.v2 import MyExampleAPI
+
+class MyWorkflowAPIUsingActor(Actor):
+    """ An example actor consuming the v2.MyExampleAPI Workflow API """
+    produces = ()
+    consumes = ()
+    apis = (MyExampleAPI,)
+
+    def process(self):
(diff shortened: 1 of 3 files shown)


docs/source/workflow-apis.md
```diff
@@ -0,0 +1,170 @@
...
+    apis = (MyExampleAPI,)
+
+    def process(self):
+        api = MyExampleAPI()
+        api.second(value=10)
+```
+
+## General workings of the Workflow API feature
+
+In the leapp repository , you want to create the API, for create a folder called `apis`.
```
pirat89: The sentence has probably typo or it's incomplete. And use please *directory* instead of folder :)

docs/source/workflow-apis.md
```diff
@@ -0,0 +1,170 @@
...
+
+class MyExampleAPI(WorkflowAPI):
+    def first(self):
+        return 'First API function result'
+```
+
+### Working with messages
+
+Now if your API is going to consume or produce any messages, those messages have to be declared the same way as they
+are declared for actors.
```
pirat89: Add a note here (or below, but in the same section) that an actor using an api doesn't have to explicitly name all messages consumed/produced by the API, as leapp framework unifies equivalent tuples. It could be short, with reference on dependencies.

docs/source/workflow-apis.md
```diff
@@ -0,0 +1,170 @@
...
+
+    def first(self):
+        return 'First API function result'
+
+    def second(self, value):
+        # Creates a new message `Produced` for each message of type `Consumed` with the additional value passed by 
+        # the caller.
+        for consumed in api.consume(Consumed):
+            self.produce(Produced(consumed=consumed, value=value))
+```
```
pirat89: Document writting of tests. To make it clear, that tests for apis are possible and recommended as well.

pirat89:
And I am still missing explanation what the workflow API is and what is the purpose of it. I can read it, but from the point of users, it's not so lear why they should be interested about the workflow APIs. As well, it's mentioned only in tutorial but it should be mentioned on more places. If there is no time for that now, create please issue so we will not froget about that.

pirat89:
oh crap. I didn't want to create merge commit.

Rezney:
@vinzenz pls rebase

PR #621, opened 2020-03-26
Show progress in non verbose executions

vinzenz:
Previously when leapp was executed not much was visible during the
upgrade except for the DNF output. This patch now introduces some
changes to the execution that shows what phase starts and which
actor gets executed.

Example output:
```
[root@localhost leapp]# leapp preupgrade
==> Processing phase `configuration_phase`
====> * ipu_workflow_config
        IPU workflow config actor
==> Processing phase `FactsCollection`
====> * scanmemory
        Scan Memory of the machine.
====> * tcp_wrappers_config_read
        Parse tcp_wrappers configuration files /etc/hosts.{allow,deny}.
====> * scan_kernel_cmdline
        No documentation has been provided for the scan_kernel_cmdline actor.
====> * storage_scanner
        Provides data about storage settings.
====> * system_facts
        Provides data about many facts from system.
====> * udevadm_info
        Produces data exported by the "udevadm info" command.
====> * scan_subscription_manager_info
        Scans the current system for subscription manager information
====> * transaction_workarounds
        Provides additional RPM transaction tasks based on bundled RPM packages.
====> * rpm_scanner
        Provides data about installed RPM Packages.
Loaded plugins: product-id, subscription-manager
====> * network_manager_read_config
        Provides data about NetworkManager configuration.
====> * repository_mapping
        Produces message containing repository mapping based on provided file.
====> * sssd_facts
        Check SSSD configuration for changes in RHEL8 and report them in model.
====> * root_scanner
        Scan the system root directory and produce a message containing
====> * pam_modules_scanner
        Scan the pam directory for services and modules used in them
====> * pci_devices_scanner
        Provides data about existing PCI Devices.
```

### DIFF
--- a/leapp/utils/output.py
+++ b/leapp/utils/output.py
@@ -1,13 +1,17 @@
 from __future__ import print_function
 
 import json
+import os
 import sys
 from contextlib import contextmanager
 
 from leapp.exceptions import LeappRuntimeError
 from leapp.models import ErrorModel
-from leapp.reporting import Flags
-from leapp.utils.report import fetch_upgrade_report_messages
+
+
+def _is_verbose():
+    """Redefinition of is_verbose from leapp.libraries.stdlib.config because it leads to import errors"""
+    return os.getenv('LEAPP_VERBOSE', '0') == '1'
 
 
 class Color(object):
@@ -50,6 +54,9 @@ def print_error(error):
 
 
 def report_inhibitors(context_id):
+    # The following imports are required to be here to avoid import loop problems
+    from leapp.reporting import Flags  # pylint: disable=import-outside-toplevel
+    from leapp.utils.report import fetch_upgrade_report_messages  # pylint: disable=import-outside-toplevel
     reports = fetch_upgrade_report_messages(context_id)
     inhibitors = [report for report in reports if Flags.INHIBITOR in report.get('flags', [])]
     if inhibitors:
@@ -114,3 +121,20 @@ def beautify_actor_exception():
             sys.stderr.write(pretty_block_text(msg, color="", width=len(msg)))
     finally:
         pass
+
+
+def display_status_current_phase(phase):
+    if not _is_verbose():
+        print('==> Processing phase `{name}`'.format(name=phase[0].name))
+
+
+def _get_description_title(actor):
+    lines = actor.description.strip().split('\n')
+    return lines[0].strip() if lines else actor.name
+
+
+def display_status_current_actor(actor, designation=''):
+    if not _is_verbose():
+        print('====> * {actor}{designation}\n        {title}'.format(actor=actor.name,
+                                                                     title=_get_description_title(actor),
+                                                                     designation=designation))
--- a/leapp/workflows/__init__.py
+++ b/leapp/workflows/__init__.py
@@ -13,6 +13,7 @@ from leapp.tags import ExperimentalTag
 from leapp.utils import reboot_system
 from leapp.utils.audit import checkpoint, get_errors
 from leapp.utils.meta import with_metaclass, get_flattened_subclasses
+from leapp.utils.output import display_status_current_phase, display_status_current_actor
 from leapp.workflows.phases import Phase
 from leapp.workflows.policies import Policies
 from leapp.workflows.phaseactors import PhaseActors
@@ -289,6 +290,7 @@ class Workflow(with_metaclass(WorkflowMeta)):
                 self.log.info('Skipping phase {name}'.format(name=phase[0].name))
                 continue
 
+            display_status_current_phase(phase)
             self.log.info('Starting phase {name}'.format(name=phase[0].name))
             current_logger = self.log.getChild(phase[0].name)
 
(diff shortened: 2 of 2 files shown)


leapp/utils/output.py
```diff
@@ -1,10 +1,17 @@
...
 import sys
 from contextlib import contextmanager
 
-from leapp.models import ErrorModel
 from leapp.exceptions import LeappRuntimeError
+from leapp.models import ErrorModel
+
+
+def _is_verbose():
+    """Redefinition of is_verbose from leapp.libraries.stdlib.config because it leads to import errors"""
```
Rezney: Could you be please more specific about the import Errors?

I got:

```python
  File "/usr/bin/snactor", line 9, in <module>
    load_entry_point('leapp==0.9.0', 'console_scripts', 'snactor')()
  File "/usr/lib/python2.7/site-packages/leapp/snactor/__init__.py", line 77, in main
    load_commands()
  File "/usr/lib/python2.7/site-packages/leapp/snactor/__init__.py", line 19, in load_commands
    _load_commands_from(commands.__file__)
  File "/usr/lib/python2.7/site-packages/leapp/snactor/__init__.py", line 33, in _load_commands_from
    mod = importer.find_module(name).load_module(name)
  File "/usr/lib64/python2.7/pkgutil.py", line 246, in load_module
    mod = imp.load_module(fullname, self.file, self.filename, self.etc)
  File "/usr/lib/python2.7/site-packages/leapp/snactor/commands/discover.py", line 8, in <module>
    from leapp.repository.scan import find_and_scan_repositories
  File "/usr/lib/python2.7/site-packages/leapp/repository/__init__.py", line 9, in <module>
    from leapp.repository.actor_definition import ActorDefinition
  File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 10, in <module>
    from leapp.actors import get_actor_metadata, get_actors
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 16, in <module>
    from leapp.workflows.api import WorkflowAPI
  File "/usr/lib/python2.7/site-packages/leapp/workflows/__init__.py", line 16, in <module>
    from leapp.utils.output import display_status_current_phase, display_status_current_actor
  File "/usr/lib/python2.7/site-packages/leapp/utils/output.py", line 8, in <module>
    from leapp.libraries.stdlib.config import is_verbose
  File "/usr/lib/python2.7/site-packages/leapp/libraries/stdlib/__init__.py", line 14, in <module>
  File "/usr/lib/python2.7/site-packages/leapp/libraries/stdlib/api.py", line 9, in <module>
    from leapp.actors import Actor
ImportError: cannot import name Actor
```
when running "snactor discover"...

Rezney: @vinzenz any update on this?

vinzenz:
leapp-ci build

leapp/utils/output.py
Rezney: As discussed offline, this was most probably related to the old build. The new one was stuck.

leapp/utils/output.py
```diff
@@ -1,10 +1,17 @@
...
+    return os.getenv('LEAPP_DEBUG', '0') == '1' or os.getenv('LEAPP_VERBOSE', '0') == '1'
```
drehak, line 14: You don't need to check both variables. When `LEAPP_DEBUG` is set to 1, `LEAPP_VERBOSE` gets set too.
```suggestion
    return os.getenv('LEAPP_VERBOSE', '0') == '1'
```

Rezney:
Sometimes, the output seems to be pretty mixed into the other output:

```
  unbound-libs-1.7.3-8.el8.x86_64                                               
  xkeyboard-config-2.24-3.el8.noarch                                            

Complete!
==> Processing phase `TargetTransactionCheck`
====> * report_set_target_release
        Reports that a release will be set in the subscription-manager after the upgrade.
====> * dnf_transaction_check
        This actor tries to solve the RPM transaction to verify the all package dependencies can be successfully resolved.
Last metadata expiration check: 0:03:41 ago on Thu Apr  2 10:23:51 2020.

```

I understand we may not be in control of this, but what about if add something in front of the arrows to distinguish those are really different logs?

Something like:

```
  unbound-libs-1.7.3-8.el8.x86_64                                               
  xkeyboard-config-2.24-3.el8.noarch                                            

Complete!
Leapp ==> Processing phase `TargetTransactionCheck`
Leapp ====> * report_set_target_release
        Reports that a release will be set in the subscription-manager after the upgrade.
Leapp ====> * dnf_transaction_check
        This actor tries to solve the RPM transaction to verify the all package dependencies can be successfully resolved.
Last metadata expiration check: 0:03:41 ago on Thu Apr  2 10:23:51 2020.
```

what do you think @pirat89 @shaded-enmity ?

vinzenz:
Actually that's part of the example output above. See RPM Scanner.

While I can see that you might want that, I feel that to be quite convoluting the output to prefix it with a string like this. 
We rather should ask for feedback from some UI guys if we really want some feedback.
The main thing this PR is handling is that it actually shows progress of leapp itself as well. So people can see that something is happening and what it is actually doing right now.

leapp/utils/output.py
vinzenz: @drehak it's a copy and paste of the original implementation

drehak: I see. Still, debug also sets verbose in cli code now.

Rezney:
@vinzenz ok, fair enough. Looks good to me then.

Rezney:
@drehak what do you think?

drehak:
@Rezney Yeah, looks good.
The displayed actor descriptions differ a lot and gramatically, some of them quite stand out. We could "normalize" those a bit when we have time, but the strings are in in leapp-repository anyway, so not a problem here.

Just address the requested change above and we can merge. :)

drehak:
Though there could also be something like
`==> Loading actors`
at the start, since it seemingly hangs there for a few seconds.

vinzenz:
e2e tests

PR #622, opened 2020-03-27
Add --no-rhsm and LEAPP_NO_RHSM for skipping RHSM

drehak:
The behavior is the same as the legacy `LEAPP_DEVEL_SKIP_RHSM`, except supported.

Requries Repo PR: oamg/leapp-repository/pull/482

### DIFF
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -161,6 +161,8 @@ def prepare_configuration(args):
         os.environ['LEAPP_NO_RHSM'] = '1'
     elif os.getenv('LEAPP_NO_RHSM') != '1':
         os.environ['LEAPP_NO_RHSM'] = os.getenv('LEAPP_DEVEL_SKIP_RHSM', '0')
+    if args.enablerepo:
+        os.environ['LEAPP_ENABLE_REPOS'] = ','.join(args.enablerepo)
     configuration = {
         'debug': os.getenv('LEAPP_DEBUG', '0'),
         'verbose': os.getenv('LEAPP_VERBOSE', '0'),
@@ -181,14 +183,16 @@ def process_whitelist_experimental(repositories, workflow, configuration, logger
             raise CommandError(msg)
 
 
-@command('upgrade', help='Upgrades the current system to the next available major version.')
+@command('upgrade', help='Upgrade the current system to the next available major version.')
 @command_opt('resume', is_flag=True, help='Continue the last execution after it was stopped (e.g. after reboot)')
 @command_opt('reboot', is_flag=True, help='Automatically performs reboot when requested.')
-@command_opt('whitelist-experimental', action='append', metavar='ActorName', help='Enables experimental actors')
+@command_opt('whitelist-experimental', action='append', metavar='ActorName', help='Enable experimental actors')
 @command_opt('debug', is_flag=True, help='Enable debug mode', inherit=False)
 @command_opt('verbose', is_flag=True, help='Enable verbose logging', inherit=False)
 @command_opt('no-rhsm', is_flag=True, help='Use only custom repositories and skip actions'
                                            ' with Red Hat Subscription Manager')
+@command_opt('enablerepo', action='append', metavar='<repoid>',
+             help='Enable specified repository. Can be used multiple times.')
 def upgrade(args):
     skip_phases_until = None
     context = str(uuid.uuid4())
@@ -253,6 +257,8 @@ def upgrade(args):
 @command_opt('verbose', is_flag=True, help='Enable verbose logging', inherit=False)
 @command_opt('no-rhsm', is_flag=True, help='Use only custom repositories and skip actions'
                                            ' with Red Hat Subscription Manager')
+@command_opt('enablerepo', action='append', metavar='<repoid>',
+             help='Enable specified repository. Can be used multiple times.')
 def preupgrade(args):
     context = str(uuid.uuid4())
     cfg = get_config()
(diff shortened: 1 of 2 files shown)


leapp/cli/upgrade/__init__.py
```diff
@@ -156,10 +156,15 @@ def prepare_configuration(args):
     else:
         os.environ['LEAPP_EXPERIMENTAL'] = '0'
     os.environ['LEAPP_UNSUPPORTED'] = '0' if os.getenv('LEAPP_UNSUPPORTED', '0') == '0' else '1'
+    if args.no_rhsm or os.environ['LEAPP_NO_RHSM'] == '1':
```
vinzenz:
```suggestion
    if args.no_rhsm or os.getenv('LEAPP_NO_RHSM', '0') == '1':
```

leapp/cli/upgrade/__init__.py
```diff
@@ -156,10 +156,15 @@ def prepare_configuration(args):
...
+        os.environ['LEAPP_NO_RHSM'] = '1'
+    else:
+        os.environ['LEAPP_NO_RHSM'] = os.environ.get('LEAPP_DEVEL_SKIP_RHSM', '0')
```
vinzenz:
```suggestion
        os.environ['LEAPP_NO_RHSM'] = os.getenv('LEAPP_DEVEL_SKIP_RHSM', '0')
```

leapp/cli/upgrade/__init__.py
```diff
@@ -156,10 +156,15 @@ def prepare_configuration(args):
...
     configuration = {
         'debug': os.getenv('LEAPP_DEBUG', '0'),
         'verbose': os.getenv('LEAPP_VERBOSE', '0'),
         'whitelist_experimental': args.whitelist_experimental or ()
+        'no_rhsm': os.environ['LEAPP_NO_RHSM']
```
vinzenz:
```suggestion
        'no_rhsm': os.getenv('LEAPP_NO_RHSM', '0')
```

leapp/cli/upgrade/__init__.py
drehak: Nice, thanks.

leapp/cli/upgrade/__init__.py
drehak: Nice, thanks.

leapp/cli/upgrade/__init__.py
drehak: In this case we certainly set `LEAPP_NO_RHSM` to some value, so I don't see the point in calling `getenv`.

vinzenz: true

leapp/cli/upgrade/__init__.py
```diff
@@ -182,6 +185,7 @@ def process_whitelist_experimental(repositories, workflow, configuration, logger
 @command_opt('whitelist-experimental', action='append', metavar='ActorName', help='Enables experimental actors')
 @command_opt('debug', is_flag=True, help='Enable debug mode', inherit=False)
 @command_opt('verbose', is_flag=True, help='Enable verbose logging', inherit=False)
+@command_opt('no-rhsm', is_flag=True, help='Skip Red Hat Subscription Manager checks', inherit=False)
```
pirat89: It's something that customers could want to use and skipping checks is not exactly what about this is. What about something like:

```suggestion
@command_opt('no-rhsm', is_flag=True, help='Use only custom repositories and skip actions with Red Hat Subscription Manager', inherit=False)
```

You can modify it. Just take the point.

pirat89:
@drehak probably broken, it seems the LEAPP_NO_RHSM is not set when --no-rhsm is used:

```
# LEAPP_NO_RHSM=1 leapp preupgrade --debug
...
                           ERRORS                           

2020-04-03 10:16:47.522338 [ERROR] Actor: verify_check_results
Message: RHSM is skipped but not custom repositories have been delivered.

                       END OF ERRORS                        

# leapp preupgrade --debug --no-rhsm
...
                           ERRORS                           

2020-04-03 10:12:46.338074 [ERROR] Actor: scan_subscription_manager_info
Message: A subscription-manager command failed to execute
Summary: 
    Details: Command ['subscription-manager', 'release'] failed with exit code 1.
    Stderr: This system is not yet registered. Try 'subscription-manager register --help' for more information.
    Hint: Please ensure you have a valid RHEL subscription and your network is up.

                       END OF ERRORS                        


```

pirat89:
@drehak: also please bump the minor version of `framework_version` capability inside the spec file.

AloisMahdal:
If leapp_upgrade_repositories.repo is empty, we ge this inhibitor: 

    Risk Factor: high (inhibitor)
    Title: Using RHSM has been skipped but no custom repositories have been delivered.
    Summary: Leapp is run in the mode when the Red Hat Subscription Manager is not used (the --no-rhsm option or the LEAPP_NO_RHSM=1 environment variable has been set) so leapp is not able to obtain YUM/DNF repositories with the content for the target system in the standard way. The content has to be delivered by the user manually.
    Remediation: [hint] Create the repository file according to instructions in the referred document on the following path with all repositories that should be used during the upgrade: "/etc/leapp/files/leapp_upgrade_repositories.repo".

The text mentions "the referred document" but does not refer any document.

How is cust going to know how to add the repos?  (Merely `cat`'ing may not work if they already  deployed the destination repos, because of duplicate id's.)

pirat89:
Just waiting for finished testing of leapp-repository PR 482 as this needs to be merged together. However it's preapred for merge btw.

bocekm:
@drehak, @pirat89, we need to add the new option and the env var to the man page.

PR #623, opened 2020-04-08
Fix output location for errors block

vinzenz:
Previously the location of the pretty block for ERRORS has been
accidentally changed from stdout to stderr. This patch reverts this
mistake.


vinzenz:
e2e tests

PR #625, opened 2020-04-20
Fix handling of debug/verbose in {pre,}upgrade

vinzenz:
Previously changes have been introduced that caused the debug/verbose
information not to be stored correctly to the database, which led to
different output behavior after reboot. (No debug/verbose output after
resuming the execution)

This patch changes the order when this data is processed so the
configuration is updated correctly.

JIRA-Taks: OAMG-3087

### DIFF
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -197,13 +197,13 @@ def upgrade(args):
     skip_phases_until = None
     context = str(uuid.uuid4())
     cfg = get_config()
+    handle_output_level(args)
     configuration = prepare_configuration(args)
     answerfile_path = cfg.get('report', 'answerfile')
     userchoices_path = cfg.get('report', 'userchoices')
 
     if os.getuid():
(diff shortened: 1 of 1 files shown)


vinzenz:
e2e tests

AloisMahdal:
Integration failed because we've recently added int.test that would depend on RHSM, but also few months ago we disabled RHSM (to increase reliability, ie avoid stage cdn).  For now I've re-enabled RHSM to keep the coverage.  If stage CDN will start acting out we'll see..

IOW, you can re-run tests of needed.

AloisMahdal:
e2e tests

AloisMahdal:
integration failed due to timeout. will reschedule

AloisMahdal:
e2e tests

drehak:
@AloisMahdal are the tests failing or is this just a CI problem?

pirat89:
@drehak code is ok. it is error of tests or ci (summarizing what we discussed with @AloisMahdal  in the past)

pirat89:
e2e tests

PR #627, opened 2020-05-12
Add --actor-config to snactor run

### LINKED ISSUE #530: Workflow configuration: missing snactor support
Current implementation of the workflow configuration is problematic when used with snactor. See https://github.com/oamg/leapp/pull/505 When we run e.g. `# snactor run <ActorName>` on actor which depends on workflow configuration model it fails. We need to find a way to provide the config to the actor upon issuing the snactor command. Additionally the exception/error message should be modified to reflect the change.

vinzenz:
This patch introduces support for specifying a model which the actor
should consume for configuration data.
Using this feature requires to run an actor before which produces this
model and saves the output.

Jira-Task: OAMG-1092

Fixes #530

### DIFF
--- a/leapp/snactor/commands/run.py
+++ b/leapp/snactor/commands/run.py
@@ -1,15 +1,15 @@
+from importlib import import_module
 import json
 import sys
 
-from leapp.exceptions import LeappError, CommandError
-from leapp.utils.clicmd import command, command_opt, command_arg
-from leapp.utils.repository import requires_repository, find_repository_basedir
+from leapp.exceptions import CommandError, LeappError
 from leapp.logger import configure_logger
 from leapp.messaging.inprocess import InProcessMessaging
-from leapp.utils.output import report_errors, beautify_actor_exception
 from leapp.repository.scan import find_and_scan_repositories
 from leapp.snactor.context import with_snactor_context
-
+from leapp.utils.clicmd import command, command_arg, command_opt
+from leapp.utils.output import beautify_actor_exception, report_errors
+from leapp.utils.repository import find_repository_basedir, requires_repository
 
 _LONG_DESCRIPTION = '''
 Runs the given actor as specified as `actor_name` in a testing environment.
@@ -21,8 +21,9 @@ https://red.ht/leapp-docs
 
 @command('run', help='Execute the given actor', description=_LONG_DESCRIPTION)
 @command_arg('actor-name')
-@command_opt('--save-output', is_flag=True)
-@command_opt('--print-output', is_flag=True)
+@command_opt('--actor-config', help='Name of the workflow config model to use')
+@command_opt('--save-output', is_flag=True, help='Save the produced messages by this actor.')
+@command_opt('--print-output', is_flag=True, help='Print the produced messages by this actor.')
 @requires_repository
 @with_snactor_context
 def cli(args):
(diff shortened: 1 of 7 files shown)


leapp/snactor/commands/run.py
```diff
@@ -19,10 +19,18 @@
 '''
 
 
+def _get_model(name, repo):
+    try:
+        return {model.__name__.lower(): model}[name.lower()]
```
Rezney, line 24:
```
  File "/usr/bin/snactor", line 9, in <module>
    load_entry_point('leapp==0.10.0', 'console_scripts', 'snactor')()
  File "/usr/lib/python2.7/site-packages/leapp/snactor/__init__.py", line 78, in main
  File "/usr/lib/python2.7/site-packages/leapp/utils/clicmd.py", line 103, in execute
    args.func(args)
  File "/usr/lib/python2.7/site-packages/leapp/utils/clicmd.py", line 125, in called
    self.target(args)
  File "/usr/lib/python2.7/site-packages/leapp/utils/repository.py", line 20, in checker
    return f(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/snactor/context.py", line 32, in wrapper
    return f(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/snactor/commands/run.py", line 46, in cli
    config_model = _get_model(args.actor_config, repository) if args.actor_config else None
  File "/usr/lib/python2.7/site-packages/leapp/snactor/commands/run.py", line 24, in _get_model
    return {model.__name__.lower(): model}[name.lower()]
NameError: global name 'model' is not defined
```

Rezney:
@vinzenz is this still WIP?

Rezney:
Tested, seems to be working fine. @pirat89 any potential comments here?

Rezney:
@vinzenz I originally thought the failed tests was due to some outage but:

~~~
=================================== FAILURES ===================================

_____________ [pylint] leapp/snactor/commands/messages/__init__.py _____________

W: 15, 0: Reimport 'requires_repository' (imported line 11) (reimported)

============= 1 failed, 98 passed, 179 deselected in 27.71 seconds =============
~~~

could you please take a look?

drehak:
Is the option only meant to be used for `IPUConfig` or other models too?
What about using the option multiple times with different models? The corresponding `BaseMessaging` attribute, `_config_models` is a tuple, but if I use the option multiple times, only the latter one gets used. (In the case it's not meant to be used multiple times, shouldn't such usage throw an error instead of some of them being silently discarded?)

Rezney:
@drehak It is meant just for "IPUConfig" model as this is used differently then other models. Please note that you still have to run "IPUWorkflowConfig" actor to generate the data, same as to get the messages from other actors, which however do not require any special treatment. HTH.

drehak:
@Rezney if it's for and only for IPUConfig, shouldn't the option just be a flag? Being able to specify any model as an argument is kinda pointless, if the argument is going to be IPUConfig anyway.

Rezney:
@drehak To give a real example. We may (or may not) want to migrate Convert2RHEL to LEAPP. In this case, there would be a new repository, with a new workflow, and a new config model C2RConfig. HTH

drehak:
@Rezney Alright, looks good to me then.

Rezney:
@vinzenz could you please also document this?

PR #632, opened 2020-05-25
Fix/empty answerfile

ZhukovGreen:
With this MR, if running leapp upgrade, the answerfile created and stored (the same like in preupgrade) in order for a possible use in the next upgrade run.

### DIFF
--- a/leapp/cli/upgrade/__init__.py
+++ b/leapp/cli/upgrade/__init__.py
@@ -240,12 +240,14 @@ def upgrade(args):
         workflow.load_answers(answerfile_path, userchoices_path)
         workflow.run(context=context, skip_phases_until=skip_phases_until, skip_dialogs=True)
 
+    logger.info("Answerfile will be created at %s", answerfile_path)
+    workflow.save_answers(answerfile_path, userchoices_path)
     report_errors(workflow.errors)
     report_inhibitors(context)
     generate_report_files(context)
     report_files = get_cfg_files('report', cfg)
     log_files = get_cfg_files('logs', cfg)
-    report_info(report_files, log_files, fail=workflow.failure)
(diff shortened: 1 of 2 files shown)


leapp/cli/upgrade/__init__.py
```diff
@@ -240,12 +240,14 @@ def upgrade(args):
         workflow.load_answers(answerfile_path, userchoices_path)
         workflow.run(context=context, skip_phases_until=skip_phases_until, skip_dialogs=True)
 
+    logger.info("Answerfile will be created at %s", answerfile_path)
+    workflow.save_answers(answerfile_path, userchoices_path)
```
vinzenz, line 244: if you have this here, then it should be also in preupgrade

leapp/cli/upgrade/__init__.py
```diff
@@ -240,12 +240,14 @@ def upgrade(args):
         workflow.load_answers(answerfile_path, userchoices_path)
         workflow.run(context=context, skip_phases_until=skip_phases_until, skip_dialogs=True)
 
+    logger.info("Answerfile will be created at %s", answerfile_path)
```
vinzenz, line 243: well at line 239 you have 'Using answerfile at '.. and here you write will be created kind of weird no?

Better saying 'Storing answerfile to ... instead since you might not 'create' it again, but just save the current state.

leapp/cli/upgrade/__init__.py
ZhukovGreen: @vinzenz it is (https://github.com/oamg/leapp/blob/master/leapp/cli/upgrade/__init__.py#L291-L292)

vinzenz: OK nevermind then, I didn't realize it was there already. I missed the context in the PR view ;-)

So then LGTM

PR #635, opened 2020-06-02
Remove LEAPP_DEVEL_SKIP_RHSM from docs

bocekm:
It has been replaced with a supported `--no-rhsm` option.


pirat89:
Actually it's not correct. There is --no-rhsm option, and `LEAPP_NO_RHSM` envar which people could use as well. So it should be documented as well.

bocekm:
It is missing also the overlay image size variable. And the link to the best practices is incorrect.

PR #636, opened 2020-06-04
Add option skip_actors_discovery in repo.load

ZhukovGreen:
The mechanism behind how we're testing actors assumes direct injection of the actor context with the help of actor.injected_context. This loads the actor context and testing of actor starts. 

However, before the testing session, we also need to load the repo itself. The current implementation of this procedure also starts the process of discovering actors (also injection occurs) and we can't bypass it. This means we're doing it two times. 

Considering that this procedure takes significant time, this MR introduces the possibility to skip discovering actors during loading the repo.

### DIFF
--- a/leapp/repository/__init__.py
+++ b/leapp/repository/__init__.py
@@ -112,7 +112,7 @@ class Repository(object):
 
         self._definitions.setdefault(kind, []).append(item)
 
-    def load(self, resolve=True, stage=None):
+    def load(self, resolve=True, stage=None, skip_actors_discovery=False):
         """
         Loads the repository resources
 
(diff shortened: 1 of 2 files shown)


leapp/repository/__init__.py
```diff
@@ -145,7 +145,7 @@ def load(self, resolve=True, stage=None):
             sys.meta_path.append(LeappLibrariesFinder(module_prefix='leapp.libraries.common', paths=self.libraries))
             sys.meta_path.append(LeappLibrariesFinder(module_prefix='leapp.workflows.api', paths=self.apis))
 
-        if not stage or stage is _LoadStage.ACTORS:
+        if not stage or stage is _LoadStage.ACTORS and not skip_actors_discovery:
```
vinzenz, line 148:
```suggestion
        if not skip_actors_discovery:
             if not stage or stage is _LoadStage.ACTORS:
```

ZhukovGreen: Accepting the suggestion. Could you please explain why do you think it is better?

ZhukovGreen:
@vinzenz I did a mistake by submitting the MR from my local master (no possibility to force push).

The MR with already applied fixed is here https://github.com/oamg/leapp/pull/638
Please accept it.

Closing this as duplicated

PR #633, opened 2020-05-25
Update and reformat manual pages

drehak:
The existing manual pages are out of date and poorly formatted. This PR reformats the pages, adds info about missing commands/options/variables and structures them a bit more.

(I might have missed things, feel free to review.)

PR #639, opened 2020-06-08
Add missing env vars to the docs

bocekm:
Got partially inspired by https://github.com/oamg/leapp/pull/633.

### DIFF
--- a/docs/source/el7toel8/envars.md
+++ b/docs/source/el7toel8/envars.md
@@ -2,11 +2,48 @@
 
 Actors in the el7toel8 repository use environment variables specified below.
 All these envars use the suggested prefixes specified in
-[the best practices document](../best-practices.html#use-the-leapp-and-leapp-devel-prefixes-for-new-envars)
+[the best practices document](../best-practices#use-the-leapp-and-leapp-devel-prefixes-for-new-envars)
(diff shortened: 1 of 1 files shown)


docs/source/el7toel8/envars.md
```diff
@@ -2,12 +2,45 @@
...
 (false) or `1` (true).
 
+## LEAPP_NO_RHSM
+
+Do not use Red Hat Subscription Management for the upgrade. Using it has the
+same effect as using the `--no-rhsm` leapp option.
+
+## LEAPP_OVL_SIZE
+
+TODO
```
bocekm, line 18: @pirat89, please help me here.

pirat89:
```suggestion

In case any partitions use XFS without ftype attribute, we need to create for every such FS file of specific size, to by able to proceed the upgrade (skipping details what we do with such files). By default, the size is 2048 MB for every such file. In some cases the size has to be changed. If you meet such case, leapp usually informs you that the environment variable needs to be specified (e.g. inside the preupgrade report).
```

bocekm: Thanks. I've tweaked it a bit if you don't mind.

PR #640, opened 2020-06-11
Introduction of the @deprecated decorator & integration into the workflow executions

vinzenz:
Introduction of the @deprecated decorator

This patch introduces the @deprecated Decorator allowing it to be used
for Models and functions/methods anywhere in the code.
The code makes use of the python warnings library and extends the usage
around this by allowing to pass the information since when the entity is
deprecated and pass a reason for the deprecation (e.g. replacement by
something else)

Log deprecation warnings to the audit db

Additionally in this patch logging of deprecations to the audit database
is introduced.
It will create a new event type 'deprecation' and logs details about
location and place.

Start reporting deprecation usages

On top of it this patch adds deprecation usages to the reports.
It will receive the items from the audit database.
If an item is deprecated for more than 6 month the severity of
the report entry is set to HIGH otherwise it is set to MEDIUM.

### DIFF
--- a/docs/source/deprecation.md
+++ b/docs/source/deprecation.md
@@ -1,0 +1,136 @@
+## Deprecation
+
+To help with the deprecation process of entities in leapp, the leapp framework introduced a decorator `deprecated`
+which can be imported from `leapp.utils.deprecation`.
+
+This decorator will mark items as being deprecated and ensures that a deprecation warning is displayed if this item
+is being used.
+
+### Warning message display behaviour
+
+Warning messages are always shown after `snactor` executions via `snactor run` or `snactor workflow run`.
+At the end of executions of `leapp upgrade` or `leapp preupgrade` the warnings are added as report entries with a
+developer audience. If the deprecation is less than 6 months old the severity is set to 'MEDIUM' if it is older than
+6 months the severity is set to 'HIGH'
+
+#### snactor warning message example
+
+```
+============================================================
+                 USE OF DEPRECATED ENTITIES
+============================================================
+
+Usage of deprecated function "deprecated_function" @ /Users/vfeenstr/devel/work/leapp/leapp/tests/data/deprecation-tests/actors/deprecationtests/actor.py:19
+Near:         deprecated_function()
+
+Reason: This function is no longer supported.
+------------------------------------------------------------
+Usage of deprecated Model "DeprecatedModel" @ /Users/vfeenstr/devel/work/leapp/leapp/tests/data/deprecation-tests/actors/deprecationtests/actor.py:20
+Near:         self.produce(DeprecatedModel())
+
+Reason: This model is deprecated - Please do not use it anymore
+------------------------------------------------------------
+Usage of deprecated class "DeprecatedNoInit" @ /Users/vfeenstr/devel/work/leapp/leapp/tests/data/deprecation-tests/actors/deprecationtests/actor.py:21
+Near:         DeprecatedNoInit()
+
+Reason: Deprecated class without __init__
+------------------------------------------------------------
+Usage of deprecated class "DeprecatedBaseNoInit" @ /Users/vfeenstr/devel/work/leapp/leapp/tests/data/deprecation-tests/actors/deprecationtests/actor.py:22
+Near:         DeprecatedNoInitDerived()
+
+Reason: Deprecated base class without __init__
+------------------------------------------------------------
+Usage of deprecated class "DeprecatedWithInit" @ /Users/vfeenstr/devel/work/leapp/leapp/tests/data/deprecation-tests/actors/deprecationtests/actor.py:23
+Near:         DeprecatedWithInit()
+
+Reason: Deprecated class with __init__
+------------------------------------------------------------
+
+============================================================
+             END OF USE OF DEPRECATED ENTITIES
+============================================================
+```
+
+#### leapp report example entries
+
+```
+----------------------------------------
+Risk Factor: medium
+Title: Usage of deprecated class "IsolatedActions" at /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/libraries/repofileutils.py:38
+Summary: IsolatedActions are deprecated
+Since: 2020-01-02
+Location: /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/libraries/repofileutils.py:38
+Near: def get_parsed_repofiles(context=mounting.NotIsolatedActions(base_dir='/')):
+
+----------------------------------------
+Risk Factor: medium
+Title: Usage of deprecated class "IsolatedActions" at /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/scansubscriptionmanagerinfo/libraries/scanrhsm.py:8
+Summary: IsolatedActions are deprecated
+Since: 2020-01-02
+Location: /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/scansubscriptionmanagerinfo/libraries/scanrhsm.py:8
+Near:     context = NotIsolatedActions(base_dir='/')
+
+----------------------------------------
+Risk Factor: medium
+Title: Usage of deprecated function "deprecated_method" at /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/deprecationdemo/actor.py:21
+Summary: Deprecated for Demo!
+Since: 2020-06-17
+Location: /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/deprecationdemo/actor.py:21
+Near:         self.deprecated_method()
+
+----------------------------------------
+```
+
+## Usage examples of @deprecated
+
+### Functions
+
+```python
+...
+from leapp.utils.deprecation import deprecated
+
+
+@deprecated(since='2020-06-20', message='This function has been deprecated!')
+def some_deprecated_function(a, b, c):
+    pass
+```
+
+### Models
+
+```python
+...
+from leapp.utils.deprecation import deprecated
+
+
+@deprecated(since='2020-06-20', message='This model has been deprecated!')
+class MyModel(Model):
+    topic = SomeTopic
+    some_field = fields.String()
+```
+
+### Classes
+
+```python
+...
+from leapp.utils.deprecation import deprecated
+
+
+@deprecated(since='2020-06-20', message='This class has been deprecated!')
+class MyClass(object):
+    pass
+
+
+# NOTE: Here we need to offset the stacklevel to get the report of the usage not in the derived class constructor
+# but where the derived class has been created.
+# How many levels you need, you will have to test, it depends on if you use the builtin __init__ methods or not,
+# however it gives you the ability to go up the stack until you reach the position you need to.
+@deprecated(since='2020-06-20', message='This class has been deprecated!', stack_level_offset=1)
+class ABaseClass(object):
+    def __init__(self):
+        pass
+
+
+class ADerivedClass(ABaseClass)
+    def __init__(self):
+        super(ADerivedClass, self).__init__()
+```
--- a/docs/source/tutorials.rst
+++ b/docs/source/tutorials.rst
@@ -16,3 +16,4 @@ Tutorials
     workflow-apis
     unit-testing
     debugging
+    deprecation
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -1,3 +1,4 @@
+import datetime
 import json
 
 from leapp.compat import string_types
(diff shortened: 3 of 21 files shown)


ZhukovGreen:
hey @vinzenz , what are the requirements for this feature?

vinzenz:
@ZhukovGreen it is still WIP sorry forgot to add the label

Rezney:
When I deprecate e.g.

```
@deprecated(since='1902-02-02', message='Ahoj !!!')
class FstabEntry(Model):
    topic = SystemInfoTopic

    fs_spec = fields.String()
    fs_file = fields.String()
    fs_vfstype = fields.String()
    fs_mntops = fields.String()
    fs_freq = fields.String()
    fs_passno = fields.String()
```

I get output as follows:

```
                 USE OF DEPRECATED ENTITIES                 

Usage of deprecated Model "FstabEntry" @ /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:136
Near:                     fs_passno=fs_passno)

Reason: Ahoj !!!
---------
Usage of deprecated Model "FstabEntry" @ /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:136
Near:                     fs_passno=fs_passno)

Reason: Ahoj !!!
---------
Usage of deprecated Model "FstabEntry" @ /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:136
Near:                     fs_passno=fs_passno)

Reason: Ahoj !!!
---------

             END OF USE OF DEPRECATED ENTITIES              

```

I have couple of questions: 

Do we want to have the output duplicated? IIUITC here, the deprecation is rendered 3x because it was called 3x, right? Could we have just a "count" summary here?

The "Near" line looks strange. In this case "fs_passno" is the last field of the model and it does not seem to help much. Can we have here a link to the file and line? The same way as in the previous line?

Rezney:
Trying with:

```
@deprecated(since='1902-02-02', message='Ahoj !!!')
class IsolatedActions(object):
    """ This class allows to perform actions in a manner as if the given base_dir would be the current root """

    _isolated = True
```

it shows on stderr:

```
/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/libraries/mounting.py:257: _DeprecationWarningContext: Usage of deprecated class "IsolatedActions"
  super(NotIsolatedActions, self).__init__(base_dir=base_dir, implementation=IsolationType.NONE)
/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/libraries/mounting.py:257: _DeprecationWarningContext: Usage of deprecated class "IsolatedActions"
  super(NotIsolatedActions, self).__init__(base_dir=base_dir, implementation=IsolationType.NONE)
/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/libraries/mounting.py:257: _DeprecationWarningContext: Usage of deprecated class "IsolatedActions"
  super(NotIsolatedActions, self).__init__(base_dir=base_dir, implementation=IsolationType.NONE)
```

but the final summary does not contain it:

```
===========================================================
                 USE OF DEPRECATED ENTITIES                 

Usage of deprecated Model "FstabEntry" @ /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:136
Near:                     fs_passno=fs_passno)

Reason: Ahoj !!!
---------
Usage of deprecated Model "FstabEntry" @ /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:136
Near:                     fs_passno=fs_passno)

Reason: Ahoj !!!
---------
Usage of deprecated Model "FstabEntry" @ /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:136
Near:                     fs_passno=fs_passno)

Reason: Ahoj !!!
---------

             END OF USE OF DEPRECATED ENTITIES              
```

vinzenz:
> Trying with:
> 
> ```
> @deprecated(since='1902-02-02', message='Ahoj !!!')
> class IsolatedActions(object):
>     """ This class allows to perform actions in a manner as if the given base_dir would be the current root """
> 
>     _isolated = True
> ```
> 
> it shows on stderr:
> 
> ```
> /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/libraries/mounting.py:257: _DeprecationWarningContext: Usage of deprecated class "IsolatedActions"
>   super(NotIsolatedActions, self).__init__(base_dir=base_dir, implementation=IsolationType.NONE)
> /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/libraries/mounting.py:257: _DeprecationWarningContext: Usage of deprecated class "IsolatedActions"
>   super(NotIsolatedActions, self).__init__(base_dir=base_dir, implementation=IsolationType.NONE)
> /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/libraries/mounting.py:257: _DeprecationWarningContext: Usage of deprecated class "IsolatedActions"
>   super(NotIsolatedActions, self).__init__(base_dir=base_dir, implementation=IsolationType.NONE)
> ```
> 
> but the final summary does not contain it:
> 
> ```
> ===========================================================
>                  USE OF DEPRECATED ENTITIES                 
> ============================================================
> 
> Usage of deprecated Model "FstabEntry" @ /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:136
> Near:                     fs_passno=fs_passno)
> 
> Reason: Ahoj !!!
> ---------
> Usage of deprecated Model "FstabEntry" @ /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:136
> Near:                     fs_passno=fs_passno)
> 
> Reason: Ahoj !!!
> ---------
> Usage of deprecated Model "FstabEntry" @ /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:136
> Near:                     fs_passno=fs_passno)
> 
> Reason: Ahoj !!!
> ---------
> 
> ============================================================
>              END OF USE OF DEPRECATED ENTITIES              
> ============================================================
> ```

The output has been now deduplicated however the items that are logged to the console will stay, to ensure that the usage is done in a way that the deprecation warnings are only triggered during a proper usage.

The console warnings in this particular example are triggered through the fact that there is a library that has this line:
```python
def get_parsed_repofiles(context=mounting.NotIsolatedActions(base_dir='/')):
```
Which will cause evaluation on library load time. Libraries are loaded outside of actors and this code gets therefore evaluated outside of actors and those aren't loaded. In general I would prefer if this stays like this and we can remove workaround this by doing it like this:
```python
def get_parsed_repofiles(context=None):
    if not context:
        context = mounting.NotIsolatedActions(base_dir='/')
```

vinzenz:
Code is ready for review, only docs need update

vinzenz:
One more note:

Particular point of improvement:
* Usage of deprecated Models in consumes/produces of actors or APIs could be warned about too
* Maybe some additional checks to avoid deprecated code usage within new PRs (e.g. have during linting & unit testing, deprecation usage checks)

vinzenz:
> Seems to be a little inconsistent now. "Line" vs. "Near".
> 
> snactor output:
> 
> ```
> ============================================================
>                  USE OF DEPRECATED ENTITIES                 
> ============================================================
> 
> Usage of deprecated Model "PartitionEntry" @ /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:75
> Near:                     name=name)
> 
> Reason: Ahoj !
> ------------------------------------------------------------
> 
> ============================================================
>              END OF USE OF DEPRECATED ENTITIES              
> ============================================================
> ```
> 
> Report:
> 
> ```
> ----------------------------------------
> Risk Factor: high
> Title: Usage of deprecated Model "PartitionEntry" at /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:75
> Summary: Ahoj !
> Since: 2012-03-03
> Location: /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py:75
> Line:                     name=name)
> 
> ----------------------------------------
> Risk Factor: high
> Title: Usage of deprecated Model "PartitionEntry" at /usr/lib/python2.7/site-packages/leapp/models/fields/__init__.py:465
> Summary: Ahoj !
> Since: 2012-03-03
> Location: /usr/lib/python2.7/site-packages/leapp/models/fields/__init__.py:465
> Line:         return self._model_type(init_method='to_model', **value)
> 
> ----------------------------------------
> ```

Near is now used in both cases

vinzenz:
@ZhukovGreen I have added some tests now

docs/source/deprecation.md
```diff
@@ -0,0 +1,136 @@
...
+@deprecated(since='2020-06-20', message='This class has been deprecated!', stack_level_offset=1)
+class ABaseClass(object):
+    def __init__(self):
+        pass
+
+
+class ADerivedClass(ABaseClass)
+    def __init__(self):
+        super(ADerivedClass, self).__init__()
+```
```
pirat89, line 136: Regarding the documentation, do we plan additional one more specific document for depracation in leapp-repository or this is planned to be the only document? If this is the only document, we should be more specific about the deprecation. E.g. what will happen with deprecated models? Does it mean that during the deprecation time it will be still possible to use the deprecated model? will be actors producing & consuming messages of such models during the 6m period or user will have to immadiately change the code?

docs/source/deprecation.md
```diff
@@ -0,0 +1,136 @@
...
+
+#### leapp report example entries
+
+```
+----------------------------------------
+Risk Factor: medium
+Title: Usage of deprecated class "IsolatedActions" at /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/libraries/repofileutils.py:38
+Summary: IsolatedActions are deprecated
+Since: 2020-01-02
```
pirat89, line 61: @vinzenz WDYT about print of expected "deadline"? It's known that deprecation process is 6month, so I would put the info that it is expected that functionality will be dropped completely after the period. Something like:
```
Risk Factor: medium
Title: Usage of deprecated class "IsolatedActions" at /usr/share/leapp-repository/repositories/system_upgrade/el7toel8/libraries/repofileutils.py:38
Summary: IsolatedActions are deprecated
Since: 2020-01-02
EOL: 2020-07-02
...
```

Rezney:
@pirat89 @ZhukovGreen @vinzenz are there still any blockers in this PR or additional changes / improvements can be done as a follow-up?

ZhukovGreen:
I didn’t complete the review yet and I won’t be able to do that until Wed, so adding reviewed status and will provide my review  later.

pirat89:
> @pirat89 @ZhukovGreen @vinzenz are there still any blockers in this PR or additional changes / improvements can be done as a follow-up?

Yes. I realized I have one blocker. SPEC file is not updated to cover new functionality. So I cannot set proper deps.

docs/source/deprecation.md
vinzenz: I find these questions quite strange to be honest. 

"What will happen with deprecated models" => What is deprecated is supposed to be still usable until it is removed.
Deprecation is a heads up, that something will stop being there and supported in future. Otherwise the fact that you deprecate it is useless. 
If this is what you want answered, the I suggest a follow up task with extending this information.

docs/source/deprecation.md
vinzenz: After 6 months it's changing the risk factor to high, that's all I know. I don't know about calling it EOL though.

pirat89: The point is, that after 6months, the decorator and functionality (model/method/...) will be removed the most probably.

docs/source/deprecation.md
pirat89: The point is, that in such case, it could be useful if we could "drop" the report about deprecated functionality use. I mean, when we need to keep the original functionality still working, the deprecation reports will be print even when customer doesn't care.

pirat89: Resolved by the `@supress_deprecation` decorator.

ZhukovGreen:
I finished the review and my report is here - basically one statement - no tests. I am not sure why this happens, but when I tested the coverage, then, I don't see the main module deprecation.py is covered at all. I also was thinking that there is a bug in the coverage pkg, and put assert 0 inside the module - no bug, tests passed what means, the feature is not covered by tests.

Follow up ticket https://github.com/oamg/leapp/issues/643

PR #648, opened 2020-06-26
Allow running as a not a root

### LINKED ISSUE #646: Impossible to run tests as non-root user
**Actual behavior**
Currently it is not possible to run tests as non-root user, nor to install deps / prep env. as non-root user.
In case I want to run unit-tests locally:
- installation of deps via `make test` is impossible (pip installation on the host system is not acceptable)
- installation of deps via `make install-deps-container` is impossible (docker requires to be run as root)

**Expected behavior**
- ability to install deps into the pyenv environment
- use of `podman` for containers
  - in this case, it could be done via envar; by default it could be still docker, but it should be possible to excute podman instead; e.g.: `CONTAINER_UTIL=podman make install-deps-container; CONTAINER_UTIL=podman make container-test;` ....

**System information (please complete the following information):**
- Fedora 31

**Additional info:**
To make it clear, the possibility to run everything using `podman` could be enough. It's not required to transform everything, but there should be at least one possible way to run tests without needs of root powers.

ZhukovGreen:
Now if you're not a root, then the venv
automatically installed with requred dependencies

Also requirements-tests.txt updated to install leapp
in editable mode

fixes #646 

Now if you're not a root, it is enough to do:
`make install-test && make test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,16 +116,33 @@ install-container-test:
 	docker build -t leapp-tests -f res/docker-tests/Dockerfile.$(subst :,,${CONTAINER}) res/docker-tests
 
 install-test:
+ifeq ($(shell id -u), 0)
 	pip install -r requirements-tests.txt
+else
+	virtualenv --python /usr/bin/python tut
+	. tut/bin/activate ; \
+	pip install -r requirements-tests.txt
+endif
 
 container-test:
 	docker run --rm -ti -v ${PWD}:/payload leapp-tests
 
 test:   lint
+ifeq ($(shell id -u), 0)
 	pytest -vv --cov-report term-missing --cov=leapp tests/scripts
+else
+	. tut/bin/activate ; \
+	pytest -vv --cov-report term-missing --cov=leapp tests/scripts
+endif
 
 lint:
-	pytest --cache-clear --pylint -m pylint leapp tests/scripts/*.py
+ifeq ($(shell id -u), 0)
+	pytest --cache-clear --pylint -m pylint leapp tests/scripts/*.py; \
 	pytest --cache-clear --flake8 -m flake8 leapp tests/scripts/*.py
+else
+	. tut/bin/activate ; \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -116,16 +116,33 @@ install-container-test:
 	docker build -t leapp-tests -f res/docker-tests/Dockerfile.$(subst :,,${CONTAINER}) res/docker-tests
 
 install-test:
+ifeq ($(shell id -u), 0)
```
ZhukovGreen, line 119: I tried to use this `` instead of shell, but that doesn't work

pirat89: You should really use backtick in makefiles instead of this.

ZhukovGreen:
@pirat89 I tested on centos7 and rhel7 with all the following options, being a root:
```Makefile
artem:
ifeq ($(shell id -u), 0)
#ifeq ($(`id -u`), 0)
#ifeq (`id -u`, 0)
	echo "root"
else
	echo "not root"
endif
```
The only option which works is the one which is not commented. Please advice

pirat89:
> @pirat89 I tested on centos7 and rhel7 with all the following options, being a root:
> 
> ```makefile
> artem:
> ifeq ($(shell id -u), 0)
> #ifeq ($(`id -u`), 0)
> #ifeq (`id -u`, 0)
> 	echo "root"
> else
> 	echo "not root"
> endif
> ```
> 
> The only option which works is the one which is not commented. Please advice

I searched a bit more and it's really the correct one. Even suggested. I am curious what exactly happens in this directive.

ZhukovGreen:
Do you want me to create a ticket to fix this throughout leapp and leapp repo? What to do with this MR? @pirat89

pirat89:
> Do you want me to create a ticket to fix this throughout leapp and leapp repo? What to do with this MR? @pirat89

I was just busy. Review and testing finished :)

PR #650, opened 2020-07-14
Add new tags

Rezney:


bocekm:
Aren't messages regarding RHUI always going to relate to public clouds and vice versa? Wouldn't just one tag suffice?

Rezney:
In my mind, these two together can reach wider audience which to me sounds like purpose of the tags.

pirat89:
> This would be nice to have at least a description inside the MR:(

Yep. This was not prepared for merge as commit msg nor PR description properly filled.

PR #651, opened 2020-07-17
dialogs: Create audit entries for dialog answers

vinzenz:
Previously dialog answers have not been tracked in the audit. This patch
introduces a new event type dialog-answer that will be emitted every
time a dialog answer is requested or received. This will report any
answers that are provided through answerfile or interactive reply.

### DIFF
--- a/tests/scripts/test_answerstore.py
+++ b/tests/scripts/test_answerstore.py
@@ -1,0 +1,314 @@
+import pytest
+
+import leapp.messaging.answerstore
+from leapp.messaging.answerstore import AnswerStore, multiprocessing
+
+
+class MockManager(object):
+    def __call__(self):
+        return self
+
+    @staticmethod
+    def dict(*args):
+        return dict(*args)
+
+
+class MockComponentKey1(object):
+    value_type = str
+    key = 'key1'
+    label = 'label key1'
+    description = 'description key1'
+    default = False
+    value = None
+
+
+class MockComponentKey2(object):
+    value_type = str
+    key = 'key2'
+    label = 'label key2'
+    description = 'description key2'
+    default = 'Default'
+    value = None
+
+
+class MockComponentKey1Bool(object):
+    value_type = bool
+    key = 'key1'
+    label = 'label bool key1'
+    description = 'description bool key1'
+    default = True
+    value = None
+
+
+class MockComponentKey2Bool(object):
+    value_type = bool
+    key = 'key2'
+    label = 'label bool key2'
+    description = 'description bool key2'
+    default = False
+    value = None
+
+
+class MockDialogScope1(object):
+    scope = 'scope1'
+    title = 'Mock Dialog scope1'
+    reason = 'Mock Test Data for scope1'
+    components = (MockComponentKey1, MockComponentKey2)
+
+
+class MockDialogScope2(object):
+    scope = 'scope2'
+    title = 'Mock Dialog scope2'
+    reason = 'Mock Test Data for scope2'
+    components = (MockComponentKey1, MockComponentKey2)
+
+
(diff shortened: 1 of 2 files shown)


vinzenz:
e2e tests

AloisMahdal:
For testing; how do i get the entry out of the db?

A totally wild guess

    SELECT data FROM audit
    WHERE event = 'dialog-answer-result';

?

ZhukovGreen:
Please add a unit tests

pirat89:
@vinzenz pls add unit tests. I will test it later this day manually but code looks pretty clear.

vinzenz:
> For testing; how do i get the entry out of the db?
> 
> A totally wild guess
> 
> ```
> SELECT data FROM audit
> WHERE event = 'dialog-answer-result';
> ```
> 
> ?
```sql
SELECT data FROM audit
WHERE event = 'dialog-answer';
```

AloisMahdal:
+1 for making it work with snactor, if possible, right away. we don't have test api to do it otherwise now.

pirat89:
> +1 for making it work with snactor, if possible, right away. we don't have test api to do it otherwise now.

That's probably off-topic in this PR. As it could require more stuff to do in snactor. This PR is just about adding the answers into the leapp.db. Test is simple in such case - just run leapp and see whether you have correct input in the database.

tests/scripts/test_answerstore.py
```diff
@@ -0,0 +1,314 @@
...
+    a.translate(MockDialogBoolScope)
+    assert store['boolscope']['key1'] is True
+    assert store['boolscope']['key2'] is False
+    assert store['scope1']['key1'] == 'scope1.key1'
+    assert store['scope1']['key2'] == 'scope1.key2'
+    assert store['scope2']['key1'] == 'scope2.key1'
+    assert store['scope2']['key2'] == 'scope2.key2'
+
+
+def test_answerfile_generate(tmpdir, answerfile):
```
ZhukovGreen, line 287: some conditions of the tested function is not tested (i.e.  if hasattr(component, 'choices'):, if answer:)

AloisMahdal:
@pirat89 "off-topic in this PR. As it could require more stuff to do in snactor" is a non-sequitur.

The test you are suggesting is what we'll fall back if there's no other option but that's not a reliable one as it relies on state of leapp-repository, and also it's not extensible.

pirat89:
> @pirat89 "off-topic in this PR. As it could require more stuff to do in snactor" is a non-sequitur.
> 
> The test you are suggesting is what we'll fall back if there's no other option but that's not a reliable one as it relies on state of leapp-repository, and also it's not extensible.

Hmm.... Are you aware that you need to write tests using leapp anyway? Becuase for leapp and snactor there is kind of differend codebase.. Including fact that some parts related to answers is exclusive only for leapp?.... The crucial thing for us isn't fact whether answers are loaded and stored using snactor. That's nice-to-have for developers. But crucial is whether the stuff is there when leapp is executed. Because the main reason for these answers in the db is for debugging purposes.

So from that point, you doesn't have to rely on leapp-repository. You can deliver to leapp everything you have to on your own.

AloisMahdal:
As I said, we don't have test api to do that, though. :/

We'll fall back to verification via leapp-repository now.

kokesak:
Verified: https://liver3.brq.redhat.com/jenkins/job/oamg-run-beaker/7657//console
Build:  
 leapp-0.10.0-0.202008051159Z.1110644.log_answers.PR651.el7_6
           leapp-repository-0.10.0-100.202008071306Z.286f83f.master.el7_6

bocekm:
Artem, please create a separate task to improve the tests per your suggestions. Thanks.

PR #652, opened 2020-07-23
Allow caplog to see log records

ZhukovGreen:
Actors are executed in a different processes what
makes caplog blind to emitted messages.

Solution assumes adding a QueueHandler to the root logger
which will gather all the logs to the queue. Then we reemitt
log records from the queue in the main thread. This allows
caplog to see it

This MR is blocker for 
https://github.com/oamg/leapp-repository/pull/467

without this MR we can't capture the logs without mocking it. Otherwise, we can capture the logs by using the caplog fixture provided by pytest, so the test might look like (https://github.com/oamg/leapp-repository/pull/467/files#diff-4385c6a4326fc7ba8f277a9045f0dee9R36-R41)

### DIFF
--- a/leapp/snactor/fixture.py
+++ b/leapp/snactor/fixture.py
@@ -1,8 +1,10 @@
+import logging
 import os
 import socket
 import sys
 import types
 import uuid
+from logging import handlers
 from multiprocessing import Process, Queue
 
 import pytest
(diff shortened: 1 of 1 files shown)


ZhukovGreen:
@vinzenz no idea how to test it:( As well as current_actor_context. Could you advise?

ZhukovGreen:
I found one issue, which needs to be fixed first

ZhukovGreen:
There is a problem when running the tests in the leapp-repository with this feature. Some tests are freezed from time to time. So far unable to resolve it.

ZhukovGreen:
ref https://github.com/oamg/leapp/pull/653#discussion_r467841028

PR #653, opened 2020-08-04
Improve deprecation doc

pirat89:
The deprecation process is not clear as people do not understand what to do when they want to deprecate a leapp entity. To help people understand the deprecation process, let's add more text to make it easier to accept the new process..

Note: it's still not complete. it's just draft. I will mark it for review when I finish it. Do not put here any comments here until it's finished. I opened the draft PR just to make it easier to reference it.

### DIFF
--- a/docs/source/deprecation.md
+++ b/docs/source/deprecation.md
@@ -1,19 +1,501 @@
-## Deprecation
+# Deprecation
 
-To help with the deprecation process of entities in leapp, the leapp framework introduced a decorator `deprecated`
-which can be imported from `leapp.utils.deprecation`.
+The deprecation process is here to make (your) life of developers easier.
+It's not possible to write perfect solution for everything and as the project
+is evolving, it happens that some functionality needs to be changed, replaced
+or dropped completely. Such situations are inevitable. To reduce negative
+impact on your code, we introduce the deprecation process described below.
 
-This decorator will mark items as being deprecated and ensures that a deprecation warning is displayed if this item
-is being used.
+## What is covered by deprecation process in leapp?
 
-### Warning message display behaviour
+In short, leapp entities that are supposed to be used by other developers.
+That means e.g.:
 
-Warning messages are always shown after `snactor` executions via `snactor run` or `snactor workflow run`.
-At the end of executions of `leapp upgrade` or `leapp preupgrade` the warnings are added as report entries with a
-developer audience. If the deprecation is less than 6 months old the severity is set to 'MEDIUM' if it is older than
-6 months the severity is set to 'HIGH'
+1. Models
+1. Shared library classes and functions in leapp repository
+1. Public APIs
+1. Actors providing functionality that could be used by any developer (produce
+   or consume messages)
 
-#### snactor warning message example
+In other words, private classes, private functions or anything in private
+libraries, may be modified or removed without the deprecation process. As well,
+it's possible we will need to change something (e.g. a behaviour of a function)
+that will not be possible to cover reasonably by the deprecation process (e.g.
+change output of the function...). We'll try our best to prevent it, but it may
+happen. To limit such problems, we recommend people to use APIs as much
+as possible.
+
+## What does it mean that something is deprecated?
+
+When you deprecate something, the only thing that changes is that the
+deprecated entity is marked in the code as deprecated which can have
+additional impact, like messages produced on the user's terminal,
+in the report, ... But the rest of the functionality is the same as before,
+until the entity is removed completely.
+
+## What is the deprecation process for leapp?
+
+
+In case a leapp entity covered by the deprecation process is to be removed for
+any reason, it needs to be marked as deprecated before the removal (if
+possible). The deprecation will be applied only for leapp entities that have
+been introduced in an official release in RHEL (IOW, a functionality
+that has been merged into the upstream, but has been removed before the release
+or was marked as experimental all the time, is going to be removed without
+the deprecation state). The time period during which the deprecated entity
+won't be removed is at least 6 months. That doesn't mean we will remove
+everything deprecated immediately after the 6 months, but it's to be expected
+that it will be dropped anytime between 6 and 12 months since the deprecation.
+
+In case of issues, deprecated entities are not going to be fixed since
+they are deprecated (unless they are fixed e.g. as a side-effect of another
+problem fix).
+
+## How do I find out what is deprecated?
+
+Mainly via release notes and changelogs. In the official leapp related projects
+(especially leapp and leapp-repository) the OAMG team takes care of release
+notes to ensure they inform about the dropped and deprecated functionality.
+
+Additionally, when using leapp or snactor, user is notified via messages about
+deprecated entities in **limited cases** (see below). In case of the leapp
+utility, such messages are presented inside the generated reports. In case
+of the snactor utility, the information message is printed in the console
+output at the end of the snactor execution. See examples in this page for
+detail.
+
+Please note, that the Deprecation warning is emitted only if:
+- the deprecated class is instantiated
+- the deprecated function is called
+
+
+## How to deprecate entities in leapp?
+
+When you want to deprecate an entity in leapp projects, use the `deprecated`
+decorator from `leapp.utils.deprecation` above the definition of the entity.
+The decorator has three input parameters:
+
+- `since` (mandatory) - specifying the start of the deprecation protection period
+- `message` (mandatory) - explaining that particular deprecation
+  (e.g. in case the deprecated functionality has a replacement, it is expected
+  it will be mentioned in the msg.)
+- `stack_level_offset` (optional) - useful to adjust the position of the reported
+  usage in the deprecation message; e.g. in case of a base class or derived classes
+
+**Warning: possible change:** *It's possible the `stack_level_offset` parameter
+will be removed (or ignored) in future, if we discover a way to improve
+the deprecation of derived classes.*
+
+In case of a class deprecation, all derived classes are considered to be deprecated
+as well. However, the current reporting could be a little bit confusing. To
+improve that, the `stack_level_offset` option can be specified.
+See [examples of the use of the @deprecated decorator for classes](deprecation.html#classes).
+
+When you mark any entity as deprecated and this entity is then used
+in the code, users will be notified about that via a terminal and report
+messages (see the previous section). However, as the author of the deprecation,
+you know that the entity is deprecated and you do not want
+to notify people about the code that still uses the deprecated entity
+just for the reason to retain the original functionality. To suppress
+the deprecation messages in such cases, use the `suppress_deprecation`
+decorator taking as arguments objects that should not be reported
+by the deprecation. E.g. in case you use it above the definition
+of an actor, any use of the deprecated entity inside the actor
+will not be reported.
+
+WARNING: It is strictly forbidden to use the `suppress_deprecation` decorator
+for any purposes but one - retaining the original official functionality over
+the deprecation protection period. If you see the message and you are not
+the *provider* of the functionality, you have to update your code
+to be independent on it.
+
+### Examples of a model deprecation
+
+Imagine we want to deprecate a *Foo* model that is produced in an actor called
+*FooProducer* and consumed in an actor called *FooConsumer*. Let's keep this
+example simple and say that we do not want to set any replacement of this
+model. The first thing we have to do is to set the model definition as deprecated:
+
+```python
+from leapp.models import Model, fields
+from leapp.topics import SomeTopic
+from leapp.utils.deprecation import deprecated
+
+
+@deprecated(since='2020-06-20', message='This model has been deprecated.')
+class Foo(Model):
+    topic = SomeTopic
+    value = fields.String()
+```
+
+If we do only this and execute actors that produce/consume messages of this model,
+we will obtain messages like these (just example from the actor
+producing the message after execution by snactor):
+
+```bash
+# snactor run fooproducer
+============================================================
+                 USE OF DEPRECATED ENTITIES
+============================================================
+
+Usage of deprecated Model "Foo" @ /path/to/repo/actors/fooproducer/actor.py:17
+Near:         self.produce(Foo(value='Answer is: 42'))
+
+Reason: This model has been deprecated.
+------------------------------------------------------------
+
+============================================================
+             END OF USE OF DEPRECATED ENTITIES
+============================================================
+```
+
+Apparently, the `Reason` is not so good. It's just example. In real world
+example, you would like to provide usually a little bit better explanation.
+Anyway, much more interesting is the point, that the message is now printed
+everytime the actor is executed.
+
+Obviously we do not want to remove the actor yet, because in such a case, the
+model could be hardly called as deprecated - we need to keep the same
+functionality during the protection period. But at the same time, we do not
+want the deprecation message to be produced in this case, as it would be kind
+of a spam for users who don't care about that model at all.
+
+The warning messages are focused on a "custom" use of the deprecated
+model - i.e. when a developer creates their own actor producing/consuming a message
+of the model. To fix this, suppress the deprecation message in this actor.
+
+To do it, the only thing that has to be done is to set the
+`suppress_deprecation` decorator with the `Foo` as an argument (in this case)
+before the actor, e.g.:
+
+```python
+from leapp.actors import Actor
+from leapp.models import Foo  # deprecated model
+from leapp.tags import IPUWorkflowTag, FactsPhaseTag
+from leapp.utils.deprecation import suppress_deprecation
+
+
+@suppress_deprecation(Foo)
+class FooProducer(Actor):
+    """
+    Just produce the right answer to the world.
+    """
+
+    name = 'foo_producer'
+    consumes = ()
+    produces = (Foo,)
+    tags = (IPUWorkflowTag, FactsPhaseTag)
+
+    def process(self):
+        self.produce(Foo(value='Answer is: 42'))
+```
+
+This is the most simple case. Let's do a small change and produce the message
+inside the private actor's library instead. The library looks like this:
+
+```python
+from leapp.models import Foo  # deprecated model
+from leapp.libraries.stdlib import api
+
+def produce_answer():
+    api.produce(Foo(value='Answer is: 42'))
+```
+
+And the updated actor looks like this:
+
+```python
+from leapp.actors import Actor
+from leapp.libraries.actor import fooproducer_lib
+from leapp.models import Foo  # deprecated model
+from leapp.tags import IPUWorkflowTag, FactsPhaseTag
+from leapp.utils.deprecation import suppress_deprecation
+
+
+@suppress_deprecation(Foo)
+class FooProducer(Actor):
+    """
+    Just produce the right answer to the world.
+    """
+
+    name = 'foo_producer'
+    consumes = ()
+    produces = (Foo,)
+    tags = (IPUWorkflowTag, FactsPhaseTag)
+
+    def process(self):
+        fooproducer_lib.produce_answer()
+```
+
+Now, if you execute the actor again you still won't get any
+deprecation message. So the `suppress_deprecation` decorator works transitively
+as expected. However, even when the actor is treated well, the current
+implementation could affect the result of unit tests. To explain the idea of what
+could be wrong, imagine a unit test like this one:
+
+```python
+from leapp.libraries.actor import fooproducer_lib
+from leapp.libraries.common.testutils import produce_mocked
+from leapp.libraries.stdlib import api
+from leapp.models import Foo  # deprecated model
+
+def test_process(monkeypatch):
+    produced_msgs = produce_mocked()
+    monkeypatch.setattr(api, 'produce', produced_msgs)
+    fooproducer_lib.produce_answer()
+    assert Foo(value='Answer is: 42') in produced_msgs.model_instance
+```
+
+If you run the test, you will get output like this (shortened):
+
+```python
+| 21:48:01 | conftest | INFO | conftest.py | Actor 'foo_producer' context teardown complete
+
+repos/system_upgrade/el7toel8/actors/fooproducer/tests/test_unit_fooproducer.py::test_process PASSED
+
+================================================== warnings summary ==================================================
+repos/system_upgrade/el7toel8/actors/fooproducer/tests/test_unit_fooproducer.py::test_process
+  /tmp/leapp-repository/repos/system_upgrade/el7toel8/actors/fooproducer/libraries/fooproducer_lib.py:5: _DeprecationWarningContext: Usage of deprecated Model "Foo"
+    api.produce(Foo(value='Answer is: 42'))
+  /tmp/leapp-repository/repos/system_upgrade/el7toel8/actors/fooproducer/tests/test_unit_fooproducer.py:10: _DeprecationWarningContext: Usage of deprecated Model "Foo"
+    assert Foo(value='Answer is: 42') in produced_msgs.model_instances
+
+-- Docs: http://doc.pytest.org/en/latest/warnings.html
+======================================== 1 passed, 2 warnings in 0.13 seconds ========================================
+```
+
+As you can see the warning have been generated again. This time on two places,
+in `test_process` and `produce_answer` functions. Unless warning messages affect
+results of tests, we do not require strictly to handle them. However, it's a good
+practice. But if the warning log could affect the test (e.g. if a test function
+checks logs) it should be treated by the `suppress_deprecation` decorator too.
+In case of the library function, just add the `@suppress_deprecation(Foo)` line
+before the definition of the `produce_answer` function. But if we do the same
+for the test function, we will get an error (see that we have now just one
+deprecation warning now):
+
+```python
+| 21:59:57 | conftest | INFO | conftest.py | Actor 'foo_producer' context teardown complete
+
+repos/system_upgrade/el7toel8/actors/fooproducer/tests/test_unit_fooproducer.py::test_process FAILED
+
+====================================================== FAILURES ======================================================
+____________________________________________________ test_process ____________________________________________________
+
+args = (), kwargs = {'monkeypatch': <_pytest.monkeypatch.MonkeyPatch object at 0x7f21924b24d0>}
+suppressed = <class 'leapp.models.foo.Foo'>
+
+    @functools.wraps(target_item)
+    def process_wrapper(*args, **kwargs):
+        for suppressed in suppressed_items:
+            _suppressed_deprecations.add(suppressed)
+        try:
+            return target_item(*args, **kwargs)
+        finally:
+            for suppressed in suppressed_items:
+>               _suppressed_deprecations.remove(suppressed)
+E               KeyError: <class 'leapp.models.foo.Foo'>
+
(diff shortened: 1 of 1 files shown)


docs/source/deprecation.md
```diff
@@ -1,14 +1,117 @@
...
+             END OF USE OF DEPRECATED ENTITIES
+============================================================
+```
+
+Obviously we do not want to remove the actor yet, as in such a case, the
+model could be hardly called as deprecated. But we do not want the deprecation
+message to be produced in this case, as we want to keep the official
+functionality around that actor during the deprecation process, but the messages
+are focused on "custom" use of the model. E.g. in case when user create own
+actor producing/consuming a msg of the model. To fix this, suppress
```
bocekm, line 101:
```suggestion
message to be produced in this case, as we want to keep the original
functionality of that actor during the deprecation protection period. The warning messages
are focused on a "custom" use of the deprecated model - in case a user creates their own
actor producing/consuming a message of the model. To fix this, suppress
```

docs/source/deprecation.md
```diff
@@ -1,14 +1,117 @@
...
+the deprecation message in this actor. To do it, only thing we have to do
```
bocekm, line 102:
```suggestion
the deprecation message in this actor. To do it, the only thing you have to do
```

docs/source/deprecation.md
```diff
@@ -1,14 +1,117 @@
...
+decorator for several examples how to use it.
+
+Since you marked the entity as deprecated, user will be notified via messages
+and reports when entity is used in the code. It is obvious that as an author
+of the deprecation, you know that the entity is deprecated and you do not want
+to notify people that the code - that still provides the deprecated
+functionality - is used. To suppress deprecation messages in such cases,
+use the `suppress_deprecation` decorator taking as arguments objects that
+should not be reported by the deprecation. E.g. in case you use it above the
+definition of an actor, any use of the deprecated entity inside the actor
```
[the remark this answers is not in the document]
pirat89: The definition of an actor starts with `class`. It will be shown in examples.

docs/source/deprecation.md
```diff
@@ -1,14 +1,164 @@
...
+everything deprecated immadiately after the 6 months, but it's to be expected
+that it will be dropped anytime between 6 and 12 months since the deprecation.
+
+In case of issues, deprecated functionality is not going to be fixed since
+it is deprecated (unless it is fixed e.g. as a side-effect of another
+problem fix).
+
+## How does leapp informs about deprecated functionality?
+
+Mainly via release notes and changelogs. In the official leapp related projects
```
ZhukovGreen, line 36: Why not to be more specific here: Via release notes, changelogs, deprecation warnings (included to the upgrade report and pytest captured warning summary)

docs/source/deprecation.md
```diff
@@ -1,14 +1,164 @@
...
+  Usage of deprecated Model "Foo" @ /usr/lib/python2.7/site-packages/leapp/models/__init__.py:113
+  Near:         return cls(init_method='to_model', **data)
+  ```
+- a deprecated class (including model) is reported only when an instance of the
+  class (object) is created during the runtime
+- use of a deprecated function/method is reported only when the function is
+  called
+
+There are probably other cases, but these are the most common, that could
+confuse people.
```
ZhukovGreen, line 69: let's remove this as it sounds, basically, not reliable. If we face other problems, we have to update the documentation

docs/source/deprecation.md
```diff
@@ -1,14 +1,164 @@
...
+
+
+## How to deprecate entities in leapp
+
+When you want to deprecate an entity in leapp projects, use the `deprecated`
+decorator from `leapp.utils.deprecation` above the definition of the entity.
+The decorator has two input parameters to be able to specify the start
+of the deprecation protection period and a message explaining that
+particular deprecation. E.g. in case the deprecated functionality has
+a replacement, it is expected it will be mentioned in the msg.
```
ZhukovGreen, line 79: code is better than a thousand words:) and link to the code is even better, so we have a single source of the truth. Let;s point to https://github.com/oamg/leapp/tree/master/tests/data/deprecation-tests

docs/source/deprecation.md
```diff
@@ -1,14 +1,164 @@
...
+
+When you mark any entity as deprecated and this entity is then used
+in the code, leapp users
+, user will be notified about that via terminal and report messages.
+As the author of the deprecation, you know that the entity is deprecated
+and you do not want
+to notify people about the code that still uses the deprecated entity
+just for the reason to retain the original functionality. To suppress
+the deprecation messages in such cases,
+use the `suppress_deprecation` decorator taking as arguments objects that
```
ZhukovGreen, line 90: Are we assuming no warnings to the developers about approaching the end of deprecation period?

docs/source/deprecation.md
```diff
@@ -1,14 +1,164 @@
...
+should not be reported by the deprecation. E.g. in case you use it above the
+definition of an actor, any use of the deprecated entity inside the actor
+will not be reported.
+
+WARNING: It is forbidden to use the `suppress_deprecation` decorator for any
+purposes but one - retaining the original official functionality over the deprecation protection period.
+If you see the message and you are not the *provider* of the functionality, you
+have to update your code to be independent on it.
```
ZhukovGreen, line 98: I would remove it, because we already specified the use case of the `suppress_deprecation` decorator. Moreover, in my opinion, this sounds a bit rude to write `forbidden` without explaining the reason.

docs/source/deprecation.md
```diff
@@ -1,14 +1,164 @@
...
 
-### Warning message display behaviour
+## How are users notified about deprecation
 
-Warning messages are always shown after `snactor` executions via `snactor run` or `snactor workflow run`.
+Beside changelogs/release notes, warning messages are always shown after
+`snactor` executions via `snactor run` or `snactor workflow run`.
 At the end of executions of `leapp upgrade` or `leapp preupgrade` the warnings are added as report entries with a
 developer audience. If the deprecation is less than 6 months old the severity is set to 'MEDIUM' if it is older than
 6 months the severity is set to 'HIGH'
```
ZhukovGreen, line 164: We already facing this third time I guess in the text. Let' have it in one place

docs/source/deprecation.md
```diff
@@ -1,14 +1,117 @@
...
+to notify people that the code - that still provides the deprecated
+functionality - is used. To suppress deprecation messages in such cases,
+use the `suppress_deprecation` decorator taking as arguments objects that
+should not be reported by the deprecation. E.g. in case you use it above the
+definition of an actor, any use of the deprecated entity inside the actor
+will not be reported.
+
+WARNING: It is forbidden to use the `suppress_deprecation` decorator for any
+purposes but the one, to keep original official functionality for a while.
+If you see the message and you are not *provider* of the functionality, you
```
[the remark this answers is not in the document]
pirat89: @bocekm The deprecation warning message.

docs/source/deprecation.md
```diff
@@ -1,14 +1,164 @@
-## Deprecation
+# Deprecation
+
+The deprecation process is here to make (your) life of developers easier.
+It's not possible to write perfect solution for everything and as the project
+is evolving, it happens that some functionality needs to be changed, replaced
+or dropped completely. Such situations are inevitable. To reduce negative
+impact on your code, we introduce the deprecation process described below.
+
```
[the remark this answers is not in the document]
pirat89: Done.

docs/source/deprecation.md
```diff
@@ -1,14 +1,164 @@
...
+until the entity is removed completely.
+
+## What is the deprecation process for leapp
+
+In case some public public functions or models are to be removed for any
+reason, these need to be marked as deprecated before the removal (if possible).
+In our projects, the deprecation will be applied only for functions and models
+that have been introduced in an official release (IOW, a functionality that
+has been merged into the upstream, but has been removed before the release
+or was marked as experimental all the time, is going to be removed without
```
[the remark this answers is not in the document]
pirat89: As a developer, you look just to the released content if you want to be sure you are using stuff that is not experimental. Then it's clear what is included and what isn't.

docs/source/deprecation.md
```diff
@@ -1,14 +1,164 @@
...
+the deprecation state). The time period during which the deprecated entity
+won't be removed is at least 6 months. That doesn't mean we will remove
```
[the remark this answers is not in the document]
pirat89: You are not deciding anything. The time period is specified. When the protection time period ends, we can remove the deprecated functionality. But we do not have to remove it just for the purpose of the removal right? When we are not working on anything related, we could keep it there as it is if we want. From our POV, even more we should focus when we deliver the functionality to customer. Have an example: we deprecate something on Jan. The release happens on April. Another release will happen on July. Does it means we have to or should remove the stuff for the July release? No. We can keep it for example for the next November release. So people have enough time to fix their code.

docs/source/deprecation.md
```diff
@@ -1,14 +1,164 @@
...
+everything deprecated immadiately after the 6 months, but it's to be expected
```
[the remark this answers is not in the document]
pirat89: Fixed.

docs/source/deprecation.md
pirat89: Because here we present what is the real source of truth. We cannot expect that people will run pytest for their custom code. We cannot expect that the report message will be generated everytime. It's described in limitations of our solution below.

docs/source/deprecation.md
pirat89: For almost sure there could be others. But do we want to write all cases? The purpose of the message above is to show, that it's just additional level to inform people about deprecated stuff, that it will not inform you in 100% cases about all deprecated stuff and still, there is the only source of truth (changelogs, release notes).

docs/source/deprecation.md
pirat89: I am missing the connection between this line and the question. So far, people are awared about deprecated functionality. Hopefully, as it wil be related to developers, they read documentation (public and official one) where we will be clear about the time period.

docs/source/deprecation.md
pirat89: Code is below. And our test cases are really not best guidance always ;-)

docs/source/deprecation.md
pirat89: I thought the reasoning is clear from the text. If people developers do it,  end user (and even other developers) will not be notified, running the code, so it could happen that something will immediately break after we remove the deprecated functionality. Telling that reasoning in the warning seems for me like duplicate, as it should be really obvious. More importantly, what is suggestion for the correct text?

@bocekm What do you think?

docs/source/deprecation.md
pirat89: That's expected as it is draft and the document is not finished.

pirat89:
@bocekm @ZhukovGreen evertyhing till the end of the examples for the model deprecation, is kind of prepared for review (obviously, links to other parts in the doc, etc. are missing. this could be done later, when the whole doc is finished). Currently, we are missing explanation how to deprecate public classes (including explanation about nested classes and the use of the third parameter in the deprecate decorator, with explanation of the own limits).

docs/source/deprecation.md
```diff
@@ -1,17 +1,458 @@
...
+    fooproducer_lib.produce_answer()
+    assert _foo(value='Answer is: 42') in produced_msgs.model_instances
+```
+
+That's the whole solution for the *FooProducer* actor. Analogically to this,
+we need to treat the `FooConsumer` actor. You could notice, that all imports
+of the `Foo` model are commented. It's good practice as it is more visible
+to all developers, that a deprecated entity is present.
+
+### Example of a derived model deprecation
```
ZhukovGreen, line 361: All this chapter is very confusing to me. I think this is absolutely ok to deprecate the base model and expect that children will fire the deprecation warning. The only thing that we can do here is to improve the message of the deprecation warning so that it will contain the real source of the deprecation. but this is just a feature improvement to the deprecation functionality.

docs/source/deprecation.md
```diff
@@ -1,17 +1,458 @@
...
+
+## What is the deprecation process for leapp
+
+In case some public functions, models,... (let's say, basically whatever that
+is supposed to be used by other people) are to be removed for any
+reason, these need to be marked as deprecated before the removal (if possible).
+In our projects, the deprecation will be applied only for leapp entities
+that have been introduced in an official release (IOW, a functionality that
+has been merged into the upstream, but has been removed before the release
+or was marked as experimental all the time, is going to be removed without
```
ZhukovGreen, line 49: I still can't understand, why do you referring this official release. As a developer, I confused here. I don't know what is experimental all the time and I don't know is it included in the official release? Probably, what makes it clear to me if I read something like this:
1. Check if the entity you would like to deprecate existed before the last release (git checkout release/bump-version-XXXX and check if this entity still presents)
2. If no, then no deprecation needed, otherwise it is needed
3. Experimented functionality means the functionality is marked as experimntal decorator (just example, since I didn't seen it so far)

docs/source/deprecation.md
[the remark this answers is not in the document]
pirat89: @ZhukovGreen the date when you deprecate the functionality. e.g. the current day ;-) in case we would like to improve it, we can update the information before the release, in case it matters

docs/source/deprecation.md
```diff
@@ -1,17 +1,458 @@
...
+detail.
+
+But be aware of limitations of our detection in leapp and snactor utilities.
+Our detection mechanisms are not bulletproof. Here is list of some examples,
+when our detection does not work well:
+
+- when an actor a message of a deprecated model, the usage of the deprecated
+  model is detected only if the message has been produced before; IOW, in
+  case the message X is produced only on machines with a specific package,
+  the deprecated usage is not detected on machines without the packagw
```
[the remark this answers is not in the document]
pirat89: man, such comment is really not helping so much, when I have to read that several times before I find the issue. use suggestions in case it's just about a few changes. if you find a lot of changes, rather open pr with changes, so it's easier to go through that.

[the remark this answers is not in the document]
pirat89: fixed.

docs/source/deprecation.md
```diff
@@ -1,17 +1,458 @@
...
+
+@suppress_deprecation(Foo)
+class FooProducer(Actor):
+    """
+    Just produce the right answer to the world.
+    """
+
+    name = 'foo_producer'
+    consumes = ()
+    produces = (Foo,)
```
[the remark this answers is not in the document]
pirat89: Yeah. I could add something like that yet. I will add another example

docs/source/deprecation.md
```diff
@@ -1,17 +1,458 @@
...
+  /tmp/leapp-repository/repos/system_upgrade/el7toel8/actors/fooproducer/libraries/fooproducer_lib.py:5: _DeprecationWarningContext: Usage of deprecated Model "Foo"
+    api.produce(Foo(value='Answer is: 42'))
+  /tmp/leapp-repository/repos/system_upgrade/el7toel8/actors/fooproducer/tests/test_unit_fooproducer.py:10: _DeprecationWarningContext: Usage of deprecated Model "Foo"
+    assert Foo(value='Answer is: 42') in produced_msgs.model_instances
+
+-- Docs: http://doc.pytest.org/en/latest/warnings.html
+======================================== 1 passed, 2 warnings in 0.13 seconds ========================================
+```
+
+As you can see the warning have been generated again. This time on two places,
```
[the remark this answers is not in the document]
pirat89: It depends. As my teacher told, there are basically no rules for commas in english. It's just matter of taste. Basically one of rules for commas is e.g. the oxford comma, which is optional as well.

docs/source/deprecation.md
```diff
@@ -1,17 +1,458 @@
...
+in `test_process` and `produce_answer` functions. Unless warning messages affect
+results of tests, we do not require strictly to handle them. However it's good
```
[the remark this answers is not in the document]
pirat89: I think it's not good idea. We already discussed something like that in the past and we agreed in the end just on the soft way for custom/third party code. In such case, we are not talking about the deprecation. It should be there until we remove it. With no changed functionality. Such blockers do not make sense really. Maybe it could result in state when people stop writting tests.

docs/source/deprecation.md
```diff
@@ -1,17 +1,458 @@
...
+could be wrong, imagine unit test like this one (ignore the meaninfulness
+of this example test, it's just for education):
+
+```python
+from leapp.libraries.actor import fooproducer_lib
+from leapp.libraries.common.testutils import produce_mocked
+from leapp.libraries.stdlib import api
+from leapp.models import Foo  # deprecated model
+
+def test_process(monkeypatch):
```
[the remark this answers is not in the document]
pirat89: Do yo umean, because it is handled by the suppress_deprecation decorator in the actor.py file in such a case?

docs/source/deprecation.md
```diff
@@ -1,17 +1,458 @@
...
+======================================== 1 passed, 2 warnings in 0.13 seconds ========================================
+```
+
+As you can see the warning have been generated again. This time on two places,
+in `test_process` and `produce_answer` functions. Unless warning messages affect
+results of tests, we do not require strictly to handle them. However it's good
+practice. But if the warning log could affect the test (e.g. if a test function
+checks logs) it should be treated by the `suppress_deprecation` decorator too.
+In case of the library function, just add the `@suppress_deprecation(Foo)` line
+before the definition of the `produce_answer` function. But if we do the same
```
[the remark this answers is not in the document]
pirat89: @ZhukovGreen  I think that you are missing point what is deprecated. We are not deprecating the `produce_answer` function but the usage of the model in the function (see the `Foo` argument in the decorator). That's completely valid scenario and it is not workaround, it is how it works. This is trully real example which people could find out. As well, it's clear that in such case the deprecation doesn't have to be used before the actor definition. Please, read it carefully and then we can sync about how you understand it. From your answer it looks that the explanation is still not clear. If you understand it right from this comment, you can come up with improvement of the doc.

docs/source/deprecation.md
pirat89: Basically the thing you suggest is not possible without actions above the parser of the python code. That's definitely not something we want to do as we have no time right now. We have been thinking about the solution and really do not have a good one so far. This is the best we have and people have to live with that until we have another.

Btw, what is confusing on this section? How it can be improved?

docs/source/deprecation.md
pirat89: No, you do not have to check whether the functionality has been part of the last release or not (git blame can tell you that very quickly. btw) you can just deprecate it always as you want. but usually it doesn't make sense to create deprecation for something, that people have no possibility to use right? Same for the experimental think. When something is experimental (or there is clear disclaimer that people should not use it, it's experimental, ...) people cannot expect we will apply the deprecation process. we can, but why, as they couldn't use it?

docs/source/deprecation.md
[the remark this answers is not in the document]
ZhukovGreen: No, I removed the suppress_deprecation, and if you run test via current_actor_context, than you can't capture warnings (as well as logs, see https://github.com/oamg/leapp/pull/652). I think the reason is the same - running the actor in a separate process.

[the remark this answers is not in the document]
pirat89: Thanks for explanation.

docs/source/deprecation.md
ZhukovGreen: Skipping from commenting:
> We have been thinking about the solution and really do not have a good one so far

But something like https://gist.github.com/ZhukovGreen/10544d9ed7150f450056b1791616c626 will work

docs/source/deprecation.md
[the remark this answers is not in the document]
ZhukovGreen: Closing it and reword my comment in other place

docs/source/deprecation.md
```diff
@@ -1,17 +1,516 @@
...
+  /tmp/leapp-repository/repos/system_upgrade/el7toel8/actors/fooproducer/tests/test_unit_fooproducer.py:10: _DeprecationWarningContext: Usage of deprecated Model "Foo"
+    assert Foo(value='Answer is: 42') in produced_msgs.model_instances
+
+-- Docs: http://doc.pytest.org/en/latest/warnings.html
+======================================== 1 passed, 2 warnings in 0.13 seconds ========================================
+```
 
-This decorator will mark items as being deprecated and ensures that a deprecation warning is displayed if this item
-is being used.
+As you can see the warning have been generated again. This time on two places,
```
ZhukovGreen, line 291:
```suggestion
As you can see the warning have been generated again. This is because each time a deprecated model instantiated - the deprecation warning emitted.```

Then I think the further text until `### Example of replacing model` is more confusing than helpful. All you try to achieve is to eliminate one extra deprecation warning, but why is it necessary?

File: docs/source/deprecation.md (hunk 15 above)
Comment (ZhukovGreen) [reply docs/source/deprecation.md:419]: As a simple hack, we might just update the deprecation message to look something like this:
before
```
Usage of deprecated Model "BaseFoo"
```
after
```
Usage of deprecated Model "BaseFoo" or models inherited from it
```

File: docs/source/deprecation.md (hunk 13 above)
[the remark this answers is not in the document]
Comment (pirat89) [reply docs/source/deprecation.md:51]: Hmm.. the date doesn't make sense. It's deprecated in the day of the commit. So since == today. not 6months in future.

File: docs/source/deprecation.md (hunk 16 above)
Comment (ZhukovGreen) [reply docs/source/deprecation.md:49]: resolved over the the call

File: docs/source/deprecation.md (hunk 13 above)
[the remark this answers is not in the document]
Comment (ZhukovGreen) [reply docs/source/deprecation.md:51]: resolved over the call

File: docs/source/deprecation.md [hunk 24]
```diff
@@ -6,46 +6,41 @@ is evolving, it happens that some functionality needs to be changed, replaced
 or dropped completely. Such situations are inevitable. To reduce negative
 impact on your code, we introduce the deprecation process described below.
 
-We are still working on improvement of the current situation to provide stable
-interfaces for people, so extensibility of the current solutions could be much
-easier. This is just on of those improvements.
```
bocekm: The whole paragraph is quite hard to grasp. It's so abstract that I don't even know what is the message I should take from it. I think it can be removed all.

docs/source/deprecation.md
pirat89: That's not exact true as well. When you do something like this:
```
for i in [1,1]:
    self.produce(DeprecatedFoo())
```
you will get just one warning - not two.

docs/source/deprecation.md
pirat89: the removal is ok for me

docs/source/deprecation.md
pirat89: > But something like https://gist.github.com/ZhukovGreen/10544d9ed7150f450056b1791616c626 will work

Currently, we need to write documentation regarding the functionality in the released version. The proposal seems nice, but it's not part of the release, so it could be improvement for the future. As well, we have to have versioned documentation vefore we start to do updates of doc regarding the upstream devel changes.

docs/source/deprecation.md
pirat89: > Then I think the further text until ### Example of replacing model is more confusing than helpful. All you try to achieve is to eliminate one extra deprecation warning, but why is it necessary?

@ZhukovGreenbecause the point is that there shouldn't be any warning message.

docs/source/deprecation.md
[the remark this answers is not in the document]
Rezney: I agree with the coma comment...

docs/source/deprecation.md
```diff
@@ -1,17 +1,515 @@
...
+1. Public APIs
+1. Actors providing functionality that could be used by any developer (produce
+   or consume messages)
+
+In other words, private classes, private functions or anything in private
+libraries, may be modified or removed without the deprecation process. As well,
+it's possible we will need to change something (e.g. a behaviour of a function)
+that will not be possible to cover reasonably by the deprecation process (e.g.
+change output of the function...). We'll try our best to prevent it, but it may
+happen. To limit such problems, we recommend people to use APIs as much
```
Rezney, line 25: Let's add hyperlink to the APIs...

pirat89: Discussed in person. That's not one specific API, it's generic (one is provided by the framework, one through shared libraries in a repository, another will be provided by workflowAPIs,...)

docs/source/deprecation.md
```diff
@@ -1,17 +1,516 @@
...
+Our detection mechanisms are not bulletproof. Here is list of some examples,
+when our detection does not work well:
+
+- when an actor a message of a deprecated model, the usage of the deprecated
+  model is detected only if the message has been produced before; IOW, in
+  case the message X is produced only on machines with a specific package,
+  the deprecated usage is not detected on machines without the package
+- when a message of a deprecated model is consumed, the message points
+  internals of the leapp framework, instead of place where the message is
+  consumed, e.g.
```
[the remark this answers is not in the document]
pirat89: @bocekm  read the example below. Do you expect reported such a place of the usage of the deprecated entity instead. Internals of the framework instead of actor. Is it clear now? If yes, any suggestion for more clear description?

pirat89:
Just to ensure we are all on the same page, keep in mind that this long doc could be split into doc about deprecation & tutorial. It doesn't have to be in future one big document. Just now for the purpose of writting&reading, I have everything together.

docs/source/deprecation.md
```diff
@@ -1,17 +1,515 @@
...
+============================================================
+```
+
+Apparently, the `Reason` is not so good. It's just example. In real world
+example, you would like to provide usually a little bit better explanation.
+Anyway, much more interesting is the point, that the message is now printed
+everytime the actor is executed.
+
+Obviously we do not want to remove the actor yet, because in such a case, the
+model could be hardly called as deprecated - we need to provide the same
```
ZhukovGreen, line 176:
```suggestion
model could be hardly called as deprecated - we need to keep the same
```

docs/source/deprecation.md
ZhukovGreen: Considering all the further information you have to add to explain how to get rid of the warning during the test, I believe it is worthless and still more confusing than helpful. I think it is not good to have this ind of workaround with def _foo()

docs/source/deprecation.md
ZhukovGreen: ok

ZhukovGreen (approved):
I added my comments, but we can merge it, and resolve them somewhere in the future

pirat89:
> Considering all the further information you have to add to explain how to get rid of the warning during the test, I believe it is worthless and still more confusing than helpful. I think it is not good to have this ind of workaround with def _foo()

@ZhukovGreen I do not agree. You expect, that the warning do not affect results of tests - which doesn't have to be true. Deprecation warnings could affect results of tests and if you have a lot of tests, it's easier to properly apply decorators, instead of modification of all tests.

docs/source/deprecation.md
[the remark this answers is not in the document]
pirat89: fixed

docs/source/deprecation.md
pirat89: Skipping for now. It can be improved in the next iteration.

ZhukovGreen (approved):
Approving it, thx

bocekm (approved):

PR #664, opened 2020-09-09
Fix deprecation

pirat89:
Two fixes:
1. The suppress_deprecation decorator cannot be used on every class. It's expected a class has the process method. Instead of confusing simple crash, raise an exception with a meaningful message when decorator is used on a class without the process method.

1. Handle 'nested' suppress of deprecation
    
    Imagine case (pseudocode):
    ```pseudocode
      (Ddep is deprecated entity)
    
      @suppress_deprecation(Ddep)
      def fA():
          create Ddep
          fB()
    
      @suppress_deprecation(Ddep)
      def fB():
          create Ddep
    ```
    
    In such a case, apps crashed, because we store information about
    suppressed deprecation into the set. After you leave the fB function
    the Ddep entity is removed from the set and after end of fA function
    there is no way to remove it again - so the app crashes. This commit
    fixes the problem.

The first case is not covered by tests. The second one it is.

### DIFF
--- a/leapp/utils/deprecation.py
+++ b/leapp/utils/deprecation.py
@@ -15,7 +15,7 @@ def suppress_deprecation(*suppressed_items):
     Decorator to suppress deprecation warnings during the execution of the decorated entity.
 
     This can be used directly on Actor classes OR functions. This relies on the fact that a class has a `process`
-    method, if the method does not exist, this may cause mayhem.
+    method, if the method does not exist the ValueError exeption is raised.
 
     :param suppressed_items: Variable number of arguments each for entities to suppress the deprecation warnings for.
     """
@@ -33,13 +33,15 @@ def suppress_deprecation(*suppressed_items):
 
         @functools.wraps(target_item)
         def process_wrapper(*args, **kwargs):
-            for suppressed in suppressed_items:
-                _suppressed_deprecations.add(suppressed)
+            # we need to remove later just items that we add right now
+            # added_items == new items we add now to ...
+            added_items = set(suppressed_items) - _suppressed_deprecations
+            _suppressed_deprecations.update(added_items)
             try:
                 return target_item(*args, **kwargs)
             finally:
-                for suppressed in suppressed_items:
-                    _suppressed_deprecations.remove(suppressed)
+                # remove the added items
+                _suppressed_deprecations.difference_update(added_items)
         if inspect.isclass(item):
             item.process = process_wrapper
             return item
--- a/leapp/utils/deprecation.py
+++ b/leapp/utils/deprecation.py
@@ -23,6 +23,13 @@ def suppress_deprecation(*suppressed_items):
         target_item = item
         if inspect.isclass(item):
             target_item = getattr(item, 'process', None)
+            if target_item is None:
+                raise ValueError(
+                    'The suppress_deprecation decorator has been used on a class which'
+                    ' does not contain the process method. The decorator can be used'
+                    ' only for classes with that method (e.g. classes derived from Actor).'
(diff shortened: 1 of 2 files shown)


pirat89:
e2e test

leapp/utils/deprecation.py
```diff
@@ -15,24 +15,34 @@ def suppress_deprecation(*suppressed_items):
...
     :param suppressed_items: Variable number of arguments each for entities to suppress the deprecation warnings for.
     """
     def decorator(item):
         target_item = item
         if inspect.isclass(item):
             target_item = getattr(item, 'process', None)
+            if target_item is None:
+                # FIXME: what error to raise?
+                raise ValueError(
+                    'The suppress_deprecation decorator has been used on class which'
```
vinzenz, line 29:
```suggestion
                    'The suppress_deprecation decorator has been used on a class which'
```

pirat89: done

leapp/utils/deprecation.py
```diff
@@ -15,24 +15,34 @@ def suppress_deprecation(*suppressed_items):
...
+    method, if the method does not exist the ValueError exeption is raised.
 
     :param suppressed_items: Variable number of arguments each for entities to suppress the deprecation warnings for.
     """
     def decorator(item):
         target_item = item
         if inspect.isclass(item):
             target_item = getattr(item, 'process', None)
+            if target_item is None:
+                # FIXME: what error to raise?
```
Rezney, line 27: Please delete the FIXME...

leapp/utils/deprecation.py
```diff
@@ -15,24 +15,34 @@ def suppress_deprecation(*suppressed_items):
     Decorator to suppress deprecation warnings during the execution of the decorated entity.
 
     This can be used directly on Actor classes OR functions. This relies on the fact that a class has a `process`
-    method, if the method does not exist, this may cause mayhem.
+    method, if the method does not exist the ValueError exeption is raised.
```
Rezney, line 18:
```suggestion
    method, if the method does not exist the ValueError exception is raised.
```

leapp/utils/deprecation.py
```diff
@@ -15,24 +15,34 @@ def suppress_deprecation(*suppressed_items):
...
+                    ' does not contain the process method. The decorator can be used'
+                    ' only for classes with that method (e.g. classes derived from Actor).'
+                    ' Use the decorator on the affected methods instead of the current class.'
+                )
 
         @functools.wraps(target_item)
         def process_wrapper(*args, **kwargs):
-            for suppressed in suppressed_items:
-                _suppressed_deprecations.add(suppressed)
+            # we need to remove later just items, that we add right now
```
Rezney, line 37:
```suggestion
            # we need to remove later just items that we add right now
```

Rezney:
Tested. Works as expected.

Scenario 1:
~~~
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py", line 73, in <module>
    class _InputData(object):
  File "/usr/lib/python2.7/site-packages/leapp/utils/deprecation.py", line 28, in decorator
    'The suppress_deprecation decorator has been used on a class which'
ValueError: The suppress_deprecation decorator has been used on a class which does not contain the process method. The decorator can be used only for classes with that method (e.g. classes derived from Actor). Use the decorator on the affected methods instead of the current class.
~~~

Scenario 2:

does not fail using the actor bundled in the tests in our wokflow.

Rezney (approved):

PR #665, opened 2020-09-14
Fix report deprecation

pirat89:
The conversion of deprecation messages to reports was broken on Python3 because `haslib.sha256` requires byte string instead of Unicode (missing `.encode(utf-8)`). Added couple of tests to cover this part of code.

Also updated the .gitignore file so the `tests/scripts` directory is not ignored anymore.


drehak:
Tested, reproduced, the fix works. LGTM.

drehak (approved):

PR #669, opened 2020-10-06
Add a stable report identifier

fernflower:
A 'key' field will be added to each report entry. Same report entries on different machines should have the same 'key' value.

From the user/contributor perspective the interface doesn't change, if the Key with the unique report key is not defined in the report entries passed to the create_report command, the framework will take care of stable identifiers by generating one from the required report fields or will raise a ValueError in case LEAPP_DEVEL_FIXED_REPORT_KEY environment
variable is defined.
This fallback behavior will be the last resort once all actors in the leapp-repository are updated.

### DIFF
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -1,7 +1,10 @@
 import datetime
 import json
+import hashlib
+import os
 
 from leapp.compat import string_types
+from leapp.logger import configure_logger
 from leapp.models import fields, Model, ErrorModel
 from leapp.topics import ReportTopic
 from leapp.libraries.stdlib.api import produce
@@ -123,6 +126,14 @@ class Flags(BasePrimitive):
         if not isinstance(value, list):
             raise TypeError('Value of "Flags" must be a list')
         self._value = value
+
+
+class Key(BasePrimitive):
+    """Stable identifier for report entries."""
+    name = 'key'
+
+    def __init__(self, uuid):
+        self._value = uuid
 
 
 class Tags(BasePrimitive):
@@ -272,6 +283,37 @@ def _sanitize_entries(entries):
         entries.append(Severity('info'))
     if not any(isinstance(e, Audience) for e in entries):
         entries.append(Audience('sysadmin'))
+    _check_stable_key(entries, raise_if_undefined=os.getenv('LEAPP_DEVEL_FIXED_REPORT_KEY', '0') != '0')
+
+
+def _check_stable_key(entries, raise_if_undefined=False):
+    """
+    Inserts a Key into the entries of the report if it doesn't exist.
+
+    The resulting id is calculated from the mandatory report
+    entries (severity, audience, title).
+    Please note that the original entries list is modified.
+    """
+    def _find_entry(cls):
+        entry = next((e for e in entries if isinstance(e, cls)), None)
+        if entry:
+            return entry._value
+
+    if _find_entry(Key):
+        # Key is already there, nothing to do
+        return
+
+    if raise_if_undefined:
+        raise ValueError('Stable Key report entry with specified unique value not provided')
+
+    # No Key found - let's generate one!
+    logger = configure_logger()
+    key_str = u'{severity}:{audience}:{title}'.format(severity=_find_entry(Severity),
+                                                      audience=_find_entry(Audience),
+                                                      title=_find_entry(Title))
+    key_value = hashlib.sha1(key_str.encode('utf-8')).hexdigest()
+    entries.append(Key(key_value))
+    logger.warning('Stable Key report entry not provided, dynamically generating one - {}'.format(key_value))
 
 
 def _create_report_object(entries):
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -133,7 +133,7 @@ class Key(BasePrimitive):
     name = 'key'
 
     def __init__(self, uuid):
-        self._value = uuid
+        self._value = str(uuid)
 
 
 class Tags(BasePrimitive):
@@ -299,15 +299,19 @@ def _check_stable_key(entries, raise_if_undefined=False):
         if entry:
             return entry._value
 
-    if _find_entry(Key):
-        # Key is already there, nothing to do
+    logger = configure_logger()
+    key = _find_entry(Key)
+    if key is not None:
+        # Key is already there, check that it's not an empty string
+        if not key.strip():
+            logger.error('An empty string is specified as Key')
+            raise ValueError('Key can not be an empty string.')
         return
 
     if raise_if_undefined:
         raise ValueError('Stable Key report entry with specified unique value not provided')
 
     # No Key found - let's generate one!
-    logger = configure_logger()
     key_str = u'{severity}:{audience}:{title}'.format(severity=_find_entry(Severity),
                                                       audience=_find_entry(Audience),
                                                       title=_find_entry(Title))
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -2,6 +2,7 @@ import datetime
 import json
 import hashlib
 import os
+import six
 
 from leapp.compat import string_types
 from leapp.logger import configure_logger
@@ -134,7 +135,7 @@ class Key(BasePrimitive):
 
     def __init__(self, uuid):
         # uuid is allowed to be string or unicode only, checking in py2/py3 friendly mode
-        if not isinstance(uuid, unicode) and not isinstance(uuid, str):
+        if not isinstance(uuid, six.string_types):
             raise ValueError('Key value should be a string.')
 
         self._value = uuid
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -133,8 +133,10 @@ class Key(BasePrimitive):
     name = 'key'
 
     def __init__(self, uuid):
-        if not isinstance(uuid, str):
+        # uuid is allowed to be string or unicode only, checking in py2/py3 friendly mode
+        if not isinstance(uuid, unicode) and not isinstance(uuid, str):
             raise ValueError('Key value should be a string.')
+
         self._value = uuid
 
 
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -133,7 +133,9 @@ class Key(BasePrimitive):
     name = 'key'
 
     def __init__(self, uuid):
-        self._value = str(uuid)
+        if not isinstance(uuid, str):
+            raise ValueError('Key value should be a string.')
+        self._value = uuid
 
 
 class Tags(BasePrimitive):
--- a/leapp/dialogs/dialog.py
+++ b/leapp/dialogs/dialog.py
@@ -18,7 +18,7 @@ class Dialog(object):
         address values.
     """
 
-    def __init__(self, scope, reason, title=None, components=None):
+    def __init__(self, scope, reason, title=None, components=None, key=None):
         """
 
         :param scope: Unique scope identifier for the data to be stored in the answer files. Scope + component key
@@ -30,11 +30,14 @@ class Dialog(object):
         :type title: str
         :param components: Components to display in the given order in the dialog
         :type components: tuple(leapp.dialogs.components.Component)
+        :param key: Key to appear in the dialog-related report entry
+        :type key: str
         """
         self.components = components or self.components
         self.title = title
         self.scope = scope
         self.reason = reason
+        self.key = key
         self._store = None
         self._min_label_width = None
 
--- a/leapp/messaging/__init__.py
+++ b/leapp/messaging/__init__.py
@@ -157,7 +157,11 @@ class BaseMessaging(object):
         userchoices = dialog.get_answers(self._answers)
         if not userchoices:
             # produce DialogModel messages for all the dialogs that don't have answers in answerfile
-            self.produce(DialogModel(actor=actor.name, answerfile_sections=dialog.answerfile_sections), actor)
+            stable_key = dialog.key if dialog.key else hashlib.sha1(
+                ','.join(sorted(dialog.answerfile_sections.keys())).encode('utf-8')).hexdigest()
+            self.produce(DialogModel(actor=actor.name,
+                                     answerfile_sections=dialog.answerfile_sections,
+                                     key=stable_key), actor)
         else:
             # update dialogs with answers from answerfile. That is necessary for proper answerfile generation
             for component, value in userchoices.items():
--- a/leapp/models/__init__.py
+++ b/leapp/models/__init__.py
@@ -156,6 +156,7 @@ class DialogModel(Model):
     answerfile_sections = fields.JSON()
     actor = fields.String()
     details = fields.Nullable(fields.String())
+    key = fields.Nullable(fields.String())
 
 
 def get_models():
--- a/leapp/utils/report.py
+++ b/leapp/utils/report.py
@@ -100,6 +100,7 @@ def generate_report_file(messages_to_report, context, path):
                 remediation = Remediation.from_dict(message.get('detail', {}))
                 if remediation:
                     f.write('Remediation: {}\n'.format(remediation))
+                f.write('Key: {}\n'.format(message['key']))
                 f.write('-' * 40 + '\n')
     elif path.endswith(".json"):
         with open(path, 'w') as f:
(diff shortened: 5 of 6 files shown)


leapp/reporting/__init__.py
```diff
@@ -125,6 +126,23 @@ def __init__(self, value=None):
...
+class Key(BasePrimitive):
+    """Stable identifier for report entries"""
+    name = 'key'
+
+    def __init__(self, entries):
+
+        def _find_entry(cls):
+            entry = next((e for e in entries if isinstance(e, cls)), None)
+            if entry:
+                return entry._value
```
ZhukovGreen, line 138: No principal comment:
seems it is not needed to be inside the constructor

leapp/reporting/__init__.py
```diff
@@ -125,6 +126,23 @@ def __init__(self, value=None):
...
 
 
+class Key(BasePrimitive):
+    """Stable identifier for report entries"""
+    name = 'key'
+
+    def __init__(self, entries):
+
+        def _find_entry(cls):
+            entry = next((e for e in entries if isinstance(e, cls)), None)
```
ZhukovGreen, line 136: not principal:
in python usually preferred https://en.wikipedia.org/wiki/EAFP

leapp/reporting/__init__.py
```diff
@@ -125,6 +126,23 @@ def __init__(self, value=None):
...
+                return entry._value
+
+        record_str = u'[{severity}:{audience}]{title}'.format(severity=_find_entry(Severity),
+                                                              audience=_find_entry(Audience),
+                                                              title=_find_entry(Title))
+        self._value = hashlib.sha1(record_str.encode('utf-8')).hexdigest()
+
+
 class Tags(BasePrimitive):
     """Report tags"""
```
ZhukovGreen, line 147:
```suggestion
    """Report tags."""
```

ZhukovGreen (commented):
Just a few comments

leapp/reporting/__init__.py
fernflower: I'd keep it as is, it's shorter

leapp/reporting/__init__.py
fernflower: But it's a helper function that is not needed anywhere else anyway.

leapp/reporting/__init__.py
fernflower: Unrelated to this patch

leapp/reporting/__init__.py
ZhukovGreen: Outside of the constructor. Now each time you instantiate the Key a new function object of _find_entry is created. If you would have the function in the outer scope, then only one object will be presented. There are no reasons to keep it as a nested function, basically. But as the result of this solution, you will have some downsides, i.e. hard to mock, hard to unit test.

As an alternative you can make it as:
```
@classmethod
def _find_entry(cls):
    """Docs."""
```

ZhukovGreen (commented):
Just some replies

pirat89:
@fernflower I am not sure whether I get it, but if you plan to provide keys for reports from leapp-repository via a file in leapp, it's no go.

leapp/reporting/__init__.py
```diff
@@ -125,6 +128,30 @@ def __init__(self, value=None):
         self._value = value
 
 
+class Key(BasePrimitive):
+    """Stable identifier for report entries"""
```
pirat89, line 132: Regarding the guidelines, the DOT should be part of every docstring title.

```suggestion
    """Stable identifier for report entries."""
```

leapp/reporting/__init__.py
ZhukovGreen: Problem with mocking and with unit testing linked with the problem to access the _find_entry object, because it is not available in the outer scope.  

Unit tests are exactly for testing such things like your function (_find_entry performs a piece of distinct functionality). Integration tests are for testing the whole thing. Unit test means testing each single unit, which allows you to focus on testing the very tiny piece of you functionality - a function and think about corner cases within a smaller scope.

Anyway, as you see in the header of my comment - it is not principle, so the only reason to maintain this thread is it is useful for me and you and it is not blocking it from progress. So consider resolving this treah anytime you want.

pirat89:
@fernflower btw, remember to bump leapp-framework in the PR (and if another following PR is in leapp-repository, the leapp-framework requested values should be bumped there as well)

ZhukovGreen:
@fernflower Hi Inna, will you add a test case when the key wasn't provided inside the report?

fernflower:
> @fernflower Hi Inna, will you add a test case when the key wasn't provided inside the report?

Already there in all flavors I could think of https://github.com/oamg/leapp/pull/669/files#diff-298ab46eb87d51b2b477ad88e9730c63ea0885d7f797933de81e9a246b96c8f1

Did I miss some testcase?

ZhukovGreen:
@fernflower already discussed that on IRC. Thx

leapp/reporting/__init__.py
```diff
@@ -272,6 +283,34 @@ def _sanitize_entries(entries):
...
+    if _find_entry(Key):
+        # Key is already there, nothing to do
+        return
+    if raise_if_undefined:
+        raise ValueError('Stable Key report entry with specified unique value not provided')
+    else:
+        logger = configure_logger()
+        record_str = u'[{severity}:{audience}]{title}'.format(severity=_find_entry(Severity),
+                                                              audience=_find_entry(Audience),
+                                                              title=_find_entry(Title))
```
ZhukovGreen, line 310: As I remember from our tech.cabal meeting, we would like that all reports have a unique identifier, independent on the report content. Just any new report added to the codebase of some workflow (i.e. IPU) should have this key.

Then why do you generating a key. It is not added to the codebase and moreover, with changing let's say title of the report the key will become different.

ZhukovGreen (commented):
Adding a question which might affect the implementation gradually
/cc @fernflower @pirat89

leapp/reporting/__init__.py
fernflower: Changes to the actors will take place in leapp. 
Having dynamic fallback as a backup while having fixed ids set for all the report entries (in the follow up patches in the leapp-repository) and requiring all new code to have Key defined in the contributor documentation will give just the needed result.

ZhukovGreen: Our ideal state is all reports has a unique key. Leapp framework checks that all reports has it. What is the benefit to add unique keys in leapp-repos later, comparing to do it right now and merge this MR and MR in leapp-repos together. Moreover, I don't think there are a lot of work to add unique identifier to all lreports inside leapp-repos.
So I would go with:
1. Additional MR for leapp-repository, which adds unique ids to all reports (some script probably should be created)
2. Simplify this MR to the state, where it is just checking that all reports have a key, or raises an error
3. Merge these MRs together  
4. Require to rebase all open MRs in leapp-repository 

@pirat89 @vinzenz what do you think guys?

fernflower: My approach is a flawless gradual adoption of the feature that is even not there yet.

You are suggesting a make-everything-at-once change that will complicate life of every single leapp developer. Why?

ZhukovGreen: Maybe you're right, there are actors written in-house and they have own reports. In this case, a deprecation process may be applied to the reports.

shaded-enmity: I think the only step we really need is 1/ @ZhukovGreen 

* Once this is merged, add explicit key to all currently existing calls of `create_report` in Leapp repository
  * All PRs will need to be rebased anyway, but I think we don't need to insist on rebasing everyhing *now*, only the PRs that are currently active
* Update our CI configuration to set `export LEAPP_DEVEL_FIXED_REPORT_KEY=1`

The problem with 2/ is that we're basically asking every 3rd party out there to refactor their actors else exceptions, and the automatic generation is nice middle ground.

@fernflower one thing I think we should solve in this PR is how are we going to treat report entries generated as a result of a dialog. IIRC report entries generated from dialogs have static title, audience and severity so they will all end up with the same digest. I think dialogs should also be able to set their key or autogenerate one from the actual question titles.

fernflower: @shaded-enmity about the dialogs - I'd say it is impossible to take care at the framework level, as currently the report about upgrade inhibition because of the unanswered dialog is happening in the leapp-repository https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/verifydialogs.py#L23 

So this will be fixed by proper generation of the Key in the verifydialogs actor only, can't think of a nice way to embed this into the framework. What do you think?

fernflower: @pirat89 Thanks for pointing out, I will keep that in mind.
This PR however shouldn't affect anyone, it is the appropriate change in the actors and our guidelines, which, once merged, will require al yet-to-be-merged-actors to contain Key parameter.

pirat89: @fernflower you are right. That's caused by my fast reading-thinking today :) So far, until users will see any issues, it should be ok to merged without problems earlier than the other changes are implemented and require changes then.

ZhukovGreen: OK, then the origin of my concern is resolved. Resolving the comment

shaded-enmity (commented):
I think we're mostly good to go, however what is not really clear to me is how do I set an ID for a dialog? i.e. add `key` field to the `DialogModel`? https://github.com/oamg/leapp/blob/master/leapp/models/__init__.py#L153 so that `verifydialogs` in `leapp-repository` can either forward what is in the model or fallback to autogeneration.

fernflower:
> I think we're mostly good to go, however what is not really clear to me is how do I set an ID for a dialog? i.e. add `key` field to the `DialogModel`? https://github.com/oamg/leapp/blob/master/leapp/models/__init__.py#L153 so that `verifydialogs` in `leapp-repository` can either forward what is in the model or fallback to autogeneration.

Great point, yeah, adding a key into the Dialog class and DialogModel should do this.

But given this I am now questioning if we need "Dialog key auto generation" introduced in the last commit? Maybe it is cleaner to go with "by default all dialogs missing answers will have the same key generated by default" and then in the parallel leapp-repository patch specifically use a dialog.key in the create_report https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/verifydialogs.py#L23? I am really not in favor of this "autogenerate the dialog stable report key in the framework by utilizing the implementation details of RelatedResource(sectionX) being present in the report fields" approach I had to use in the last commit :(

ZhukovGreen (approved):
Looks really good to me!

For test_create_report_stable_key I would go with the pytest.parametrize, as the current impl seems a bit bulky, but this is really a matter of taste.

shaded-enmity:
+1 that the current tight coupling between leapp and leapp-repository via `verifydialogs` is a bit unwieldy (well, more than a bit I guess), but honestly I have no idea how to break the cycle. What I think could work:

1. Introduce a new `Phase` flag, like `ReportCollection`
2. Move `verifydialogs` somewhere to the framework
3. Run `verifydialogs` actor in the `ReportCollection` flagged phase
4. (maybe) ensure no other actors are run in the phase

This way we still retain the flexibility that workflows can decide *when* to collect+verify reports by flagging the particular phase with the `ReportCollection` flag (or skip completely if there's no phase with that flag).

Same ID for all dialog-based reports makes me really nervous, I can see people automating based on IDs and waiving more dialogs than they expect.

fernflower:
> Same ID for all dialog-based reports makes me really nervous, I can see people automating based on IDs and waiving more dialogs than they expect.

No-no, I meant temporary (the "stable report key" is a feature that is not used anyway yet) until a parallel patch lands into leapp-repository for verifydialogs actor, smth like 

```
--- a/repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/verifydialogs.py
+++ b/repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/verifydialogs.py
@@ -19,5 +19,6 @@ def check_dialogs(inhibit_if_no_userchoice=True):
                        reporting.Severity(reporting.Severity.HIGH),
                        reporting.Summary(summary.format('\n'.join(sections))),
                        reporting.Flags([reporting.Flags.INHIBITOR] if inhibit_if_no_userchoice else []),
-                       reporting.Remediation(hint=dialogs_remediation, commands=cmd_remediation)]
+                       reporting.Remediation(hint=dialogs_remediation, commands=cmd_remediation),
+                       reporting.Key(dialog.key)]

```

vojtechsokol:
leapp-ci build

AloisMahdal:
leapp-ci build

AloisMahdal:
As of

      - leapp-0.11.0-0.202101051232Z.7f5d189.stablereportid.PR669.el7_8
      - leapp-repository-0.12.0-100.202012030845Z.2c7ba21.master.el7_6

all acceptance tests passed.

From QE POV this is ready for merge.

AloisMahdal:
> From QE POV this is ready for merge.

Sorry, I forgot about Myrphy Regression tests.  I've scheduled them now.

AloisMahdal:
I've checked that the failure in murphy-ci/integration fail is irrelevant to leapp and can be safely waived.

IOW, this is *ready to merge.*

fernflower:
@AloisMahdal unfortunately, this patch can't be merged alone - only with the leapp-repository PR618. Do the 2 prs together pass all the tests?

fernflower:
@AloisMahdal That's a fantastic example of the importance of cross-repo testing - the failure you discovered with the leapp-repository patch is actually caused by the leapp bug. 
Should be fixed with the latest commit, once the build is there 2 patches can be retested together.

By the way the upgrade was never continued, it broke pretty bad but didn't run without the filled-in answerfile.

AloisMahdal:
Fixed: as of

    leapp-0.11.0-0.202101111435Z.04c5ef0.stablereportid.PR669.el7_8
    leapp-repository-0.12.0-0.202101081216Z.5cf1bc4.stablereport.PR618.el7_8

the upgrade is inhibited properly:

    Risk Factor: high (inhibitor)
    Title: Missing required answers in the answer file
    Summary: One or more sections in answerfile are missing user choices: remove_pam_pkcs11_module_check.confirm
    For more information consult https://leapp.readthedocs.io/en/latest/dialogs.html
    Remediation: [hint] Please register user choices with leapp answer cli command or by manually editing the answerfile.
    [command] leapp answer --section remove_pam_pkcs11_module_check.confirm=True
    Key: d35f6c6b1b1fa6924ef442e3670d90fa92f0d54b
    Risk Factor: high
    Title: Upgrade is unsupported
    Summary: Environment variable LEAPP_UNSUPPORTED has been detected. A successful and safe upgrade process cannot be guaranteed. From now on you are continuing at your own risk.

    Key: 9e5088e3c1f371e020ec777c3d86578f4be143cf
    Risk Factor: high
    Title: Difference in Python versions and support in RHEL 8
    Summary: In RHEL 8, there is no 'python' command. Python 3 (backward incompatible) is the primary Python version and Python 2 is available with limited support and limited set of packages. Read more here: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/configuring_basic_system_settings/#using-python3
    Remediation: [hint] Please run "alternatives --set python /usr/bin/python3" after upgrade
    Key: 0c98585b1d8d252eb540bf61560094f3495351f5
        [...]
        [...]
        [...]

Thanks!

pirat89 (approved):
Lgtm and it seems it works as expected.

pirat89:
@fernflower you can do it in separate PR, but pls update the doc abou tthe reporting to reflect the possibility to add Key - including how people should generate the key.

AloisMahdal:
latest **murphy-ci/*** failures are caused by internal bug (my bad -- sorry) which is fixed now; re-running

AloisMahdal:
**murphy-ci/integration** - I have confirmed that the fail can be waived safely.

i.e.: from QE POV, this PR is *good to go*.

fernflower:
leapp-ci build

fernflower:
leapp-ci build

pirat89:
leapp-ci build

pirat89:
just for info, current implementation includes the severity field when the default key is generated, but in case of the `spamassassinconfigcheck` actor in the official leapp-repository repo, the severity is variable - which is valid case - and I would expect really stable ID. In this particular case it can be easily fixed defining the particular stable id in the report. but just pointing the fact the severity is expected right now to be variable in some cases and from my pov it still about only one specific report.

pirat89:
leapp-ci build

fernflower:
e2e tests

examon:
e2e tests

pirat89:
@Rezney it's not test. it's waiting for https://github.com/oamg/leapp-repository/pull/637 - tests already passed so I am going to merge it and rebase.

pirat89:
@fernflower  :partying_face:

PR #670, opened 2020-10-13
spec: Remove pytest residuals in spec file

vinzenz:
In order build properly on public copr, we drop all the pytest
build dependencies as it is not necessary to run the tests during the
build.

Related: https://issues.redhat.com/browse/OAMG-3811


vinzenz:
Here's a test build triggered manually by telling it to look at this PR
https://copr.fedorainfracloud.org/coprs/g/oamg/leapp/build/1705872/

PR #672, opened 2020-10-19
Add new field type = field.Dict

ZhukovGreen:
Ref https://github.com/oamg/leapp-repository/issues/593

This structure validates types of all keys and values of the model

Usage example:
```python
class RestrictedPCIDevices(Model):
    driver_names = fields.Dict(fields.String(), fields.Model(RestrictedPCIDevice))
    pci_ids = fields.Dict(fields.String(), fields.Model(RestrictedPCIDevice))
```

### DIFF
--- a/leapp/models/fields/__init__.py
+++ b/leapp/models/fields/__init__.py
@@ -478,6 +478,82 @@ class Model(Field):
         return result
 
 
+class Dict(Field):
+    """
+    Represents a dict with keys and values of a given type (TypedDict like).
+    """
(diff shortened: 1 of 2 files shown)


pirat89:
I will wait for cabal, but the approach is no-go for me as this PR is not validating kez values. Which is the point, why dict has not been implemented.

pirat89:
Just adding info, that it was decided by team on cabal that we do not want the Dict field the leapp. Closing the PR.

PR #673, opened 2020-10-27
Add test of suppress_deprecation with pytest.fixture

pirat89:
Currently the suppress_deprecation decorator doesn't work well
on Python2 in combination with pytest.fixture. This is simple test
to show the issue.


pirat89:
It is covered by #674

PR #674, opened 2020-10-27
Replace functools.wraps with six.wraps

vinzenz:
Previously we have been using functools.wraps. However if any decorator
that uses functools.wraps is used, some code that relies on __wrapped__
being present, e.g. pytest, will fail on Python2. Switching to six.wraps
will solve this problem, as it introduces this change as well and solves
the missing __wrapped__ attribute.

### DIFF
--- a/leapp/utils/clicmd.py
+++ b/leapp/utils/clicmd.py
@@ -4,6 +4,8 @@ import sys
 
 from argparse import ArgumentParser, _SubParsersAction, RawDescriptionHelpFormatter
 
+import six
+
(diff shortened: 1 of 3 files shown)


pirat89:
Adding reference for the orig issue discussion on pytest upstream https://github.com/pytest-dev/pytest/issues/2782

Fix seems good. I will wait for test results.

pirat89 (dismissed):
There is currently big backlog for jobs on jenkins, so let's wait. But our tests passed, so approving.

pirat89 (approved):

PR #675, opened 2020-11-02
Feature/add possibility to overwrite the name of venv

ZhukovGreen:
Ref https://github.com/oamg/leapp-repository/pull/613

Adds the same functionality as we did at https://github.com/oamg/leapp-repository/pull/613. However, also adds a possibility to overwrite the python executable by the PYTHON_VENV variable (because otherwise, the feature doesn't make sense).

I don't think the naming PYTHON_VENV is good for referring to python executable, however, this is how it is called inside the https://github.com/oamg/leapp-repository/blob/master/Makefile#L28, so I keep the same naming here.

This feature allows overwriting the venv name by passing the
VENVNAME environment variable.

If the variable is not provided then tut will be used as a default value.

With this feature, you can create multiple python envs for easy testing,
i.e.

PYTHON_VENV=python3 VENVNAME=.venv_py3 make install-test
PYTHON_VENV=python2 VENVNAME=.venv_py2 make install-test

this will create python2 and python3 environments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PYTHON_VENV ?= python
-
+VENVNAME ?= tut
 CONFDIR=${DESTDIR}/etc/leapp
 LIBDIR=${DESTDIR}/var/lib/leapp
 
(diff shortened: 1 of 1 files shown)


ZhukovGreen:
@oamg/developers, please review. Thanks

vinzenz (approved):

PR #676, opened 2020-11-03
Draft: Add profiling to leapp and snactor

drehak:
Picking up where #482 left off. For science, profit and fun.

`LEAPP_CPROFILE=1` turns on the profiler and shows stats after a _successful_ execution. Maybe an option would be better instead...

### DIFF
--- a/leapp/cli/__init__.py
+++ b/leapp/cli/__init__.py
@@ -1,5 +1,9 @@
 import os
 import socket
+# profiling
+import cProfile
+import pstats
+import six
 
 from leapp.utils.clicmd import command, command_opt
 from leapp.cli import upgrade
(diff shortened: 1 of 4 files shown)


ZhukovGreen:
FYI https://github.com/benfred/py-spy

leapp/cli/__init__.py
```diff
@@ -12,7 +16,19 @@ def cli(args): # noqa; pylint: disable=unused-argument
 
 
 def main():
+    os.environ['LEAPP_CPROFILE'] = os.environ.get('LEAPP_CPROFILE', '0')
```
vinzenz, line 19: what is this about? This doesn't seem to make any sense to me

leapp/cli/__init__.py
```diff
@@ -12,7 +16,19 @@ def cli(args): # noqa; pylint: disable=unused-argument
...
+    profile_enabled = os.environ['LEAPP_CPROFILE'] == '1'
```
vinzenz, line 20:
```suggestion
    profile_enabled = os.environ.get('LEAPP_CPROFILE', '0') == '1'
```

leapp/snactor/__init__.py
```diff
@@ -73,6 +77,18 @@ def cli(args):
 
 
 def main():
+    os.environ['LEAPP_CPROFILE'] = os.environ.get('LEAPP_CPROFILE', '0')
+    profile_enabled = os.environ['LEAPP_CPROFILE'] == '1'
```
vinzenz, line 81:
```suggestion
    profile_enabled = os.environ.get('LEAPP_CPROFILE', '0') == '1'
```

ZhukovGreen (commented):
How do you see the usage of this tool? How actually the profiling process will look like? Could you provide a snippet, what the developer will see?

pirat89:
Closing. so far we do not see it valuable right now - and if we find, we can start to work on it again.

PR #680, opened 2020-11-27
snactor.sanity-check: Add option to ignore models

vinzenz:
Previously, if a Model was being consumed by actor before it was
produced,  `snactor sanity-check` reported a failure. This however
blocks some use case in leapp-repository now, where there is the
possibility to supply user data and it is being verified before in a
later stage it actually is used. This sanitization actor causes this
check to fail and would force an ugly workaround by having a dummy actor
pretending to produce the message earlier.

This patch will introduce two ways of handling this situation by
ignoring the models. Either each model can be specified as a commandline
option --ignore or a path to a file containing a list of such models can
be supplied via --ignorefile.
Any of those specified model names supplied via --ignore or the
ignore file will no longer be considered a failure.

### DIFF
--- a/leapp/snactor/commands/workflow/sanity_check.py
+++ b/leapp/snactor/commands/workflow/sanity_check.py
@@ -1,19 +1,23 @@
 from __future__ import print_function
+
 import sys
 
-from leapp.exceptions import LeappError, CommandError
+from leapp.exceptions import CommandError, LeappError
 from leapp.logger import configure_logger
 from leapp.repository.scan import find_and_scan_repositories
 from leapp.snactor.commands.workflow import workflow
-from leapp.utils.clicmd import command_arg
+from leapp.utils.clicmd import command_arg, command_opt
 from leapp.utils.output import Color
-from leapp.utils.repository import requires_repository, find_repository_basedir
+from leapp.utils.repository import find_repository_basedir, requires_repository
 
 _DESCRIPTION = 'The following messages are attempted to be consumed before they are produced: {}'
 _LONG_DESCRIPTION = '''
 Perform workflow sanity checks
 
 - check whether there is a message in the given workflow which is attempted to be consumed before it was produced
+  --ignore <MODEL> can be used to specify one or multiple times to ignore the above condition
+  --ignorefile <PATH to ignore file> can be used to specify a file to load a list of ignored models. The list shall be
+    new line separated. Lines beginning with # are considered comments.
 
 For more information please consider reading the documentation at:
 https://red.ht/leapp-docs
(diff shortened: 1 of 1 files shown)


vinzenz:
@pirat89 the review request is more of a FYI at this point.

pirat89:
Seems it should do what it is supposed to do. I will test it manually soon. I am even thinking about the --ignorefile - what abuot to have a default path inside the repo, e.g. `.ignoresanity` which in case it exists, it is loaded automatically? Just thinking about the way we could use without the need of changes in CI and having possibility to specify it automatically. If any such default is set, we should probably provide a --noignore or --noignorefile option, to have a possibility to shutdown the check. @vinzenz wdyt about that?

pirat89:
we should finish this stuff :)

pirat89:
:eye:

leapp/snactor/commands/workflow/sanity_check.py
```diff
@@ -24,10 +28,19 @@ def print_fail(error):
     print('{red}FAIL: {error}{reset}'.format(red=Color.red, error=error, reset=Color.reset), file=sys.stderr, end='\n')
 
 
+def _ignore_file_content(path):
+    if path:
+        with open(path) as f:
+            return [line.strip() for line in f.read().split('\n') if not line.strip().startswith('#')]
```
pirat89, line 34: @vinzenz one small fix to prevent crashes when no ignorefile is specified:
```suggestion
    if not path:
        return []
    with open(path) as f:
        return [line.strip() for line in f.read().split('\n') if not line.strip().startswith('#')]
```

before:
```
  File "/repocopy/tut/bin/snactor", line 8, in <module>                            
    sys.exit(main())                                                               
  File "/repocopy/tut/lib/python3.6/site-packages/leapp/snactor/__init__.py", line 78, in main
  File "/repocopy/tut/lib/python3.6/site-packages/leapp/utils/clicmd.py", line 111, in execute
    args.func(args)                                                                
  File "/repocopy/tut/lib/python3.6/site-packages/leapp/utils/clicmd.py", line 133, in called
    self.target(args)                                                              
  File "/repocopy/tut/lib/python3.6/site-packages/leapp/utils/repository.py", line 20, in checker
    return f(*args, **kwargs)                                                      
  File "/repocopy/tut/lib/python3.6/site-packages/leapp/snactor/commands/workflow/sanity_check.py", line 43, in cli 
    ignored_models = set(params.ignore or []) | set(_ignore_file_content(params.ignorefile))

```

pirat89 (changes requested):

pirat89 (approved):

pirat89:
@fernflower it seems that CI tests are invalid. - tests fail mostly on repomapping.json - informing it's invalid. however tested manually, everything is ok. we will need to investigate that later to see what's the issue. posponing it for now as we are running out of time.

PR #681, opened 2021-01-13
Introduced a JSON schema of a leapp report

fernflower:
Generated by https://jsonschema.net and adjusted manually.

OAMG-4263

### DIFF
--- a/report-schema-v100.json
+++ b/report-schema-v100.json
@@ -1,0 +1,648 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "http://example.com/example.json",
+    "type": "object",
+    "title": "The (pre)upgrade report schema",
+    "description": "The root schema comprises the entire JSON document.",
+    "default": {},
+    "version": "1.0.0",
+    "examples": [
+        {
+            "leapp_run_id": "a91dccab-84e8-4a28-b28e-102b073200a9",
+            "entries": [
+                {
+                    "hostname": "vagrant-leapp-20210113121507",
+                    "severity": "info",
+                    "tags": [
+                        "repository"
+                    ],
+                    "timeStamp": "2021-01-13T13:04:06.895402Z",
+                    "title": "Excluded RHEL 8 repositories",
+                    "detail": {
+                        "external": [
+                            {
+                                "url": "https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/package_manifest/codereadylinuxbuilder-repository.",
+                                "title": "CodeReady Linux Builder repository"
+                            }
+                        ]
+                    },
+                    "actor": "repositories_blacklist",
+                    "summary": "The following repositories are not supported by Red Hat and are excluded from the list of repositories used during the upgrade.\n- codeready-builder-for-rhel-8-x86_64-rpms",
+                    "audience": "sysadmin",
+                    "flags": [
+                        "failure"
+                    ],
+                    "id": "88e518df66f34e0097ef7b4a4121f50eb5bcf9e8b2c017d1d07dca5e22db65e5"
+                },
+                {
+                    "hostname": "vagrant-leapp-20210113121507",
+                    "severity": "high",
+                    "title": "Packages not signed by Red Hat found on the system",
+                    "timeStamp": "2021-01-13T13:04:20.582335Z",
+                    "tags": [
+                        "sanity"
+                    ],
+                    "actor": "red_hat_signed_rpm_check",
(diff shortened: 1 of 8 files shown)


pirat89:
@fernflower  We should refer to the schema from the upstream documentation.

fernflower:
@pirat89 I will put the reference once the schema is merged. Are you ok with the placing in leapp's root?

pirat89:
@fernflower I am thinking where to put it as we want to reduce the main menu. We should discuss it in future with @mportman12  probably. But temporary it could be ok. Btw, why to do it separately? Why not in this PR? As these two things are two sides of the same coin. I guess it's ok unless the jira task is kept in development or if you create the followup task, so we do not foret to finish it.

pirat89:
@fernflower aah...I thought you are talking about readthedocs menu. Now I see where the file is located :) I think it should be in a subdir. As well, haven't we talked about version schema? As we plan to change it in future, so I believe there should be something like version 1.0 .. maybe the file should reflect that as well and should be called `report-schema-v1.0.json`. wdyt?
I am not sure which directory to use to put the schema inside. @vinzenz any idea?

pirat89:
leapp-ci build

fernflower:
@pirat89 what about putting 'version' in the schema itself (did it in the latest commit)? I'd rather keep the filename as it is `report-schema.json` tbh, if we follow the 'do not remove any fields from json' concept the latest file should be supported anyway.

pirat89:
@fernflower  And how you will point people to various versions of schemas then? we will need keep several versions of schemas documented in the same time..

pirat89:
I really believe we need to have info about the version inside the schema and reflect that in the filename as well.

pirat89 (approved):
giving my approval to not block the pr. I will try to think where to move the file. in the worst case, we can still move it after the PR merge.

pirat89:
replaced tabs by whitespaces in the fixup commits

report-schema-v100.json
```diff
@@ -0,0 +1,600 @@
...
+                                                        }
+                                                    ],
+                                                    "required": [
+                                                        "url",
+                                                        "title"
+                                                    ],
+                                                    "properties": {
+                                                        "url": {
+                                                            "$id": "#/properties/entries/items/anyOf/0/properties/detail/properties/external/items/anyOf/0/properties/url",
+                                                            "type": "string",
```
pirat89, line 326: What about to add the `uri` format to make validation even more strict?"

```suggestion
                                                            "type": "string",
                                                            "format": "uri",
```

ref: https://json-schema.org/understanding-json-schema/reference/string.html#resource-identifiers

report-schema-v100.json
```diff
@@ -0,0 +1,600 @@
...
+                                                           "type": "hint",
+                                                           "context": "Please update your scripts to be compatible with the changes."
+                                                       }
+                                                   ],
+                                                   "required": [
+                                                       "type",
+                                                       "context"
+                                                   ],
+                                                   "properties": {
+                               "anyOf": [
```
pirat89, line 446: properties cannot contain "anyOf". this error is present in both files

pirat89:
Fixing the indentation as it was broken on several places.

report-schema-v100.json
```diff
@@ -0,0 +1,600 @@
...
+                                                                    "$id": "#/properties/detail/properties/remediations/items/anyOf/0/properties/context",
+                                                                    "type": "array",
+                                                                    "title": "The context schema",
+                                                                    "description": "List of remediation commands.",
+                                                                    "default": [],
+                                                                    "examples": [
+                                                                        [
+                                                                            "rm -rf /tmp/42"
+                                                                        ]
+                                                                    ],
```
pirat89, line 492: This is wrong actually. E.g. 
```
{
  "type": "command",
  "context": [
    "leapp", "answer", "--section", "remove_pam_pkcs11_module_check.confirm=True"
  ]
}

```

We should add more examples to be able to validate the json schema better. Let's do it the next morning.

report-schema-v100.json
```diff
@@ -0,0 +1,600 @@
...
+                                                    ],
+                                                    "required": [
+                                                        "type",
+                                                        "context"
+                                                    ],
+                                                    "properties": {
+                                                        "anyOf": [
+                                                            {
+                                                                "type": {
+                                                                    "$id": "#/properties/detail/properties/remediations/items/anyOf/0/properties/type",
```
pirat89, line 449: and this is probably not possible as well. properties should be always key-value dicts as it looks like.

pirat89 (dismissed):

pirat89 (approved):
:partying_face:

PR #684, opened 2021-02-02
Bump framework_version to 1.4

pirat89:
Regarding the new functionality (reporting.Key) providing the stable
key id for every report, the framework version should be bumped to
keep stuff compatible.


vinzenz (approved):

pirat89:
We do not need to run additional level of tests for this change. Merging

PR #686, opened 2021-02-09
Control leapp-report format

fernflower:
In order to be completely safe that any changes that land in leapp
reporting don't break older clients there should be a possibility
to force specific version of json schema.
Currently there are 2 schemas that are actually backward compatible,
but it may not be the rule soon.

This patch adds a --report-schema option to leapp preupgrade/upgrade
commands that force sticking to specific version of the report schema.
For example with --report-schema '1.0.0' passed there will be no stable
keys in the report messages, as those have been added in the recent
'1.1.0' report schema.

### DIFF
--- a/leapp/utils/report.py
+++ b/leapp/utils/report.py
@@ -89,7 +89,9 @@ def importance(message):
     return SEVERITY_LEVELS.get(message['severity'], 99)
 
 
-def generate_report_file(messages_to_report, context, path):
+def generate_report_file(messages_to_report, context, path, report_schema='1.1.0'):
+    # NOTE(ivasilev) Int conversion should not break as only specific formats of report_schema versions are allowed
+    report_schema_tuple = tuple([int(x) for x in report_schema.split('.')])
     if path.endswith(".txt"):
         with open(path, 'w') as f:
             for message in sorted(messages_to_report, key=importance):
@@ -100,8 +102,17 @@ def generate_report_file(messages_to_report, context, path):
                 remediation = Remediation.from_dict(message.get('detail', {}))
                 if remediation:
                     f.write('Remediation: {}\n'.format(remediation))
-                f.write('Key: {}\n'.format(message['key']))
+                if report_schema_tuple > (1, 0, 0):
+                    # report-schema 1.0.0 doesn't have a stable report key
+                    f.write('Key: {}\n'.format(message['key']))
                 f.write('-' * 40 + '\n')
     elif path.endswith(".json"):
         with open(path, 'w') as f:
+            # Here all possible convertions will take place
+            if report_schema_tuple < (1, 1, 0):
+                # report-schema 1.0.0 doesn't have a stable report key
+                # copy list of messages here not to mess the initial structure for possible further invocations
+                messages_to_report = list(messages_to_report)
+                for m in messages_to_report:
+                    m.pop('key')
             json.dump({'entries': messages_to_report, 'leapp_run_id': context}, f, indent=2)
--- a/man/leapp.1
+++ b/man/leapp.1
@@ -55,6 +55,9 @@ leapp \- command line interface for upgrades between major OS versions
     Enable an experimental actor. Can be used multiple times.
     To use this variable, \fBLEAPP_UNSUPPORTED\fR has to be set. See \fBDeveloper variables\fR for more information.
 
+\fB--report-schema\fR <\fIversion\fR>
+    Force the report format to conform to a specific version of the report schema. Expected values: \fB1.0.0\fR, \fB1.1.0\fR. Defaults to \fB1.1.0\fR.
+
 \fB--reboot\fR
     (upgrade only) Automatically perform reboot when requested.
 
--- a/tests/scripts/test_reporting.py
+++ b/tests/scripts/test_reporting.py
@@ -2,6 +2,7 @@ from collections import namedtuple
 import datetime
 import json
 import os
+import tempfile
 
 import jsonschema
 import pytest
@@ -14,6 +15,7 @@ from leapp.reporting import (
     _create_report_object,
     Audience, Key, Title, Summary, Severity, RelatedResource
 )
+from leapp.utils.report import generate_report_file
 
 
 ReportPrimitive = namedtuple('ReportPrimitive', ['data', 'path', 'value', 'is_leaf_list'])
@@ -220,3 +222,33 @@ def test_report_jsonschema():
             # schemas directory contain a symlink to report json-schemas
             schema_data = _load_json(os.path.join(reports_dir, 'schemas', schema))
             jsonschema.validate(report_data, schema_data)
+
+
+def test_report_backwards_compatibility():
+    msg1 = {u'groups': [u'selinux', u'security'],
+            u'title': u'SElinux relabeling has been scheduled',
+            'timeStamp': u'2021-02-10T13:13:40.576437Z',
+            'hostname': u'ivasilev-tagsflagsgroups',
+            'actor': u'check_se_linux',
+            u'summary': u'SElinux relabeling has been scheduled as the status was permissive/enforcing.',
+            u'audience': u'sysadmin',
+            u'key': u'c12a05a22be0b5bc0af3f1119898ea6d8639d9c4',
+            'id': '29279fc2e3100cfdf8ac30f48b293185192672d5a5ae9e0cfbc76cbfc919d807',
+            u'severity': u'info'}
+    report = [msg1]
+    # make sure normal mode works
+    for report_format in ['.json', '.txt']:
+        reportfile = tempfile.NamedTemporaryFile(suffix=report_format)
+        generate_report_file(report, 'leapp-run-id', reportfile.name, '1.1.0')
+    # make sure report output conversion to specific version works as well
+    for report_format in ['.json', '.txt']:
+        reportfile = tempfile.NamedTemporaryFile(suffix=report_format)
+        generate_report_file(report, 'leapp-run-id', reportfile.name, '1.0.0')
+        with open(reportfile.name) as f:
+            data = f.read()
+        if report_format == '.json':
+            a_report = json.loads(data)
+            msg = a_report['entries'][0]
+            assert 'key' not in msg
+        else:
+            assert 'Key' not in data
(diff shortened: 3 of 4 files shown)


pirat89:
@fernflower  no review yet; just I think that the argument for the `--report-schema` option should be enumarated list of values. If in example we provide schema e.g. 1.5.0 in upstream, but customer will use older version of leapp (e.g. because the new one was not delivered yet), obviously such a schema will not be valid. from that point, leapp should check whether the required schema is supported or not. as well, it could happen we will have to drop completely some old schemas, because it will be completely incompatible with the new one (not searching for possible reasons, it's just possibility that could happen in future). so the enumeration/list of supported schemas should be rather present. wdyt?

fernflower:
> @fernflower no review yet; just I think that the argument for the `--report-schema` option should be enumarated list of values. If in example we provide schema e.g. 1.5.0 in upstream, but customer will use older version of leapp (e.g. because the new one was not delivered yet), obviously such a schema will not be valid. from that point, leapp should check whether the required schema is supported or not. as well, it could happen we will have to drop completely some old schemas, because it will be completely incompatible with the new one (not searching for possible reasons, it's just possibility that could happen in future). so the enumeration/list of supported schemas should be rather present. wdyt?

Hmm why a list of values? My idea was that --report-schema version will mandatory force leapp to create a leapp-report.json compatible with selected version of report schema (1.0.0 without "key" and 1.1.0 with "key" for example). 
If the client uses super old version of leapp he won't be able to use --report-schema cli option anyway. If the version is not old but doesn't match the one in upstream (say, 1.4.0 for client and 1.5.0 for upstream) then only convertions to earlier than 1.4.0 versions are available.

fernflower:
leapp-ci build

fernflower:
@oamg/developers please review

pirat89:
Passing  the `choices` to argparse (as you did already) was the answer :)

pirat89:
I think this could be very good candidate for our ci-hackathon.

leapp/cli/upgrade/__init__.py
```diff
@@ -201,6 +202,14 @@ def upgrade(args):
     configuration = prepare_configuration(args)
     answerfile_path = cfg.get('report', 'answerfile')
     userchoices_path = cfg.get('report', 'userchoices')
+    # report_schema None will mean that no special backwards-compatibility conversion is necessary
+    report_schema = None
+    if args.report_schema:
+        if args.report_schema < cfg.get('report', 'schema'):
+            report_schema = args.report_schema
+        else:
+            raise CommandError('--report-schema version can not be greater that the '
```
drehak, line 211:
```suggestion
            raise CommandError('--report-schema version can not be greater than the '
```
typo

leapp/cli/upgrade/__init__.py
```diff
@@ -201,6 +202,14 @@ def upgrade(args):
     configuration = prepare_configuration(args)
     answerfile_path = cfg.get('report', 'answerfile')
     userchoices_path = cfg.get('report', 'userchoices')
+    # report_schema None will mean that no special backwards-compatibility conversion is necessary
+    report_schema = None
+    if args.report_schema:
+        if args.report_schema < cfg.get('report', 'schema'):
```
drehak, line 208: When I pass an argument with schema version equal to the actual one, leapp throws an error:

```
$ leapp preupgrade --report-schema 1.1.0

Error: --report-schema version can not be greater that the actual 1.1.0 one.
```

Is the error intended? The error message says that the provided schema version cannot be greater than the actual one, but that isn't the case here.

drehak (changes requested):

drehak (dismissed):

drehak:
Seens good to me!

leapp/cli/upgrade/__init__.py
```diff
@@ -183,6 +183,18 @@ def process_whitelist_experimental(repositories, workflow, configuration, logger
...
+def process_report_schema(args, cfg):
+    report_schema = None
+    default_report_schema = cfg.get('report', 'schema')
+    if args.report_schema and args.report_schema != default_report_schema:
+        if args.report_schema < default_report_schema:
+            report_schema = args.report_schema
+        else:
+            raise CommandError('--report-schema version can not be greater that the '
+                               'actual {} one.'.format(default_report_schema))
+    return report_schema
```
vinzenz, line 195: Just to be sure, is it intended, by default to return `None` if args.report_schema isn't set or equal to the default_report_schema?

fernflower: Yeah, it was a shortcut to not doing any checks later if report messages processing needs to be done at all (otherwise you won't differentiate between 'convert to 1.1.0' and 'don't do anything, because 1.1.0 is the latest version anyway' without comparing the passed version to the actual from the config file).
But now I think I know a way to improve it, thank you!

tests/scripts/test_reporting.py
```diff
@@ -220,3 +222,33 @@ def _load_json(a_file):
...
+            u'summary': u'SElinux relabeling has been scheduled as the status was permissive/enforcing.',
+            u'audience': u'sysadmin',
+            u'key': u'c12a05a22be0b5bc0af3f1119898ea6d8639d9c4',
+            'id': '29279fc2e3100cfdf8ac30f48b293185192672d5a5ae9e0cfbc76cbfc919d807',
+            u'severity': u'info'}
+    report = [msg1]
+    # make sure normal mode works
+    for report_format in ['.json', '.txt']:
+        reportfile = tempfile.NamedTemporaryFile(suffix=report_format)
+        generate_report_file(report, 'leapp-run-id', reportfile.name, '1.1.00')
```
drehak, line 242: I'm a bit confused about the `1.1.00`... why is it there?

also feel free to squash my commit, I don't mind :)

fernflower: Looks like I need to call it a day and finally clean the keyboard. Thanks, will fix.
As for the squashing - I'd rather keep 2 separate commits :)

drehak (dismissed):
Looks good to me!

leapp/cli/upgrade/__init__.py
pirat89: Thinking about that, what about to add the `default` value argument for `add_option`?

fernflower:
@pirat89 love the idea, got rid of the separate report schema processing. Thanks!

leapp/utils/report.py
```diff
@@ -100,8 +100,17 @@ def generate_report_file(messages_to_report, context, path):
                 remediation = Remediation.from_dict(message.get('detail', {}))
                 if remediation:
                     f.write('Remediation: {}\n'.format(remediation))
-                f.write('Key: {}\n'.format(message['key']))
+                if not report_schema or report_schema > '1.0.0':
```
pirat89, line 103: @fernflower  wdyt about transform it to tuple? `(1, 0, 0)`? Just thinking about that. This will be ok so far we have just numbers up-to-9. however in case we will have a version "1.10.0", such comparison will not be the good way. It's not problem right now, this should work fine so far. But thinking whether wouldn't be better to parse it to a better data format already to not have to think about that in future.

fernflower: Perfect, will do

pirat89: @fernflower  Thanks. Looks good to me. Btw, in case you would like to do it using the callback functions for the `action` argument, the solution could looks something like that:

```python
#!/usr/bin/python3

import argparse

class VersionParserAction(argparse.Action):
    def __init__(self, option_strings, dest, **kwargs):
        super(VersionParserAction, self).__init__(option_strings, dest, **kwargs)
    def __call__(self, parser, namespace, version_string, option_string=None):
        parsed_version = tuple([int(i) for i in version_string.split('.')])
        setattr(namespace, self.dest, parsed_version)

parser = argparse.ArgumentParser()
parser.add_argument(
    '--report-schema', help='Specify report schema version for leapp-report.json',
    choices=['1.0.0', '1.1.0'], default=(1, 1, 0),
    action=VersionParserAction
)

### for tests tests
print(parser.parse_args('--report-schema 1.1.0'.split()))
print(parser.parse_args('--report-schema 1.0.0'.split()))

args = parser.parse_args('--report-schema 1.1.0'.split())
print(args.report_schema)
args = parser.parse_args(''.split())
print(args.report_schema)
```

fernflower: tbh I prefer my version and tuple conversion before actual usage.

man/leapp.1
```diff
@@ -55,6 +55,9 @@ leapp \- command line interface for upgrades between major OS versions
     Enable an experimental actor. Can be used multiple times.
     To use this variable, \fBLEAPP_UNSUPPORTED\fR has to be set. See \fBDeveloper variables\fR for more information.
 
+\fB--report-schema\fR <\fIversion\fR>
+    Force report format to conform to a specific version of the report schema. Expected values: \fB1.0.0\fR, \fB1.1.0\fR. Defaults to \fB1.1.0\fR.
```
pirat89, line 59:
```suggestion
    Force the report format to conform to a specific version of the report schema. Expected values: \fB1.0.0\fR, \fB1.1.0\fR. Defaults to \fB1.1.0\fR.
```

leapp/cli/upgrade/__init__.py
```diff
@@ -261,6 +262,8 @@ def upgrade(args):
                                            ' with Red Hat Subscription Manager')
 @command_opt('enablerepo', action='append', metavar='<repoid>',
              help='Enable specified repository. Can be used multiple times.')
+@command_opt('report-schema', help='Specify report schema version for leapp-report.json', choices=['1.0.0', '1.1.0'],
```
pirat89, line 265:
```suggestion
@command_opt('report-schema', help='Specify the report schema version that should for leapp-report.json', choices=['1.0.0', '1.1.0'],
```

pirat89 (dismissed):
Lgtm and works as expected. Just some english nitpicks and I think the help message is not so clear. But thiking about that, I am not sure how to re-word it to keep it short and clear.

leapp/utils/report.py
pirat89: no problem.

pirat89:
@fernflower is it still WIP? If not, I would like to merge it after the next release.

pirat89:
Something seems to be pretty broken.

fernflower:
@pirat89 Should be fixed

pirat89 (approved):
works as expected and lgtm.

PR #688, opened 2021-02-10
Doc: add lists of the deprecated functionalites

pirat89:
Just document the stuff that is already deprecated. Added:
- a new document for the deprecated stuff in the el7toel8
  leapp repository
- a new section about deprecated stuff in leapp itself

The main deprecation document is pretty long now and the section
for the deprecated list in leapp is not good there. However I do not
see better way how to do it right now. The doc needs to be split
for sure in future, but this is not scope of this PR.

### DIFF
--- a/docs/source/deprecation.md
+++ b/docs/source/deprecation.md
@@ -6,6 +6,15 @@ is evolving, it happens that some functionality needs to be changed, replaced
 or dropped completely. Such situations are inevitable. To reduce negative
 impact on your code, we introduce the deprecation process described below.
 
+## List of the deprecated functionality in leapp
+
+The following lists cover deprecated functionality in the leapp utility, snactor,
+the leapp standard library, etc. But don't cover deprecated functionalities
+from particular leapp repositories (e.g. the [elt7toel8](https://github.com/oamg/leapp-repository/tree/master/repos/system_upgrade/el7toel8) leapp repository). For
+such information, see [Deprecated functionality in the el7toel8 repository](el7toel8/deprecation.html#deprecated-functionality-in-the-el7toel8-repository).
+
+*Note: No new deprecation present yet.*
+
 ## What is covered by deprecation process in leapp?
 
 In short, leapp entities that are supposed to be used by other developers.
--- a/docs/source/el7toel8/deprecation.md
+++ b/docs/source/el7toel8/deprecation.md
@@ -1,0 +1,28 @@
+# Deprecated functionality in the el7toel8 repository
+
+Deprecated functionality is listed under the first version that the functionality
+is deprecated in. Note that functionality may be deprecated in later versions
+but are not listed again.
+The dates in brackets correspond to the end of the deprecation protection period,
+after which the related functionality can be removed at any time.
+
+*Note* The lists cover just the functionality provided inside the el7toel8
+repository only. For the functionality deprecated in the leapp
+framework, see [List of deprecated functionality in leapp](../deprecation.html#list-of-deprecated-functionality-in-leapp)
+
+## current upstream development
+
+*Note: No new deprecation present yet.*
+
+## v0.12.0  <span style="font-size:0.5em; font-weight:normal">(till April  2021)</span>
+- Models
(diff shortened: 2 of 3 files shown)


docs/source/deprecation.md
```diff
@@ -6,6 +6,13 @@ is evolving, it happens that some functionality needs to be changed, replaced
 or dropped completely. Such situations are inevitable. To reduce negative
 impact on your code, we introduce the deprecation process described below.
 
+## List of the deprecated functionality in leapp
+
+The following lists do not cover the functionality provided by particular leapp
```
mportman12, line 11: Not clear what "particular leapp repositories" means.

docs/source/deprecation.md
```diff
@@ -6,6 +6,13 @@ is evolving, it happens that some functionality needs to be changed, replaced
...
+repositories. To see such a list, see [Deprecated functionality in the el7toel8 repository](el7toel8/deprecation.html#deprecated-functionality-in-the-el7toel8-repository)
```
mportman12, line 12: Add period at end of sentence.

docs/source/el7toel8/deprecation.md
```diff
@@ -0,0 +1,27 @@
+# Deprecated functionality in the el7toel8 repository
+
+Here is the list of deprecated functionality in the el7toel8 repository. Each
+version section contains just the newly introduced deprecation. The dates
+in brackets correspond to the end of the deprecation protection period,
+after which the related functionality can be whenever removed.
```
mportman12, line 6: "after which the related functionality can be removed at any time."

docs/source/el7toel8/deprecation.md
```diff
@@ -0,0 +1,27 @@
+# Deprecated functionality in the el7toel8 repository
+
+Here is the list of deprecated functionality in the el7toel8 repository. Each
+version section contains just the newly introduced deprecation. The dates
```
mportman12, line 4: Add info about when a deprecated feature will be removed from this list. How long is "newly introduced"?

mportman12 (changes requested):
Left a few comments. Otherwise looks good.

docs/source/deprecation.md
pirat89: Not sure how to write it. For example, `el7toel8` is a leapp repository. Soon we will add `el8toel9` leapp repository. These are the repositories containing leapp actors, leapp workflows, ... that leapp is executing. (in case of the official repositories, we are speaking about leapp repositories inside the *leapp-repository* git repository (https://github.com/oamg/leapp-repository).

docs/source/deprecation.md
pirat89: Done.

docs/source/el7toel8/deprecation.md
pirat89: Done.

docs/source/el7toel8/deprecation.md
pirat89: @mportman12  I have used probably wrong wording regarding the chenglish. I mean when a new deprecation is introduced e.g. in v 0.11.0, and it still present (still deprecated) in additional releases (e.g. 0.12.0), it is not listed again. like in case `TMPTargetRepositoriesFacts`

docs/source/deprecation.md
fernflower: Miriam is right here, it is very confusing (even for me). Maybe instead of saying what the lists don't cover we should fix what is actually covered in this section (like framework stuff being deprecated etc)? This won't solve the general "leapp repositories" confusion, but will at least eliminate it here.

pirat89: Synced with Miriam in person. The paragraph is now reworded @fernflower

docs/source/el7toel8/deprecation.md
mportman12: “Deprecated functionality is listed under the first version that the functionality is deprecated in. Note that functionality may be deprecated in later versions but are not listed again.”

pirat89: Updated.

pirat89:
review please

mportman12 (approved):

fernflower (approved):
lgtm

PR #695, opened 2021-04-15
Tests: Use centos repository for docker images

vinzenz:
Due to the docker hub rate limitation we will switch to be using the
registry.centos.org registry for our tests.


Rezney (approved):
Looks good to me...

PR #697, opened 2021-04-16
upgrade: CLI stops logging to log file after resume

vinzenz:
Previously when the upgrade was resumed, leapp did not continue to write
to the leapp-upgrade.log.
This patch fixes the issue.

Bug description:

`configure_logger()` was called after resume without a parameter. Which
caused the log_file parameter of configure_logger to be None. If log_file
is None, the `FileHandler` instance wasn't created and in consequence
not added the internal logger. Which then of course never writes to the
leapp-upgrade.log


pirat89 (approved):

pirat89:
leapp-ci build

PR #698, opened 2021-04-16
testing: Adding second place for centos registry

vinzenz:
During the previous fix, I overlooked the git pull of the container in
the Makefile. This caused a failure at that point. This PR adds the
second location where this is happening.


Rezney (approved):
Ack

PR #700, opened 2021-04-20
cli: Automatic command discovery for leapp CLI

vinzenz:
This patch introduces a new approach in detecting the leapp cli commands.
The main function will now look into the directory of leapp.cli.commands
and list all directories within that directory and attempts to import a
`register` function from them. If that succeeds it will call this function
passing the toplevel command object to it allowing the function to register
the commands and if necessary sub commands to it.

This patch will also drop leapp/cli/upgrade and reintroduce it to
leapp-repository in a different PR.

Framework version will now be 2.0


leapp/cli/__init__.py
```diff
@@ -1,18 +1,40 @@
...
+from leapp.utils.repository import find_repos, get_repository_name
 
 
 @command('')
-def cli(args): # noqa; pylint: disable=unused-argument
-    pass
+def cli(args):  # noqa; pylint: disable=unused-argument
+    """
+        Top level base command dummy function
+    """
```
pirat89, line 18: Indentation
```suggestion
    """
    Top level base command dummy function
    """
```

pirat89:
@vinzenz btw do not forget to raise the `framework-version` capability. From this point, it seems to me we will need to raise it to 2.0 as this is not a change compatible with previous versions of leapp. (and update of spec is expected on the leapp-repository side as well)

pirat89:
(review in progress, I will finish it the next morning)

vinzenz:
> @vinzenz btw do not forget to raise the `framework-version` capability. From this point, it seems to me we will need to raise it to 2.0 as this is not a change compatible with previous versions of leapp. (and update of spec is expected on the leapp-repository side as well)

Done

pirat89:
See the comments in https://github.com/oamg/leapp-repository/pull/662 PR. Basically tests / environemnt preparation... are broken. It's not possible to run e2e tests correctly for this PR and integration tests are broken. Merging.

PR #702, opened 2021-05-03
Add GitHub workflow that triggers tmt tests in Testing Farm service

vojtechsokol:
The actual command that triggers a new test on TF service is commented out as
it would fail at this moment, because there are no tmt plans ready in the
repository.

Depends on: #704

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,9 +98,9 @@ srpm: source
 
 copr_build: srpm
 	@echo "--- Build RPM ${PKGNAME}-${VERSION}-${RELEASE}.el7.rpm in COPR ---"
-	@echo copr --config $(_COPR_CONFIG) build $(_COPR_REPO) \
+	@echo copr-cli --config $(_COPR_CONFIG) build $(_COPR_REPO) \
 		packaging/SRPMS/${PKGNAME}-${VERSION}-${RELEASE}*.src.rpm
-	@copr --config $(_COPR_CONFIG) build $(_COPR_REPO) \
+	@copr-cli --config $(_COPR_CONFIG) build $(_COPR_REPO) \
 		packaging/SRPMS/${PKGNAME}-${VERSION}-${RELEASE}*.src.rpm
 
 print_release:
--- a/.github/workflows/tmt-tests.yml
+++ b/.github/workflows/tmt-tests.yml
@@ -1,0 +1,135 @@
+name: tmt@TF
+
+on:
+  issue_comment:
+    types:
+      - created
+
+jobs:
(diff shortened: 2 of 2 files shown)


vojtechsokol:
This workflow won't be active until it's merged into master. therefore I've merged the exact same changes into master branch of my [fork](https://github.com/vojtechsokol/leapp/commit/a47965e43a7e190ce267f63678ee536dbd77c02c). The workflow reacts on `/rerun` comments in pull request by a owner, memeber or collaborator of the repo/org. You can see the results in [this PR](https://github.com/vojtechsokol/leapp/pull/3), while logs of the action's run is available [here](https://github.com/vojtechsokol/leapp/runs/2570220200).

Makefile
```diff
@@ -98,9 +98,9 @@ srpm: source
 
 copr_build: srpm
 	@echo "--- Build RPM ${PKGNAME}-${VERSION}-${RELEASE}.el7.rpm in COPR ---"
-	@echo copr --config $(_COPR_CONFIG) build $(_COPR_REPO) \
+	@echo copr-cli --config $(_COPR_CONFIG) build $(_COPR_REPO) \
 		packaging/SRPMS/${PKGNAME}-${VERSION}-${RELEASE}*.src.rpm
-	@copr --config $(_COPR_CONFIG) build $(_COPR_REPO) \
+	@copr-cli --config $(_COPR_CONFIG) build $(_COPR_REPO) \
```
pirat89, line 103: why that change? I know the copr-cli is probably the name, but copr is available as well. At least in my case on my system. Just want to understand the need for the change @vojtechsokol  @jmikovic ?

pirat89: Explained. On the testing farm the copr-cli is installed via `pip` - so it's not installed rpm as we would expect in this case (e.g. `yum install copr-cli`) -> so the symlink does not exist as the python package doesn't care about that.

pirat89: Btw, regarding the gazillion problems around `pip`, in case it start to fail on the copr-cli installation, I suggest to be prepared for installation via `yum` or `dnf`.

vojtechsokol: Yep, the issue was that `copr` command is not available when installing from pip. When installed from rpm package, there are both (and `copr` being a symlink to `copr-cli`), so it should be safer to use `copr-cli`.

vojtechsokol: And regarding using yum/dnf - the runner is ubuntu, so no system package for copr...

fernflower (approved):
Very anxious to see that in action, thanks, Vojta!

PR #704, opened 2021-05-12
Set the default shell used in Makefile to bash

vojtechsokol:
The default shell in Makefile is `sh` and since bashisms are used in the
Makefile, some of the targets fail, depending on setting of the system
(not an issue on vanilla Fedora where `/bin/sh` is a symlink to `bash`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+# there are bashisms used in this Makefile
+SHELL=/bin/bash
+
 PYTHON_VENV ?= python
 VENVNAME ?= tut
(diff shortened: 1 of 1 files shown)


vojtechsokol:
@pirat89 please take a look, this needs to be merged before #702 (part of our CI hackathon). I applied the same change also to oamg/leapp-repository#665.

pirat89 (approved):
lgtm. please, next time add info what type of bashisms is in the file. I will update the commit msg. for others: `"str/{str1,str2}/"` is bashism too.

EDIT: crap.. I have forgotten to update the commit msg...

PR #706, opened 2021-05-18
Fix incorrect request to TF service

vojtechsokol:
This patch fixes incorrect TF request and also adds print of the request/response jsons that can be enabled by setting the `ACTIONS_STEP_DEBUG` secret to `true` (this also enables step debug logging).

### DIFF
--- a/.github/workflows/tmt-tests.yml
+++ b/.github/workflows/tmt-tests.yml
@@ -105,10 +105,20 @@ jobs:
           }
           EOF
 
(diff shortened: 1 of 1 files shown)


vojtechsokol:
@fernflower I forgot to update the compose used for testing in #702, and there was also a syntax error in the request json... So I also added prints for the TF request/response json for better debugging.

fernflower (approved):

PR #708, opened 2021-05-19
Pass pr_number as env var and copr build as an artifact

fernflower:
Also add link to the full pipeline log.


vojtechsokol (approved):
LGTM

PR #714, opened 2021-06-09
Pass copr build properly

fernflower:
Artifacts to be installed should come in build_id:chroot_name form.

Ex. "artifacts": [{"type": "fedora-copr-build", "id": "2220887:epel-7-x86_64"}]


pirat89 (approved):

pirat89 (approved):
Seem good to me. Regarding the current error msgs without the PR, it seems pretty straighforward.

PR #717, opened 2021-06-15
Packaging: prepare building for RHEL 8+  and Fedora (iow, for Python3)

pirat89:
The SPEC file has been simplified to make the maintaining simple for
people. As the set of changes seems biger regarding that, here
is the list of changes explicitly:

* The building for Python2 and Python3 is now exclusive.
  Python2 is supported only on RHEL 7. For all newer systems
  (Fedora including) build just with Python3.

* Additionaly the python3-leapp packages provides leapp-framework
  as expected.

* Remove auto generated python(abi) dependency on Python3 systems
  Various versions of systems (RHEL 8, RHEL 9, Fedora 33, Fedora 34)
  use different versions of Python. This means the python(abi) is
  different on each system, however leapp packages (excluding the
  deps metapackage) has to survive the upgrade of the system
  untouched. This is not possible when requirements on python(abi)
  is defined, so drop this dependency from all packages.
  * python(abi) for Python2 is kept untouched for historical reasons
    and because python 2.7 is on both - RHEL 7 and RHEL 8 systems.

* Remove autogenerated /usr/libexec/platform-python dependency on EL8
  It's analogy to the problem above. Just in this case, this dependency
  has been generated just on EL 8 systems. However the platform-python
  will not exist anymore on newer systems.

* Additionally disable auto generation of Python3 dependencies.
  The dependencies are expected to be handled via the deps package,
  however the autogenerated deps were put into the python3-leapp rpm,
  which is not what we want (it breaks our handy solution with deps
  in case of IPU).

* Set the conflifcts between python2-leapp and python3-leapp pkgs
  We build only for Python2 of Python3, so enable installation only
  one of those two packages. As well we do not want to set provides
  and bosoletes for Python2, because packages from one system cannot
  be just updated by packages from another system.

* Enable building of leapp on Fedora
  Since the leapp tool doesn't provide any specific commits, it's
  no problem to enable it's building for the Fedora - as it doesn't
  depends on leapp-repository neither. The manpage is still confusing
  but that's going to be updated in the separate PR.

---
## old notes
* **Comment out the python abi requirement.**
  * We need to discuss this one later.
currently, with the generated python(abi) = 3.6 it's not possible to preserve
the rpm on RHEL 9 system. let's talk about consequences of the removal later
    * **UPDATE**: I have discussed that already and think it should be ok for us. However I need to find out any comment from the past where I put the result of the discussion with the python team. The problem is that python-abi on RHEL 8 is 3.6, but on RHEL 9 it's newer, which breaks the upgrade path with the leapp.

* **Comment out auto generation of Python3 dependencies**
  * The dependencies are expected to be handled via the deps package,
however the autogenerated deps are put into the python3-leapp rpm,
which is not what we want (it breaks our handy solution with deps
in case of IPU.

### DIFF
--- a/packaging/leapp.spec
+++ b/packaging/leapp.spec
@@ -20,15 +20,25 @@
 %global framework_dependencies 3
 
 # Do not build bindings for python3 for RHEL == 7
+# # Currently Py2 is dead on Fedora and we don't have to support it. As well,
+# # our current packaging is not prepared for Py2 & Py3 packages in the same
+# # time. Instead of that, make Py2 and Py3 exclusive. Possibly rename macros..
 %if 0%{?rhel} && 0%{?rhel} == 7
-%define with_python2 1
+  %define leapp_python 2
+  %define leapp_python_sitelib %{python2_sitelib}
+  %define leapp_python_name python2
+  %define leapp_py_build %{py2_build}
+  %define leapp_py_install %{py2_install}
+
 %else
-%if 0%{?rhel} && 0%{?rhel} > 7
-%bcond_with python2
-%else
-%bcond_without python2
-%endif
-%bcond_without python3
+  %define leapp_python 3
+  %define leapp_python_sitelib %{python3_sitelib}
+  %define leapp_python_name python3
+  %define leapp_py_build %{py3_build}
+  %define leapp_py_install %{py3_install}
+  # we have to drop the dependency on python(abi) completely on el8+ because
+  # of IPU (python abi is different between systems)
+  %global __requires_exclude ^python\\(abi\\) = 3\\..+|/usr/libexec/platform-python
 %endif
 
 Name:       leapp
@@ -41,20 +51,29 @@ URL:        https://oamg.github.io/leapp/
 Source0:    https://github.com/oamg/%{name}/archive/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
 BuildArch:  noarch
 
+Requires: %{leapp_python_name}-%{name} = %{version}-%{release}
+%{?python_disable_dependency_generator}
+
 %if !0%{?fedora}
-%if %{with python3}
-Requires: python3-%{name} = %{version}-%{release}
-%else
-Requires: python2-%{name} = %{version}-%{release}
-%endif
+# TODO: We plan to drop the dependency once we get ack for the planned changes.
+# # In such a case, leapp will not install leapp-repository automatically. For
+# # the full installation, people will have to install a metapackage which will
+# # be part of a particular leapp-repository (e.g. leapp-upgrade metapackage)
+# # Currently the leapp utility does not provide any subcommands (but help)
+# # so it doesn't require the leapp-repository anymore.
 
 # Just ensure the leapp repository will be installed as well. Compatibility
 # should be specified by the leapp-repository itself
 Requires: leapp-repository
 %endif # !fedora
 
+# FIXME: update the description
+# NOTE: tha man page is not updated yet!
 %description
-Leapp tool for handling upgrades.
+Leapp utility provides the possibility to use the Leapp framework via CLI.
+The utility itself does not define any subcommands but "help". All leapp
+subcommands are expected to be provided by other packages under a specific
+directory. See the man page for more details.
 
 
 ##################################################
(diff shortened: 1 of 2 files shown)


pirat89:
@vinzenz btw, i think we should stop prohibiting of the leapp rpm for fedora. As the CLI is already split, I think that nothing prevents to deliver the leapp tool under the fedora as well. WDYT?

packaging/leapp.spec
```diff
@@ -94,14 +96,51 @@ BuildRequires:  python2-setuptools
 Provides: leapp-framework = %{framework_version}
 Requires: leapp-framework-dependencies = %{framework_dependencies}
 
+# We expect people to manually remove this rpm if python3-leapp should be
+# installed
+Conflicts:      python3-%{name}
```
vinzenz, line 101: hmm, so an update on RHEL8 would conflict and you would have to remove it manually?

Scenario:
- I install leapp on RHEL8 to check upgrades and wait for a bug fix.
- Next release I update leapp on that RHEL8 machine and get a conflict
- Manually remove python3-leapp and install leapp again

That is what I am inferring from this `Conflicts` directive.


What about:
```
Conflicts: framework-version < 2.0
```

packaging/leapp.spec
```diff
@@ -240,7 +244,13 @@ rm -f %{buildroot}/%{_bindir}/leapp
 %dir %{_datadir}/leapp/
 %dir %{_datadir}/leapp/report_schema/
 %{_datadir}/leapp/report_schema
+
+%if %{with python2}
 %{python2_sitelib}/leapp/cli
```
vinzenz, line 249: define python_leapp_sitelib on the top so you don't have so many conditions around python2_sitelib / python3_sitelib

packaging/leapp.spec
pirat89: The scenario you wrote is invalid. You install leapp on RHEL 8, so you can update packages without problems. This is only about scenario: you have upgraded from rhel 7 to rhel 8, so you have the old el7 packages still. Remove these manually and than install leapp for rhel8 (or update using the `--allowerasing` option. 

However, I believe we would like to be super strict and ensure even the `--allowerasing` will not work. Look at the gdoc I shared yesterday (see the jira) to see reasoning behind. It's going to be discussed with Pavel to see his opinion on user experience. I will do a mtg the next week once you all are familiar with the document. The conflicts are set between python2-leapp and python3-leapp packages - we do not want to provide python2-leapp (or anything for python2) in packages build for el8.

packaging/leapp.spec
pirat89: I have been thinking exactly about the same! :D I wanted to discuss planned changes later before I do them as I would like to do even more stuff to make it more simple, but want to be sure everyone is ok with it and I haven't overlooked something.

packaging/leapp.spec
vinzenz: duh - I was looking at the wrong thing. Of course this is correct.

packaging/leapp.spec
vinzenz: Please propose the changes - I am only looking forward to simplification.

packaging/leapp.spec
```diff
@@ -41,67 +48,79 @@ URL:        https://oamg.github.io/leapp/
...
+BuildRequires:  python3-setuptools
+Conflicts:      python2-%{name}
+
+# NOTE:
+# comment out the python abi requirement. we need to discuss this one later.
+# currently, with the generated python(abi) = 3.6 it's not possible to preserve
+# the rpm on RHEL 9 system. let's talk about consequences of the removal later
+# and how to deal with possible problems
+# %%{?system_python_abi}
+# %%{?python_disable_dependency_generator}
```
pirat89, line 110: keeping the whole note with commented out macros to make sure people are aware why we are not using them (currently not sure whether we need or not the second one, I will check it later. it can be removed or allowed in future if we discover it's really (not) needed.

pirat89:
@vinzenz @Rezney I am going to check fully all results for el7, el8, fedora tomorrow. So far el7 builds are identical. But I need to check we get the expected results for fedora and rhel as well. - btw, I will drop tomorrow probably the limitation to build the leapp tool rpm just for rhel - so it is built on fedora too. Tell me wdyt about the changes or whether you have another ideas to make the specfile more simple / readable. I think that I will not get any idea now how to simplify it even more.

packaging/leapp.spec
pirat89: done :)

pirat89:
currently the builds are broken. I am trying to figure out, why I cannot drop the `python(abi)` dependency - which has to happened, unless we fake it and tell we provide non-existing python-abi....- which is definitely no-go. The original workaround we used in the past is not working now for an unknown reason to me.

packaging/leapp.spec
pirat89: dropping the note. I've probably realized what the poem was.

pirat89:
@Rezney  @vinzenz  Finito. Ready for review. I squashed all commits. Currently the results should be unchaned on Python2.

To see easily all provides and requires for all packages, you can use this script (when packages installed, or modify it to work with downloaded rpms locally)

```
#!/usr/bin/bash

for i in leapp leapp-deps python2-leapp python3-leapp snactor; do
    echo "======================================================="
    echo "::::::::::: $i"
    rpm -q "$i" > /dev/null 2>&1 || { echo "SKIP" ; continue ; }

    echo ":: Requires"
    rpm -q --requires $i
    echo "--------------------------------"
    echo ":: Provides"
    rpm -q --provides $i
    echo "--------------------------------"
done

```

packaging/leapp.spec
```diff
@@ -41,67 +51,75 @@ URL:        https://oamg.github.io/leapp/
...
+# # Currently the leapp utility does not provide any subcommands (but help)
+# # so it doesn't require the leapp-repository anymore.
 
 # Just ensure the leapp repository will be installed as well. Compatibility
 # should be specified by the leapp-repository itself
 Requires: leapp-repository
 %endif # !fedora
 
+# FIXME: update the description
+# NOTE: tha man page is not updated yet!
```
Rezney, line 71: What prevents us form updating it at once?

pirat89: @Rezney to be honest, I would need couple of hours to come up how to update the manpage, put the current stuff in the manpage to a new manpage provided by the leapp-repository, .... :) That's why we have the jira ticket about the documentation

Rezney (approved):

PR #722, opened 2021-07-08
Add the _lookup_actors function into the repo manager

pirat89:
The RepositoryManager.lookup_actor function returns always just one
Actor or None. However, in case the actor of the same name is specified
in multiple repositories I would like to have an info about all these
actors. Currently I am not sure whether there is a mechanism to prevent
people having actors with identical name in multiple repositories when
leapp is running, but so far, tests do not react on the situation when
some actors of the same name exists. For now, creating this private
function to be able to run unit tests for actors of specific name
in all available repos.

@vinzenz @Rezney  somthing for the discussion guys. I choosed this way (the private method) for now as we will not introduce functionality anyone should be care of so easy to remove whenever we want in the future and do not have to discuss this so much for now I believe.

### DIFF
--- a/leapp/repository/__init__.py
+++ b/leapp/repository/__init__.py
@@ -63,6 +63,7 @@ class Repository(object):
         :type name: str
         :return: None or Actor
         """
+        # TODO: what's the behaviour in case of multiple actors of the same name?
         name = name.lower()
         for actor in self.actors:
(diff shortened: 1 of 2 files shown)


pirat89:
See https://github.com/oamg/leapp-repository/pull/685

vinzenz (approved):

PR #723, opened 2021-07-14
Packaging: enable installation of manpage, etc.. on fedora

pirat89:
In the previous commit (81994a5) I have enabled building of the leapp
tool for fedora. However the files like manpage, report schema, ...
related to the tool has not been install in the install section.
Install them now.


Rezney (approved):
Looks good to me...

PR #725, opened 2021-07-22
Add packit config for Fedora copr builds

vojtechsokol:
Builds are triggered by commits to master branch and by pull requests

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,42 @@
+specfile_path: packaging/leapp.spec
+
+upstream_project_url: https://github.com/oamg/leapp
+
+jobs:
+- job: copr_build
+  trigger: pull_request
+  metadata:
+    owner: "@oamg"
+    project: leapp
+    targets:
+    - epel-7-ppc64le
+    - epel-7-x86_64
+    - epel-8-x86_64
+    - fedora-all-aarch64
+    - fedora-all-x86_64
+  actions:
+    post-upstream-clone:
+    # builds from PRs should have lower NVR than those from master branch
+    - sed -i "s/1%{?dist}/0%{?dist}/g" packaging/leapp.spec
+    get-current-version:
+    # get version from spec file instead from git tag
+    - bash -c "grep -m1 '^Version:' packaging/leapp.spec | grep -om1 '[0-9].[0-9.]**'"
+- job: copr_build
+  trigger: commit
+  metadata:
+    branch: master
+    owner: "@oamg"
+    project: leapp
+    targets:
+    - epel-7-ppc64le
+    - epel-7-x86_64
+    - epel-8-x86_64
+    - fedora-all-aarch64
+    - fedora-all-x86_64
+  actions:
+    post-upstream-clone:
+    # builds from master branch should have the highest NVR
+    - sed -i "s/1%{?dist}/100%{?dist}/g" packaging/leapp.spec
+    get-current-version:
+    # get version from spec file instead from git tag
+    - bash -c "grep -m1 '^Version:' packaging/leapp.spec | grep -om1 '[0-9].[0-9.]**'"
--- a/.github/workflows/pr-welcome-msg.yml
+++ b/.github/workflows/pr-welcome-msg.yml
@@ -23,6 +23,7 @@ jobs:
             If you want to re-run tests or request review, you can use following commands as a comment:
             - **leapp-ci build** to run copr build and e2e tests in **OAMG CI**
             - **review please** to notify leapp developers of review request
+            - **/packit copr-build** to submit a public copr build using packit
 
             Please [open ticket](https://url.corp.redhat.com/oamg-ci-issue) in case you experience technical problem with the CI. (RH internal only)
 


.packit.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - fedora-33-aarch64
+    - fedora-33-x86_64
+    - fedora-34-aarch64
+    - fedora-34-x86_64
+    - fedora-rawhide-aarch64
+    - fedora-rawhide-x86_64
+  actions:
+    post-upstream-clone:
+    # builds from PRs should have lower NVR than those from master branch
+    - sed -i "s/1%{?dist}/0%{?dist}/g" packaging/leapp.spec
```
bocekm, line 26: In Convert2RHEL, @TomasTomecek used `rpmdev-bumpspec`:
https://github.com/oamg/convert2rhel/blob/v0.22/.packit.yaml#L34

I don't know what's better so take this just as FYI.

vojtechsokol:
The only difference in copr builds from old pipelines is that naming schema is a bit different.

For master builds the change is from
`<version>-100.<timestamp:%Y%m%d%H%MZ>.<7-chars-sha>.<branch>.el7`
to
`<version>-100.<timestamp:%Y%m%d%H%M%S%f>.<branch>.<nr-commits-from-last-release>.<8-chars-sha>.fc33`

For PR builds the change is from
`<version>-0.<timestamp:%Y%m%d%H%MZ>.<7-chars-sha>.<branch>.PR<pr-number>.el7`
to
`<version>-0.<timestamp:%Y%m%d%H%M%S%f>.pr<pr-number>.<nr-commits-from-last-release>.<8-chars-sha>.fc33`

The only two regressions in the new naming schema are:
* timestamps are no longer explicitly in UTC (but have finer resolution)
* source branch is missing for PR builds - this regression is actually not a regression any more because the old pipelines use `HEAD` instead of the correct branch

@pirat89 Are you OK with these changes? If not then there is no way how we could use packit for copr build (at least until/if they provide a way to define the nvr). If yes, then you need to grant `builder` permission to packit.

.packit.yaml
vojtechsokol: Actually I used c2r's packit config as a starting point but wanted to keep the naming schema as much similiar as possible to the already existing one, that's why I'm using `sed` to set the release number to `100`.

pirat89:
@vojtechsokol I would be ok with it but as I remember, murphy / morf / .... is searching for exact NVRA, so the change could possibly break the QE tooling. Let's sync with QE - ping  @mkluson  @AloisMahdal , can you check the tooling is ok with this change?

TomasTomecek:
since you tagged me in this PR, I'd like to point out that I am going to be on a vacation for one week so don't expect any activity from me in that time, will tag some hardworkers from our team: @lachmanfrantisek @jpopelka

.packit.yaml
TomasTomecek: it's really just a matter of preference: rpmdev-bumpspec actually parses the spec file and does some magic which can lead to inconsistent results while sed is very explicit - just please make sure to get the regex right so it doesn't change more things than you'd want it to

.packit.yaml
```diff
@@ -0,0 +1,52 @@
+specfile_path: packaging/leapp.spec
+
+upstream_package_name: leapp
+downstream_package_name: leapp
```
jpopelka, line 4: These should be safe to remove, they both default to the name of the Github repository, i.e. `leapp`

```suggestion
```

.packit.yaml
```diff
@@ -0,0 +1,50 @@
...
+    targets:
+    - epel-7-ppc64le
+    - epel-7-x86_64
+    - epel-8-x86_64
+    - fedora-33-aarch64
+    - fedora-33-x86_64
+    - fedora-34-aarch64
+    - fedora-34-x86_64
+    - fedora-rawhide-aarch64
+    - fedora-rawhide-x86_64
```
lachmanfrantisek, line 20: You can use an [alias](https://packit.dev/docs/configuration/#available-copr-build-targets) to not need to update the chroot list twice a year.

```suggestion
    - fedora-all-aarch64
    - fedora-all-x86_64
```

.packit.yaml
```diff
@@ -0,0 +1,50 @@
...
```
lachmanfrantisek, line 43:
```suggestion
    - fedora-all-aarch64
    - fedora-all-x86_64
```

vojtechsokol: Not sure if this is wanted. @pirat89 Do you think it's a good idea to have the copr targets moved automatically or do you want to retain manual control of them?

Also where can I find the aliases and their mapping? From packit [docs](https://packit.dev/docs/configuration/#available-copr-build-targets) `fedora-all` maps to `fedora-stable` + `fedora-development`. IIUC Fedora does not know anything like stable, only [supported releases (f33 + f34)](https://fedoraproject.org/wiki/Releases#Current_Supported_Releases) and [developmnet releases (f35 + rawhide)](https://fedoraproject.org/wiki/Releases#Development). If `fedora-stable` = supported releases and `fedora-development` = development releases then it does not correspond to [packit config](https://github.com/packit/packit/blob/main/packit/config/aliases.py#L34)...

lachmanfrantisek: > IIUC Fedora does not know anything like stable, only [supported releases (f33 + f34)](https://fedoraproject.org/wiki/Releases#Current_Supported_Releases) and [developmnet releases (f35 + rawhide)](https://fedoraproject.org/wiki/Releases#Development). If `fedora-stable` = supported releases and `fedora-development` = development releases then it does not correspond to [packit config](https://github.com/packit/packit/blob/main/packit/config/aliases.py#L34)...

Your expectations are correct: `stable=supported` and `development=development`. The wording might not be accurate (will improve that), but I hope it can be understood. There are two problems with the `f35`:
* The mapping you've linked is a backup, we try to get this info from bodhi. The hardcoded one is used only when the API is not available. (Thanks for the reminder, we need to update that.)
* We also check, if the chroots/targets are available on `Copr` so it does not need to be available even when it's given to us by Bodhi.

pirat89: @vojtechsokol I think we could give it a try. In case we find a not-simple python issue with a newest fedora, we will just change it. It's better to know when leapp becomes broken for the newest python/system.

.packit.yaml
vojtechsokol: The exact same regex is used in [Makefile](https://github.com/oamg/leapp/blob/408d99628ad4b1c35542f7d2430cf145281c34f5/Makefile#L91) for quite some time so I believe it should be reasonably safe.

.packit.yaml
lachmanfrantisek: @vojtechsokol

Here is an update of the documentation if you want to take a look: https://github.com/packit/packit.dev/pull/314
and here is the update of the backup values: https://github.com/packit/packit/pull/1315

pirat89:
@mkluson @oamg/qe  guys are aware we are waiting for your response here?

pirat89:
There is an outage now for packit, so I let's try it once it's resolved:
- https://status.packit.dev/

pirat89 (approved):
Giving my unlimited love

PR #726, opened 2021-07-27
Drop dependency on /usr/bin/python3.6

pirat89:
There is difference in the automaticly generated dependencies when the RPM is built in the internal COPR for RHEL-8 or when it's built in public COPR for EPEL-8:
 - EPEL-8: `/usr/bin/python3.6`
 - RHEL-8: `/usr/libexec/platform-python`

Extending the `__requires_exclude` macro to drop both.

In case the dependency is not dropped, the `leapp` rpm is automatically removed during the upgrade process which ends with the emergency mode as we are not able to run leapp after the rpm transaction is processed.

### DIFF
--- a/packaging/leapp.spec
+++ b/packaging/leapp.spec
@@ -38,7 +38,7 @@
   %define leapp_py_install %{py3_install}
   # we have to drop the dependency on python(abi) completely on el8+ because
   # of IPU (python abi is different between systems)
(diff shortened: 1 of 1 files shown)


pirat89:
@oamg/developers @Rezney  @vinzenz  ping

Rezney (approved):
Looks good to me...

PR #727, opened 2021-08-02
Packaging: requires leapp-repository on RHEL 7 only

pirat89:
The leapp tool doesn't depend on leapp-repository anymore. However,
when no leapp repository is present on the system, there is nothing
to do usually when leapp is installed as all subcommands for leapp
should be provided by specific leapp repositorise.

I wanted to remove the dependency completly and keep it just in
downstream, however not sure whether that could be a good way.
So let's keep the requirement to provide the leapp-repository
at least as a capability for rpms that want to use leapp on RHEL 7.

People are going to be instructed to install e.g. `leapp-upgrade`
or `leapp-convert` to install everything what they need for the
action they want to do.

It should be merged together with https://github.com/oamg/leapp-repository/pull/701 ideally, but it's not hard requirement.

### DIFF
--- a/packaging/leapp.spec
+++ b/packaging/leapp.spec
@@ -23,7 +23,7 @@
 # # Currently Py2 is dead on Fedora and we don't have to support it. As well,
 # # our current packaging is not prepared for Py2 & Py3 packages in the same
 # # time. Instead of that, make Py2 and Py3 exclusive. Possibly rename macros..
-%if 0%{?rhel} && 0%{?rhel} == 7
+%if 0%{?rhel} == 7
   %define leapp_python 2
   %define leapp_python_sitelib %{python2_sitelib}
   %define leapp_python_name python2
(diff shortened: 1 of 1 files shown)


pirat89:
leapp-ci build

pirat89:
(setting back to the draft until I finish testing and development. I've tried the switch to PR to see whether it has effect on the CI/CD that is broken now).

pirat89:
Prepared for the review. I am keeping it in the draft state to ensure we do not merge it before the testing tools and CI is adapted to the changes.

pirat89:
the man page is not updated yet (still no idea how to update it to be honest right now, but consider it as minor thing which could be done in a separate PR - we need to finally move this to the merge)

pirat89:
Currently there are some issues arround ppc64le builds caused by infrastracture (last couple of hours). I will check it again tomorrow

Rezney:
Discussed offline with @pirat89 . Looks good to me.

Rezney (approved):

PR #729, opened 2021-08-10
fix link to inplace_upgrade.py

evgeni:
it was moved in https://github.com/oamg/leapp-repository/commit/83365f565de8893a2692b7f3e028d052fcdf0393, but the docs weren't updated


vinzenz (approved):

vinzenz:
Thank you very much for your fixing the link!

PR #730, opened 2021-08-19
Add name parameter to the tft request

fernflower:
As tmt-plans repository now contains morf tests for convert2rhel this
is needed to filter out tests that make no sense for leapp.


mkluson (approved):

vojtechsokol (approved):
LGTM

PR #734, opened 2021-10-04
Add cli support for choices and default

fernflower:
Extending leapp command to support traditional argparse-style choices
and default arguments.


vinzenz (approved):
LGTM

pirat89 (approved):

PR #739, opened 2021-11-02
Change tmt plans filtering

fernflower:
Instead of specifying which tests to run (leapp, integration, e2e)
switch to specifying which tests to skip (c2r, sap).


fernflower:
Proof of viability - http://artifacts.osci.redhat.com/testing-farm/0b93cf88-9263-47d0-99d8-1a3461a70744

vojtechsokol:
Looks like all the tests that should have been run were run and none of the excluded ones were run, so LGTM.

vojtechsokol (approved):

PR #740, opened 2021-11-02
Add default upgrade_paths generation

fernflower:
For POC leapp.conf will store basic information about possible
upgrade_paths. Maybe will move to some other place later.

### DIFF
--- a/etc/leapp/leapp.conf
+++ b/etc/leapp/leapp.conf
@@ -4,3 +4,5 @@ repo_path=/etc/leapp/repos.d/
 [database]
 path=/var/lib/leapp/leapp.db
 
+[upgrade]
+upgrade_paths=/usr/share/leapp-repository/upgrade_paths.json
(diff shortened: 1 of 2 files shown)


pirat89 (changes requested):
this is bad by design as the leapp has nothing to do with specific upgrade. we spent a lot of time to separate logic, this data cannot be part of the leapp

fernflower:
@pirat89 We can make it a leapp-repository specific config file then.

pirat89 (changes requested):
There is still the connection to the upgrade paths, which I mean in the way that we should not see any PR working anyhow about any upgrade paths..

In the framework, only the generic solutions are expected. We do not expect to install some upgrade_paths notes when people are going to convert. THe logic about upgrade paths has to be purely just in the leapp-repository located.

fernflower:
> In the framework, only the generic solutions are expected. We do not expect to install some upgrade_paths notes when people are going to convert. THe logic about upgrade paths has to be purely just in the leapp-repository located.
The framework will be used for 2 things - conversions and upgrades. I suggest we add a dedicated section about 'upgrades' to the leapp config where all upgrade-only-specific config file references go, like upgrade_paths. 

The latest commit should give you the idea.

pirat89:
closing regarding the investigation.

PR #742, opened 2021-11-22
get_common_X_path functions into a library

fernflower:
Those methods in actor/__init__.py have been static anyway, by moving them
out not only we do the right thing, but make them accessible in non-actor
libraries like command libraries.


leapp/libraries/stdlib/path.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    for path in directories:
+        path = os.path.join(path, name)
+        if os.path.isfile(path) and os.access(path, os.X_OK):
+            return path
+
+
+# NOTE(ivasilev) You can use this standalone with
+# manager = load_repositories_from('repo_path', '/etc/leapp/repo.d/', manager=None)
+def get_common_file_path(manager, name):
```
vinzenz, line 51: `manager` is implementing the repository protocol, in terms of Python 3.6+ typing.Protocol 
So I would call it `repository`, not `manager`

An instance of `leapp.repository.Manager` is returned, when more than one repository has been scanned

leapp/actors/__init__.py
```diff
@@ -206,7 +191,8 @@ def get_folder_path(self, name):
         :return: Found folder path
         :rtype: str or None
         """
-        return self._get_folder_path(self.files_paths, name)
+        from leapp.libraries.stdlib import path
```
vinzenz, line 194: If you cannot do it globally, because of circular imports, I'd probably import it once in __init__ and assign it to self._path, WDYT?

fernflower: to be honest I don't really like any option in "Inplace import and code duplication" vs "adding a rather obscure instance attribute". 
But probably your way is better.

leapp/libraries/stdlib/path.py
fernflower: Should be done now for all 3 functions

fernflower:
@vinzenz All resolved, thanks!

pirat89 (approved):
Seems good to me and so far I haven't discovered any problem functional problem with this change

PR #747, opened 2022-01-05
Add python(abi) == 3.x deps for RHEL 8+ into the meta-pkg

pirat89:
On RHEL 8 it could happen that people replace the original Python 3.6
by Python 3.8+. However, regarding the build & installation process,
we can support just ONE version of python per RPM - except the newer
version of python on the upgraded system, which is handled during
the upgrade by our workarounds. For that reason we put the abi
dependency back into the metapackage for RHEL 8 system only.

Example of the problem on RHEL 8:
```
# leapp
-bash: /usr/bin/leapp: /usr/bin/python3.6: bad interpreter: No such file or directory
```

### DIFF
--- a/packaging/leapp.spec
+++ b/packaging/leapp.spec
@@ -17,7 +17,7 @@
 
 # IMPORTANT: everytime the requirements are changed, increment number by one
 # - same for Provides in deps subpackage
-%global framework_dependencies 3
+%global framework_dependencies 4
 
 # Do not build bindings for python3 for RHEL == 7
 # # Currently Py2 is dead on Fedora and we don't have to support it. As well,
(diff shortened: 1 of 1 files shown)


packaging/leapp.spec
```diff
@@ -137,6 +140,11 @@ Requires: python-requests
 Requires: python3-six
 Requires: python3-setuptools
 Requires: python3-requests
+%if 0%{?rhel} == 8
+# this prevents situation when someone would like to install leapp with
+# python-3.8+ on RHEL 8, but does not affect any else systems nor upgrades.
+Requires: python(abi) = 3.6
```
vinzenz, line 146: We should this right away for el9/py3.9

pirat89: @pirat89  Currently we are not creating any builds for el9 yet but probably you are right. we could do that already. give me a minute.

pirat89: Done.

Rezney (approved):
Works fine. Thanks for the patch.

Rezney (changes requested):
De-approving for now, as I am not sure if the fail in the e2e test is repository dependency related.

Rezney (approved):

PR #753, opened 2022-01-27
Check answerfile upon loading

fernflower:
This patch will forbid having values violating `key = value` format in *answerfile* and *answerfile.userchoices*.
If any of those files has invalid value - leapp will fail early informing the user about the location of the problematic option.

OAMG-5861

### DIFF
--- a/leapp/messaging/answerstore.py
+++ b/leapp/messaging/answerstore.py
@@ -3,6 +3,7 @@ import multiprocessing
 import six
 from six.moves import configparser
 
+from leapp.exceptions import CommandError
 from leapp.utils.audit import create_audit_entry
 
 
@@ -50,15 +51,34 @@ class AnswerStore(object):
             conf.write(afile)
         return not_updated
 
+
+    def _load_ini(self, inifile):
+        """
+        Loads an ini config file from the given location.
+
+        :param inifile: Path to the answer file to load.
+        :return: configparser.SafeConfigParser object
+        :raises CommandError if any of the values are not in key=value format
+        """
+        conf = configparser.SafeConfigParser(allow_no_value=False)
+        try:
+            conf.read(inifile)
+            return conf
+        except configparser.ParsingError as exc:
+            # Some of the sections were not in key = value format
+            raise CommandError('Failed to load answer file {inifile} with the following errors: {errors}'.format(
+                inifile=inifile, errors=exc.message))
+
+
     def load(self, answer_file):
         """
         Loads an answer file from the given location and updates the loaded data with it.
 
         :param answer_file: Path to the answer file to load.
         :return: None
+        :raises CommandError if any of the values are not in key=value format
         """
-        conf = configparser.SafeConfigParser(allow_no_value=True)
-        conf.read(answer_file)
+        conf = self._load_ini(answer_file)
         for section in conf.sections():
             self._storage[section] = self._manager.dict(conf.items(section=section, raw=True))
 
@@ -70,9 +90,9 @@ class AnswerStore(object):
         :param answer_file: Path to the answer file to load.
         :param workflow:
         :return: None
+        :raises CommandError if any of the values are not in key=value format
         """
-        conf = configparser.SafeConfigParser(allow_no_value=True)
-        conf.read(answer_file)
+        conf = self._load_ini(answer_file)
         for section in conf.sections():
             self._storage[section] = dict(conf.items(section=section, raw=True))
         self.translate_for_workflow(workflow)


leapp/messaging/answerstore.py
```diff
@@ -50,15 +51,33 @@ def update(self, answer_file, allow_missing=False):
...
+        :return: configparser.SafeConfigParser object
+        :raises CommandError if any of the values are not in key=value format
+        """
+        conf = configparser.SafeConfigParser(allow_no_value=False)
+        try:
+            conf.read(inifile)
+            return conf
+        except configparser.ParsingError as exc:
+            # Some of the sections were not in key = value format
+            raise CommandError(exc.message)
```
pirat89, line 69: @fernflower @vinzenz   shouldn't be this rather `ValueError`? Or something else? Just curious why `CommandError`

fernflower: My first thought was to take some leapp specific exception as we are dealing with leapp answerfile but I have no hard feelings about it.

vinzenz: I think CommandError is fine, because it will be caught by leapp and displayed as an appropriate failure of the execution. However I feel like it's necessary to bring a bit more light here. 

In the end I'd prefer if it would read like this:
```
Failed to load answer file `{inifile}` with following errors: {exc.message}
```

vinzenz: I'd like to see a bit more of a message like this:
```
Failed to load answerfile `{inifile}` due to errors reading the file: {exc.message}
```

pirat89:
## Original reproducer for the issue
1. install nowadays upstream build of leapp-upgrade*
2. add newlines into a dialog that will be hit on the system (see the code below ↓)
3. run leapp preupgrade (You can see the corrupted answerfile is created.)
4. answer the question (do not use `leapp answer` !!)
5. run leapp upgrade (leapp does not crash now, it creates answerfile.userchoice with not ideal content... but good enough hack.. after the reboot, leapp crashes on invaild content inside the answerfile)



```shell
dnf install leapp-upgrade
sed -i "s/configuration? /configuration?\\\n/" /etc/leapp/repos.d/system_upgrade/el7toel8/actors/removeoldpammodulescheck/actor.py
leapp preupgrade --debug
echo "confirm = True" >> /var/log/leapp/answerfile
leapp upgrade --debug
```

Content of the */var/log/leapp/answerfile* file after the first run:
```ini
[remove_pam_pkcs11_module_check]
# Title:              None
# Reason:             Confirmation
# =================== remove_pam_pkcs11_module_check.confirm ==================
# Label:              Disable pam_pkcs11 module in PAM configuration?
If no, the upgrade process will be interrupted.
# Description:        PAM module pam_pkcs11 is no longer available in RHEL-8 since it was replaced by SSSD.
# Type:               bool
# Default:            None
# Available choices: True/False
# Unanswered question. Uncomment the following line with your answer
# confirm = 

```

Also, after the second run you see how the nowadays `answerfile.userchoice` looks like (originally, the `= None` has not been present IIRC):
```ini
[remove_pam_pkcs11_module_check]
if no, the upgrade process will be interrupted. = None
confirm = True

```

See the line is now a little bit fixed to the answerfile.userchoice is not so heavily corrupted. however, it seems weird to me to have a key with spaces (multiple words).

## Testing with the PR build

Using the build for this PR (with updated particular actor) and files created before (to reflect situation of some customers who already generated invalid content), I can see the answerfile and the answerfile.userchoice files are checked when loaded. However, the **answerfile.userchoice** is not checked.  Probably it is not strictly an issue anymore as the invalid line inside *answerfile.userchoice* is updated to contain the ` = None` value. However, I would expect that  

To check that answerfile.userchoice is checked when loaded, you can create new one with the error or just modify the existing one e.g. like this:
```shell
sed "s/ = None$//" /var/log/leapp/answerfile.userchoices
```
Or just add invalid line:
```
echo >> "my invalid line" /var/log/leapp/answerfile.userchoices
```

Some results in case of invalid files:
```
...
Error: Failed to load answer file /var/log/leapp/answerfile.userchoices with the following errors: File contains parsing errors: /var/log/leapp/answerfile.userchoices
	[line  2]: 'if no, the upgrade process will be interrupted.\n'

....

Error: Failed to load answer file /var/log/leapp/answerfile with the following errors: File contains parsing errors: /var/log/leapp/answerfile
	[line  6]: 'If no, the upgrade process will be interrupted.\n'

```

## What is missing? (separate PR?)
However, imagine this situation:
* customer removes the answerfile AND have a valid answerfile.userhoices file AND and an actor produces the problematic dialog
* or customer answer the question using the `leapp answer` command AND an actor produces the problematic dialog

In such a situation, they can execute `leapp upgrade` without errors until the reboot. After the reboot upgrade is immadiately stopped with the errors above. So the situation is better but still it's not the final solution for OAMG-5861. As discussed, this problem is giong to be resolved probably in another separate PR so the testing is easier.

pirat89 (dismissed):
lgtm and works as expected. As mentioned in the comment above, the generation of invalid answerfile will be resolved probably in the separate PR.

pirat89 (approved):

PR #754, opened 2022-01-31
Introduce 8to9 regression testing

fernflower:
Both 7to8 and 8to9 upgrade paths will be tested for every
upstream pr.
For rare situations where leapp pr had to be tested together with
specific version of leapp-repository pr enable the same
'/rerun dependent-leapp-repository-pr-num' mechanism as is present
in leapp-repository's tmt-plans workflow.

OAMG-6387

### DIFF
--- a/.github/workflows/tmt-tests.yml
+++ b/.github/workflows/tmt-tests.yml
@@ -11,7 +11,7 @@ jobs:
     name: Run tmt tests on Testing Farm service
     if: |
       github.event.issue.pull_request
-      && github.event.comment.body == '/rerun'
+      && startsWith(github.event.comment.body, '/rerun')
       && contains(fromJson('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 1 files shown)


fernflower:
Proof for /rerun:
Action - https://github.com/fernflower/leapp/actions/runs/1772660977
https://github.com/fernflower/leapp/pull/3#issuecomment-1025714870 and https://github.com/fernflower/leapp/pull/3#issuecomment-1025714892 .
Proof for /rerun 42:
Action - https://github.com/fernflower/leapp/runs/5006178533?check_suite_focus=true

.github/workflows/tmt-tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     name: Run tmt tests on Testing Farm service
     if: |
       github.event.issue.pull_request
-      && github.event.comment.body == '/rerun'
+      && startsWith(github.event.comment.body, '/rerun')
```
vinzenz, line 14: no spaces allowed huh?

fernflower: TBH I didn't give it much thought though, it's how it has always been with our comment-based commands.
How about I fix that in a separate patch? So that I could simultaneously do that for both leapp and leapp-repository?

vinzenz: It wasn't really a 'Hey you need to fix that', just more like I didn't even realize. But TBH I am fine with it as is.

vinzenz (approved):

PR #757, opened 2022-02-07
[WIP] Comment out strings with newlines properly

fernflower:
Introduce a _comment_out helper function to be used
in answerfile generation. Tests included.

Also unify ini file loading in answerstore and refactor
into a class method.

Related: OAMG-5861


leapp/messaging/answerstore.py
```diff
@@ -7,6 +7,14 @@
 from leapp.utils.audit import create_audit_entry
 
 
+def _comment_out(key, text):
+    """Returns a commented-out string. If any newlines are present it properly deals with them"""
+    split_by_newline = [s for s in text.split('\n') if s.strip()]
+    res = '# {:<20}{}\n{}'.format('{}:'.format(key), split_by_newline[0],
```
vinzenz, line 13: Hmm can't we just do:
```
first, rest = text.split('\n', 1)
if rest:
    rest = "# {}".format(rest.replace('\n', '\n# '))
...
```
Instead of looping over all of them?

vinzenz: Ofc that requires a newline to be present

leapp/messaging/answerstore.py
```diff
@@ -183,9 +193,10 @@ def generate(self, dialogs, answer_file_path):
                         answer_entry += '# {key} = {value}\n'.format(key=component.key, value=default or '')
 
                     f.writelines([
-                        '# {}\n'.format(' {}.{} '.format(dialog.scope, component.key).center(77, '=')),
-                        '# {:<20}{}\n'.format('Label:', component.label),
-                        '# {:<20}{}\n'.format('Description:', component.description),
+                        '# {}\n'.format(' {}.{} '.format(dialog.scope,
+                                                         component.key.replace('\n', '')).center(77, '=')),
```
fernflower, line 197: These \n replacement should not be here, sorry, this would be solved differently.

leapp/messaging/answerstore.py
fernflower: I wanted to get rid of excessive newlines (like '\n\n\n') if they happen to appear.
But I'll give it some thinking, thanks.

fernflower: In the end I think it'll be more beneficial to have it like this https://github.com/oamg/leapp/pull/757/files#diff-3d756cc0c41cb26dd63ed05cfe760651cf1b33ec460fe203a5ae9aa8cdc0d7fdR319

vinzenz (dismissed):

leapp/dialogs/components.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, key=None, label=None, description=None, default=None, reason=
         :param reason: The reason why we need this value.
         """
         self.key = key or type(self).key
-        self.label = label or type(self).label
+        self.label = label.replace('\n', '') if label else type(self).label
```
pirat89, line 26: replacing `Give me your\nhand` should not result in `Give me yourhand`
```suggestion
        self.label = label.replace('\n', ' ') if label else type(self).label
```

vinzenz: I'd add a strip :D
```suggestion
        self.label = label.replace('\n', ' ').strip() if label else type(self).label
```

tests/scripts/test_answerstore.py
```diff
@@ -312,3 +312,13 @@ def test_answerfile_generate(tmpdir, answerfile):
...
+
+
+def test__comment_out():
+    text = 'A really long line of text.\nYep, with newlines.\n\nCouple of them.'
+    expected_text = ('# inikey:             A really long line of text.\n'
+                     '#                     Yep, with newlines.\n'
+                     '#                     Couple of them.')
+    key = 'inikey'
+    commented_text = _comment_out(key, text)
+    assert commented_text == expected_text
```
pirat89, line 324: add test for the string without the newline. current solution could crash if I see right.

leapp/messaging/answerstore.py
```diff
@@ -7,6 +7,14 @@
 from leapp.utils.audit import create_audit_entry
 
 
+def _comment_out(key, text):
+    """Returns a commented-out string. If any newlines are present it properly deals with them"""
+    split_by_newline = [s for s in text.split('\n') if s.strip()]
+    res = '# {:<20}{}\n{}'.format('{}:'.format(key), split_by_newline[0],
+                                  '\n'.join(['# {}{}'.format(' ' * 20, l) for l in split_by_newline[1:]]))
+    return res
```
pirat89, line 15: @fernflower  I could not help myself, but somehow I like the recursion in this function :) no need to do it in this way of course
```suggestion
def _comment_out(key, text, key_separator=':'):
    """Returns a commented-out string. If any newlines are present it properly deals with them"""
    if not text.strip():
        return ''
    split_by_newline = text.split('\n', 1)
    if len(split_by_newline) == 1:
        return '# {:<20}{}\n'.format('{}{}'.format(key, key_separator), split_by_newline[0])
    else:
        return '# {:<20}{}\n{}'.format('{}{}'.format(key, key_separator), split_by_newline[0], _comment_out('', split_by_newline[1], ''))
```

tests/scripts/test_answerstore.py
fernflower: No it won't crash :) Test added.

leapp/dialogs/components.py
fernflower: Resolved for both label and description, thanks!

leapp/messaging/answerstore.py
fernflower: Making it recursive is a nice idea, thanks! Took a bit different approach though.

tests/scripts/test_answerstore.py
pirat89: you are right! :) I though the behaviour is different in python for `[][5:]` :D nice to know!

leapp/dialogs/dialog.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, scope, reason, title=None, components=None, key=None):
         :type key: str
         """
         self.components = components or self.components
-        self.title = title
+        self.title = title.replace('\n', ' ').strip() if title else title
```
pirat89, line 37: @fernflower  just curious, why the condition? I thought the title is mandatory and cannot be anything else than string. Unless it can be `None` object, why to have the condition?

pirat89 (approved):
I have one question regarding the title, but it's good otherwise :)

leapp/dialogs/dialog.py
fernflower: Title can be None, calling replace on None would result in AttributeError

pirat89: @fernflower thx for the explanation. It's weird to me that title can actually `None` at all, but the condition makes sense now.

PR #759, opened 2022-02-25
Fix store of answers on Python3.6 (DictProxy)

pirat89:
On Python3 the loaded answer is "sometimes" represented by DictProxy
object. In such a case the answer cannot be serialized into the
JSON object and storing into the audit db fails with crash.

"sometimes" ==> I have a suspicion the result is different in case
                the answer is loaded from the answerfile and
                userchoices file.

See: https://github.com/oamg/leapp-repository/pull/821

Note: This is not a proper fix. I am sending the commit from a train
      with unstable network.

### DIFF
--- a/leapp/messaging/answerstore.py
+++ b/leapp/messaging/answerstore.py
@@ -118,7 +118,12 @@ class AnswerStore(object):
         :return:
         """
         answer = self._storage.get(scope, fallback)
-        create_audit_entry('dialog-answer', {'scope': scope, 'fallback': fallback, 'answer': answer})
+        try:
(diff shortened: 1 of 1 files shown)


fernflower:
I'd say this is the proper fix https://github.com/oamg/leapp/pull/760

pirat89:
Closing it in favor of #760 

btw, I am hitting it now in 100% cases on Py3 with dialogs. I have no idea how I passed through it earlier this day when the dialog has been raised. the object is `DictProxy` all the time the question is answered.

PR #760, opened 2022-02-25
Fix TypeError during JSON serialization in dialogs

fernflower:
Previously a DictProxy entry was passed as part of
{dialog_section: entry} data to save in store which
failed during serialization.

### DIFF
--- a/leapp/messaging/answerstore.py
+++ b/leapp/messaging/answerstore.py
@@ -115,9 +115,11 @@ class AnswerStore(object):
 
         :param scope: Scope of the data to retrieve.
         :param fallback: Fallback value to return if not found.
-        :return:
+        :return: A shallow copy of data stored in _storage by scope key
         """
-        answer = self._storage.get(scope, fallback)
+        # NOTE(ivasilev) self.storage.get() will return a DictProxy. To avoid TypeError during later
+        # JSON serialization a copy() should be invoked to get a shallow copy of data
+        answer = self._storage.get(scope, fallback).copy()
         create_audit_entry('dialog-answer', {'scope': scope, 'fallback': fallback, 'answer': answer})
         return answer
 
--- a/leapp/dialogs/dialog.py
+++ b/leapp/dialogs/dialog.py
@@ -98,7 +98,7 @@ class Dialog(object):
         :return: Dictionary with answers once retrieved
         """
         store.translate(self)
-        return dict(store.get(self.scope, {}))
+        return store.get(self.scope, {})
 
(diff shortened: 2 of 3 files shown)


fernflower:
Hmm I see now that this doesn't fix the issue with the following configuration
answerfile
```
[check_vdo]
# totally empty
#
```
answerfile.userchoices
[check_vdo]
no_vdo_devices = False

The failure happens elsewhere, looking

leapp/messaging/answerstore.py
```diff
@@ -115,9 +115,11 @@ def get(self, scope, fallback=None):
 
         :param scope: Scope of the data to retrieve.
         :param fallback: Fallback value to return if not found.
-        :return:
+        :return: A shallow copy of data stored in _storage by scope key
         """
-        answer = self._storage.get(scope, fallback)
+        # NOTE(ivasilev) self.storage.get() will return a DictProxy. To avoid TypeError during later
+        # JSON serialization a copy() should be invoked to get a shallow copy of data
+        answer = self._storage.get(scope, fallback).copy()
```
pirat89, line 122: @fernflower  you are right. Regarding the amount of existing and expected dialogs (and use case when it contains usually just one key) at all, it does not matter to create the shallow copy always.

pirat89 (approved):
works good except the problem mentioned above. but that's a different issue already present probably for a longer time. We do not expect people to remove the `key = value` line from the ansewrfile, keeping the section defined. This solution fixes the general problem.

@fernflower bumped  `leapp-framework = 2.2` so we could start to require it in leapp-upgrade packages to ommit possible issues when people could test leapp-upgrade builds with dialogs but older leapp builds. Feel free to squash it directly into the one commit (i didn't want to touch the orig one as tests are running for it already).

fernflower:
> works good except the problem mentioned above. but that's a different issue already present probably for a longer time. We do not expect people to remove the key = value line from the ansewrfile, keeping the section defined. This solution fixes the general problem.

@pirat89 what do you mean by "except the problem mentioned above"? If you are referring to my comment https://github.com/oamg/leapp/pull/760#issuecomment-1051309929 then it is obsolete, the current version of the patch resolves this issue as well.

pirat89:
@fernflower I have still some problems regarding that (tried manually, when the answer in the answerfile is not present but the section exists, the answer is not loaded from the userchoices file)

pirat89:
merging after the discussion. the issue mentioned above will be fixed in another PR as it's  a different problem.

PR #763, opened 2022-03-09
Unify leapp-preupgrade and leapp-upgrade logs

fernflower:
Instead of having a dedicated leapp-CMDNAME file for
each command log latest run as leapp.log.

### DIFF
--- a/leapp/config.py
+++ b/leapp/config.py
@@ -18,7 +18,6 @@ _REPORTS = [
 # debug logs that will get reported at the end of a preupgrade/upgrade run if they were created/modified during it
 _LOGS = [
     'leapp-upgrade.log',
-    'leapp-preupgrade.log'
 ]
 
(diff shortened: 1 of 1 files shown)


vinzenz:
Well if you would do this now, this would mean you would have to update all the documentation. Also would this clash a bit with C2R - While I am pretty sure why you think we need this, I think there might be an alternative solution. I think it might work if you would set leapp.logger._logger to None after you're done.

fernflower:
Well iiuc clash with convert2rhel is not something we have to worry about anymore, as they won't be using leapp for conversions.
If we don't have doc freeze I'd rather have 1 log file, but maybe `leapp.log` is a very bad name yeah. How about keeping `leapp-upgrade.log` for both preupgrade and upgrade? Or maybe just `leapp-run.log`?

fernflower:
This one came in pair with https://github.com/oamg/leapp-repository/pull/850 which is now closed in favor of https://github.com/oamg/leapp/pull/764

PR #764, opened 2022-03-10
logging: Fix missing logging

vinzenz:
Previously we started using logging before it had been initialized.
Which resulted in leapp stopping to log for some logger names the debug
output. We did not know why this happened. It turns out that fileConfig
from the logging library has a parameter disable_existing_loggers which
is by default set to True and will set all loggers that have existed
before the call to fileConfig to disabled. Which in turn will make them
stop logging of course.
This patch will change this behaviour and set disable_existing_loggers
to False in order to have existing loggers be still enabled.


Rezney (approved):
The original issue has been fixed.

PR #770, opened 2022-03-29
Enable building leapp for specific chroots only

fernflower:
Instead of building for all project chroots use -r/--chroot
option to specify chroots.
Tmt-tests workflow and copr-build workflow will be passing proper
chroots now via COPR_CHROOT env var.
Also some refactoring has been done to the workflows to move all
magic constants to env section.

OAMG-6735


fernflower:
My bad, leapp is supposed to be built everywhere. So closing this and will limit chroots for leapp-repository only

PR #772, opened 2022-04-28
Pass enable root auth post-install-script

fernflower:
This should allow root auth on every guest, even those that
don't have it enabled by default.
This patch also pins tft github action to 1.2.10.

OAMG-6748


fernflower:
Proof it works - workflow file https://github.com/fernflower/leapp/actions/runs/2239316914/workflow , workflow run https://github.com/fernflower/leapp/actions/runs/2239316914

fernflower:
Okay, now it should work (added a missing krb5-config installation step) https://github.com/fernflower/leapp/actions/runs/2239393551

pirat89 (approved):

PR #781, opened 2022-06-21
Preserve the original inhibitor detection valid for now

pirat89:
As we substituted Tags and Flags with Groups recently, it has
impact on original detection of inhibitor inside leapp-repository
which is not compatible with the new reporting implementation.
As it could happen that new framework and old leapp-repository could
be installed on the system, the consequence in such a case is that
inhibitors will not be caught and user running:
```
  # leapp upgrade --reboot
```
could hit various serious issues. This change make sure we keep the
original functionality working, until we drop flags & tags
completely, providing enough time to everyone to adapt to new
changes.

### DIFF
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -326,6 +326,12 @@ def _create_report_object(entries):
     _sanitize_entries(entries)
     for entry in entries:
         entry.apply(report)
+    if Groups.INHIBITOR in report.get('groups', []):
+        # Before removing Flags, the original inhibitor detection worked
+        # by checking the `flags` field; keep the `flags` field until we drop Flags completely
+        # Currently we know that 'flags' does not exist otherwise so it's
+        # safe to just set it
+        report['flags'] = [Groups.INHIBITOR]
 
     return Report(report=report)
 
--- a/leapp/utils/report.py
+++ b/leapp/utils/report.py
@@ -139,4 +139,15 @@ def generate_report_file(messages_to_report, context, path, report_schema='1.1.0
                     groups = msg.pop('groups', [])
                     msg['flags'] = [g for g in groups if g in _DEPRECATION_FLAGS]
                     msg['tags'] = [g for g in groups if g not in _DEPRECATION_FLAGS]
+            if report_schema_tuple == (1, 2, 0):
+                # DEPRECATED: flags, tags
+                # NOTE(pstodulk): This is a temporary solution to ensure that
+                # flags are not present in the JSON output as we need to re-introduce
+                # flags internally to ensure that any inhibitor checks will not
+                # be negatively affected until we drop flags and tags completely.
+                messages_to_report = list(messages_to_report)
+                for msg in messages_to_report:
+                    # NOTE(ivasilev) As flags field is created only if there is an inhibitor
+                    # a default value to pop has to be specified not to end up with a KeyError
+                    msg.pop('flags', None)
             json.dump({'entries': messages_to_report, 'leapp_run_id': context}, f, indent=2)


leapp/reporting/__init__.py
```diff
@@ -326,6 +326,11 @@ def _create_report_object(entries):
     _sanitize_entries(entries)
     for entry in entries:
         entry.apply(report)
+    if Groups.INHIBITOR in report.get('groups', []):
+        # Keep the orig inhivitor functionality until we drop it completely
```
MichalHe, line 330:
```suggestion
        # Keep the original inhibitor functionality until we drop it completely
```
Idea: maybe the comment could provide insights how the inhibitor detection works  
```suggestion
        # Before removing Flags, the original inhibitor detection worked by checking the `flags` field; keep the `flags` field until we drop Flags completely
```

leapp/utils/report.py
```diff
@@ -139,4 +139,13 @@ def generate_report_file(messages_to_report, context, path, report_schema='1.1.0
                     groups = msg.pop('groups', [])
                     msg['flags'] = [g for g in groups if g in _DEPRECATION_FLAGS]
                     msg['tags'] = [g for g in groups if g not in _DEPRECATION_FLAGS]
+            if report_schema_tuple == (1, 2, 0):
```
MichalHe, line 142: Nitpick: imho checking for exact version of report schema might cause problems when we bump the schema version sooner than the flags/tags deprecation period has elapsed. However, I am not sure how often such version change is performed, and therefore, my concerns might not be necessary. 
```suggestion
            if report_schema_tuple >= (1, 2, 0):
```

leapp/utils/report.py
```diff
@@ -139,4 +139,13 @@ def generate_report_file(messages_to_report, context, path, report_schema='1.1.0
...
+                # DEPRECATED: flags, tags
+                # NOTE(pstodulk): This is a temporary solution to ensure that
+                # flags are not present in the JSON output as we need to re-introduce
+                # flags internally to ensure that any inhibitor checks will not
+                # be negatively affected until we drop flags and tags completely.
+                messages_to_report = list(messages_to_report)
```
MichalHe, line 148: Maybe I am too tired, but why is the conversion to list necessary?

We are iterating over it on the next line, however, we know that the collection is iterable, otherwise the conversion to list would fail. And we are not modifying the collection itself as we iterate, but rather the individual `msg` dictionaries, therefore, we should be OK without creating a list from the messages.

(Tested using the following snippet)
```python3
a_list = [{'flags': ['someflags']}]
for msg in a_list:
    msg.pop('flags')
```

MichalHe (commented):
I've just reviewed the changes, thus, I cannot approve it until I test it. However, looks good to me with a minor nitpicks.

leapp/utils/report.py
pirat89: Basically I chose this solution as we do not have any idea what will be in another version. We know only what's in the current one and previous versions. With the bump of the schema, I expect this function will be updated to cover the function properly. And there is high chance this piece of code will be removed before we introduce new schema. So I prefer to keep it as it is now.

leapp/reporting/__init__.py
pirat89: Good point!

leapp/utils/report.py
pirat89: It is not strictly necessary, the reason is mentioned here:
https://github.com/oamg/leapp/blob/567400d46a844d744b62fb938be81fe50ab9e342/leapp/utils/report.py#L131

Basically I am just keeping consistency with the rest of the code in the function.

MichalHe:
Tested the patch, works as expected. Builds used during testing:
- `leapp` - current master
- `leapp-PR781` - this PR
- `leapp-repository-old` - leapp repository without modified inhibition detection 
- `leapp-repository-new` - leapp repository with modified inhibition detection 

|      Correctly inhibits?    | `leapp`                        | `leapp-PR781`           |
|--------------------------------|------------------------------|------------------------------|
|`leapp-repository-old`    | :red_circle:                  | :heavy_check_mark: | 
|`leapp-repository-new`  | :heavy_check_mark:   | :heavy_check_mark: |

fernflower:
Please fix linter error.
```
************* Module leapp.reporting.__init__
leapp/reporting/__init__.py:330:0: C0301: Line too long (157/120) (line-too-long)
```

pirat89:
@fernflower fixed & squashed

leapp/utils/report.py
```diff
@@ -139,4 +139,13 @@ def generate_report_file(messages_to_report, context, path, report_schema='1.1.0
...
+                for msg in messages_to_report:
+                    msg.pop('flags')
```
fernflower, line 150: Your patch creates `flags` field only in case report has an inhibitor, but here you are popping `flags` from any message. I suspected this to crash and it really does in case of report-schema 1.2.0
```

  File "/usr/bin/leapp", line 11, in <module>

    load_entry_point('leapp==0.14.0', 'console_scripts', 'leapp')()

  File "/usr/lib/python3.6/site-packages/leapp/cli/__init__.py", line 43, in main

  File "/usr/lib/python3.6/site-packages/leapp/utils/clicmd.py", line 106, in execute

    args.func(args)

  File "/usr/lib/python3.6/site-packages/leapp/utils/clicmd.py", line 128, in called

    self.target(args)

  File "/usr/lib/python3.6/site-packages/leapp/cli/commands/upgrade/breadcrumbs.py", line 94, in wrapper

    return f(*args, breadcrumbs=breadcrumbs, **kwargs)

  File "/usr/lib/python3.6/site-packages/leapp/cli/commands/preupgrade/__init__.py", line 73, in preupgrade

    util.generate_report_files(context, report_schema)

  File "/usr/lib/python3.6/site-packages/leapp/cli/commands/upgrade/util.py", line 147, in generate_report_files

    generate_report_file(messages, context, report_json, report_schema)

  File "/usr/lib/python3.6/site-packages/leapp/utils/report.py", line 150, in generate_report_file

    msg.pop('flags')

KeyError: 'flags'
```
http://artifacts.osci.redhat.com/testing-farm/255b2826-8ee9-471f-881e-30b94fa85fef/

vinzenz: `dict.pop` and `dict.get` are annoying, pop behaves differently if no default parameter was passed.

fernflower (dismissed):
`msg.pop('flags')` should be changed to `msg.pop('flags', None)`

fernflower:
Tried adding a commit into your branch and got permission denied. Could you please allow it?
Or maybe just fix the issue I pointed out? 
```
diff --git a/leapp/utils/report.py b/leapp/utils/report.py
index 6248101..71a2f29 100644
--- a/leapp/utils/report.py
+++ b/leapp/utils/report.py
@@ -147,5 +147,7 @@ def generate_report_file(messages_to_report, context, path, report_schema='1.1.0
                 # be negatively affected until we drop flags and tags completely.
                 messages_to_report = list(messages_to_report)
                 for msg in messages_to_report:
-                    msg.pop('flags')
+                    # NOTE(ivasilev) As flags field is created only if there is an inhibitor a default value to pop
+                    # has to be specified not to end up with a KeyError
+                    msg.pop('flags', None)
             json.dump({'entries': messages_to_report, 'leapp_run_id': context}, f, indent=2)
```

pirat89:
@fernflower I cannot find out why you do not have the permission to commit into my fork, as regarding the github settings you should be able to do it. Let's experiment with that in future. I will for now update the commit to unblock it.

leapp/utils/report.py
pirat89: nice catch.

fernflower (approved):
Now the tests are fine.

PR #784, opened 2022-07-19
api: Preserve environment variables when calling a command

matejmatuska:
When calling a command with `_call` and `env` parameter is not `None`,
create and modify a copy of caller process environment variables and
call the command with the copy in order to preserve them.

Jira ref: OAMG-7203


vinzenz (approved):

vinzenz:
Very nice, thanks a lot!

PR #785, opened 2022-07-22
cli: Handle missing commands gracefully

vinzenz:
Previously when leapp was executed but no cli commands were installed,
the leapp execution failed with an exception.

This patch will add specific handling to this particular scenario and
return a 1 exit code.

Jira ref.: OAMG-7148

### DIFF
--- a/leapp/cli/__init__.py
+++ b/leapp/cli/__init__.py
@@ -1,11 +1,12 @@
-
 import os
 import pkgutil
 import socket
 import sys
+import textwrap
 
 from leapp import VERSION
 from leapp.cli import commands
+from leapp.exceptions import UnknownCommandError
 from leapp.utils.clicmd import command
 
 
@@ -40,4 +41,13 @@ def main():
 
     os.environ['LEAPP_HOSTNAME'] = socket.getfqdn()
     _load_commands(cli.command)
-    cli.command.execute('leapp version {}'.format(VERSION))
+    try:
+        cli.command.execute('leapp version {}'.format(VERSION))
+    except UnknownCommandError as e:
+        print(textwrap.dedent('''
+        Command "{CMD}" is unknown.
+        Most likely there is a typo in the command or particular leapp repositories that provide this command
+        are not present on the system.
+        You can try to install the missing content e.g. by the following command: `dnf install 'leapp-command({CMD})'`
+        '''.format(CMD=e.requested)))
+        sys.exit(1)
--- a/leapp/cli/__init__.py
+++ b/leapp/cli/__init__.py
@@ -44,10 +44,13 @@ def main():
     try:
         cli.command.execute('leapp version {}'.format(VERSION))
     except UnknownCommandError as e:
-        print(textwrap.dedent('''
-        Command "{CMD}" is unknown.
-        Most likely there is a typo in the command or particular leapp repositories that provide this command
-        are not present on the system.
-        You can try to install the missing content e.g. by the following command: `dnf install 'leapp-command({CMD})'`
-        '''.format(CMD=e.requested)))
+        bad_cmd = (
+            "Command \"{CMD}\" is unknown.\nMost likely there is a typo in the command or particular "
+            "leapp repositories that provide this command are not present on the system.\n"
+            "You can try to install the missing content e.g. by the following command: "
+            "`dnf install 'leapp-command({CMD})'`")
+        if e.requested.startswith('-'):
+            # A quick ack not to confuse users with install a leapp-command(--some-wrong-argument) suggestion
+            bad_cmd = "No such argument {CMD}"
+        print(bad_cmd.format(CMD=e.requested))
         sys.exit(1)
--- a/leapp/exceptions.py
+++ b/leapp/exceptions.py
@@ -102,6 +102,12 @@ class CommandDefinitionError(LeappError):
     pass
 
 
+class UnknownCommandError(LeappError):
+    def __init__(self, command):
+        super().__init__('Unknown command: {}'.format(command))
+        self.requested = command
+
+
 class LeappRuntimeError(LeappError):
     pass
 
--- a/leapp/utils/clicmd.py
+++ b/leapp/utils/clicmd.py
@@ -1,12 +1,12 @@
 import functools
 import os
 import sys
-
-from argparse import ArgumentParser, _SubParsersAction, RawDescriptionHelpFormatter
+from argparse import (ArgumentParser, RawDescriptionHelpFormatter,
+                      _SubParsersAction)
 
 import six
-
-from leapp.exceptions import CommandDefinitionError, UsageError, CommandError
+from leapp.exceptions import (CommandDefinitionError, CommandError,
+                              UnknownCommandError, UsageError)
 
 
 class _LeappArgumentParser(ArgumentParser):
@@ -23,6 +23,7 @@ class _LeappHelpFormatter(RawDescriptionHelpFormatter):
     """
     Capitalizes section headings in the help output
     """
+
     def start_section(self, heading):
         return super(_LeappHelpFormatter, self).start_section(heading.capitalize())
 
@@ -38,6 +39,7 @@ class _SubParserActionOverride(_SubParsersAction):
 
     The additional code will not be executed if python 2.7.9 or higher is found.
     """
+
     def __call__(self, parser, namespace, values, option_string=None):
         super(_SubParserActionOverride, self).__call__(parser, namespace, values, option_string)
         if sys.version_info >= (2, 7, 9):
@@ -54,6 +56,7 @@ class Command(object):
     """
     Command implements a convenient command-based argument parsing the framework.
     """
+
     def __init__(self, name, target=None, help='', description=None):  # noqa; pylint: disable=redefined-builtin
         """
         :param name: Name of the sub command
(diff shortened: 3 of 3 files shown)


leapp/cli/__init__.py
```diff
@@ -40,4 +41,11 @@ def main():
 
     os.environ['LEAPP_HOSTNAME'] = socket.getfqdn()
     _load_commands(cli.command)
-    cli.command.execute('leapp version {}'.format(VERSION))
+    try:
+        cli.command.execute('leapp version {}'.format(VERSION))
+    except UnknownCommandError as e:
+        print(textwrap.dedent('''
+        Command "{}" is unknown.
+        Did you forget to install leapp repositories which provide this command?
```
pirat89, line 49: @vinzenz @fernflower thinking about the msg, I think we should change it a little bit as when reading it, I can imagine that some users could feel "weird". Like: ,,How I forget? I installed leapp?!", if you know what I mean. What about something like:

```suggestion
        Hint: Possibly you did not installed particular leapp repositories which provide this command?
              Installing just `leapp`  the system get the framework and utility, not the content that is actually using
              leapp to perform a desired action.
```

fernflower: @pirat89 I really appreciate the move towards improving customer experience! How about we make it super clear what package need to be installed, like "Hint: Possibly you haven't installed the particular leapp repositories (leapp-upgrade-* package) which provide this command?"

pirat89: @fernflower I have been thinking about that, but as this is upstream project and leapp does not have to be used just for the upgrade, I would like to stay with a generic msg. I have been thinking also about something like: "e.g. try to install leapp-{command}", but then we would need to take care about: leapp-answer, leapp-preupgrade, ... it's doable, but not sure we want that.

What about to add something like: E.g. if you are interested about upgrading your system, try to look for leapp-upgrade, etc.

fernflower: Well, I myself as a user would always prefer clear instructions to the generic educational message. 
BTW what do you mean by "leapp does not have to be used just for the upgrade"? leapp won't be used for conversions by convert2rhel, are there any other usages but the upgrade?

pirat89: First, I understand your point on that and I am fighting with myself as I would like to rather write just`"You have not installed the leapp repositories for upgrade. run 'dnf install leapp-upgrade' first"` but I do not see it as the right solution.

It will not be used for conversions for now. The question whether it will be used in future or not has not been decided. This will depend on more stuff in future - e.g. whether some painpoints for other use-cases will be delivered. If we start to focus just on upgrades in this framework and ignore any other possible use-cases (no discussing now just conversions), we end again in situation to work on separation from upgrades if another use-case will be required in the end. If we start to ignore other posibilities, well in such a case we do not need even this PR, as we can just start to require always leapp-upgrade packages and that's it. This discussion we already had couple of times and we agreed to keep project separated and keep it as the upstream project for the framework, that could be used for whatever purposes by anyone. 

I prefer to provide a guidance how the content should be called, so in case that one day the framework is used e.g. for conversions, people will have idea they are searching probably for leapp-convert, etc. This is why we did all renames in packages in the last year.

I think we should stop thinking about leapp as just an upgrade tooling as it's pushing us back, unless we decide it's just the upgrade tooling.

pirat89: btw, discussed with @vinzenz , in case we agree on to provide instructions that will be ok just for rpm/dnf, we could do a simple change in leapp-repository to provide capabilities like `leapp-command(<command>)`, so we could point people to try to install e.g. `leapp-command(preupgrade)`, etc. It's quite simple change. Let's discuss it on cabal.

(P.S. dnf is doing the exactly the same stuff. when the command is not found, it suggests to install `dnf-command(<command>)`)

fernflower: That's a nice idea! However if we'd like to get this PR in before it's implemented I would still add a hint `For example, if you attempt to perform an upgrade, try searching for leapp repositories with dnf --showduplicates list leapp-upgrade*`.
Once the situation changes (leapp gets other usecases or we implement this easy install-leapp-command change, we can change the message).

pirat89: well, the implementation is matter of minutes in this case. it's just one line per cmd in leapp-repository specfile. if we want to go this way, implementation is not problem. but I want to be sure that team is ok to go in this way.

fernflower:
@pirat89 Sorry, being lazy here - how about we point the user to search for any `leapp-command(*)` package. And then, when they installs it and still invokes leapp with a non-existent command, we will let the framework+repositories do the job of telling them that it's not how leapp should be used?
```


    [root@ip-10-29-163-158 ~]# leapp dddd
     
    Command "dddd" is unknown.
    Most likely you are missing particular leapp repositories that provide this command.
    Try installing leapp-command(*) package first.
```
If they still invoke leapp with bad argument, the second invocation will be pretty informative
```
[root@ip-10-29-163-158 ~]# leapp ddd
usage: leapp [-h] [--version] command ...

Optional arguments:
  -h, --help  show this help message and exit
  --version   show program's version number and exit

Required arguments:
  command
    answer    Manage answerfile generation: register persistent user choices
              for specific dialog sections
    list-runs
              List previous Leapp upgrade executions
    preupgrade
              Generate preupgrade report
    rerun     Re-runs the upgrade from the given phase and using the
              information and progress from the last invocation of leapp
              upgrade.
    upgrade   Upgrade the current system to the next available major version.
error: argument command: invalid choice: 'ddd' (choose from 'answer', 'list-runs', 'preupgrade', 'rerun', 'upgrade')
```

pirat89:
```
# leapp dddd
     
Command "dddd" is unknown.
Most likely you are missing particular leapp repositories that provide this command.
Try installing leapp-command(*) package first.
```

@fernflower this could be confusing for users. They will not find the `leapp-command(*)` package as it does not exist. If they just copy paste that, it could have various outputs based on the current setup. So the minimum we should do is:
```
You can try to install the missing content by: `dnf install 'leapp-command(CMD)'`, replacing CMD by the desired leapp command.
```

fernflower:
@pirat89 Well, they can run `dnf install y leapp-command(*)` and they will get the leapp-repository package installed. Isn't it the whole idea?

pirat89:
@fernflower  let's discuss that. in both cases, the command is wrong as apostrophes/quotes are missing. Additionally, asterisk is not good as it will start to try to install all packages providing any leapp command. The idea is to install the right package providing the desired command. So no, I am taking the replacement of the asterisk by the particular command as good practice.

leapp/cli/__init__.py
```diff
@@ -40,4 +41,12 @@ def main():
...
     _load_commands(cli.command)
-    cli.command.execute('leapp version {}'.format(VERSION))
+    try:
+        cli.command.execute('leapp version {}'.format(VERSION))
+    except UnknownCommandError as e:
+        print(textwrap.dedent('''
+        Command "{0}" is unknown.
+        Most likely you are missing particular leapp repositories that provide this command.
+        Try installing leapp-command(*) package first.
+        '''.format(e.requested)))
```
pirat89, line 51:
```suggestion
        print(textwrap.dedent('''
        Command "{CMD}" is unknown.
        Most likely there is a typo in the command or particular leapp repositories that provide this command
        are not present on the system.
        You can try to install the missing content e.g. by the following command: `dnf install 'leapp-command({CMD})'`
        '''.format(CMD=e.requested)))
```

pirat89:
@vinzenz @fernflower so I have 1 good and 2 bad news.  The fix works as expected for the commands when there is not any command installed:
```
[root@localhost ~]# leapp upgrade

Command "upgrade" is unknown.
Most likely there is a typo in the command or particular leapp repositories that provide this command
are not present on the system.
You can try to install the missing content e.g. by the following command: `dnf install 'leapp-command(upgrade)'`
```

But it has different behaviour when any command is provided:
```
leapp ohoo
usage: leapp [-h] [--version] command ...

Optional arguments:
  -h, --help  show this help message and exit
  --version   show program's version number and exit

Required arguments:
  command
    answer    Manage answerfile generation: register persistent user choices
              for specific dialog sections
    list-runs
              List previous Leapp upgrade executions
    preupgrade
              Generate preupgrade report
    rerun     Re-runs the upgrade from the given phase and using the
              information and progress from the last invocation of leapp
              upgrade.
    upgrade   Upgrade the current system to the next available major version.
error: argument command: invalid choice: 'ohoo' (choose from 'answer', 'list-runs', 'preupgrade', 'rerun', 'upgrade')
```

And when no commands are provided, also CLI options are treated as commands:
```
[root@localhost ~]# leapp --oko

Command "--oko" is unknown.
Most likely there is a typo in the command or particular leapp repositories that provide this command
are not present on the system.
You can try to install the missing content e.g. by the following command: `dnf install 'leapp-command(--oko)'`
```

@vinzenz @fernflower  how we will deal with this?

fernflower:
@pirat89 Only a straightforward easy fix comes to my mind, please see the latest commit (we can definitely tune the error message wording, it is just an example).
```
[root@ip-10-29-163-91 ~]# leapp --noarg
No such argument --noarg
[root@ip-10-29-163-91 ~]# leapp nocmd
Command "nocmd" is unknown.
Most likely there is a typo in the command or particular leapp repositories that provide this command are not present on the system.
You can try to install the missing content e.g. by the following command: `dnf install 'leapp-command(nocmd)'`
``` 

Imho we have solved the original problem with this patch - there is no more ugly confusing traceback and the user actually gets pointed in the right direction.

pirat89 (approved):

PR #788, opened 2022-07-27
Deprecate Tags and Flags

fernflower:
Apply official deprecation process for Tags and Flags
primitives that has been superseeded by Groups.

OAMG-7262

### DIFF
--- a/docs/source/deprecation.md
+++ b/docs/source/deprecation.md
@@ -13,7 +13,11 @@ the leapp standard library, etc. But don't cover deprecated functionalities
 from particular leapp repositories (e.g. the [elt7toel8](https://github.com/oamg/leapp-repository/tree/master/repos/system_upgrade/el7toel8) leapp repository). For
 such information, see [Deprecated functionality in the el7toel8 repository](el7toel8/deprecation.html#deprecated-functionality-in-the-el7toel8-repository).
 
-*Note: No new deprecation present yet.*
+## current upstream development <span style="font-size:0.5em; font-weight:normal">(till the next release + 6months)</span>
+
+- Reporting primitives
+    - **`leapp.reporting.Flags`** - The `Flags` report primitive has been deprecated in favor of the more general `Groups` one.
+    - **`leapp.reporting.Tags`** - The `Tags` report primitive has been deprecated in favor of the more general `Groups` one.
 
 ## What is covered by deprecation process in leapp?
 
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -9,6 +9,7 @@ from leapp.compat import string_types
 from leapp.models import fields, Model, ErrorModel
 from leapp.topics import ReportTopic
 from leapp.libraries.stdlib.api import produce
+from leapp.utils.deprecation import deprecated
 
 
 class Report(Model):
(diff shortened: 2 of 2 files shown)


pirat89 (changes requested):
Document the deprecated entities here:
* https://leapp.readthedocs.io/en/latest/deprecation.html#list-of-deprecated-functionality-in-leapp

Here you can see how the deprecation looks like for leapp-repository:
* https://leapp.readthedocs.io/en/latest/el7toel8/deprecation.html#deprecated-functionality-in-the-el7toel8-repository

pirat89:
@fernflower  also, if I understand the impact right, this change requires update of all actors now as well. Otherwise users will see a lot of deprecation reports each run of leapp

fernflower:
> Document the deprecated entities here:
> 
>     * https://leapp.readthedocs.io/en/latest/deprecation.html#list-of-deprecated-functionality-in-leapp
> 
> 
> Here you can see how the deprecation looks like for leapp-repository:
> 
>     * https://leapp.readthedocs.io/en/latest/el7toel8/deprecation.html#deprecated-functionality-in-the-el7toel8-repository

Should be ok now https://github.com/oamg/leapp/blob/e2505c40e5a91ae6729e3564553fc3046242381d/docs/source/deprecation.md

docs/source/deprecation.md
```diff
@@ -13,7 +13,11 @@ the leapp standard library, etc. But don't cover deprecated functionalities
 from particular leapp repositories (e.g. the [elt7toel8](https://github.com/oamg/leapp-repository/tree/master/repos/system_upgrade/el7toel8) leapp repository). For
 such information, see [Deprecated functionality in the el7toel8 repository](el7toel8/deprecation.html#deprecated-functionality-in-the-el7toel8-repository).
 
-*Note: No new deprecation present yet.*
+## v0.14.0 <span style="font-size:0.5em; font-weight:normal">(till September 2022)</span>
```
pirat89, line 16: @fernflower  deprecation is usually 6m since the next release. this feature has not been deprecated with the release 0.14.0 but it will be deprecated with the next release.  so it's expected something like:

```suggestion
## current upstream development <span style="font-size:0.5em; font-weight:normal">(till the next release + 6months)</span>
```

the doc needs to be updated then with the new release PR

leapp/reporting/__init__.py
```diff
@@ -9,6 +9,7 @@
 from leapp.models import fields, Model, ErrorModel
 from leapp.topics import ReportTopic
 from leapp.libraries.stdlib.api import produce
+from leapp.utils.deprecation import deprecated
```
pirat89, line 12: if we are not going to change actors now, the code change must be skipped. otherwise actors need to be updated first. tbh, i prefer to update the actors. it's not so be deal a believe.

fernflower: I have been down this road (all actor code update with Tags/Flags->Groups transition) with previous POC and unfortunately it is a big deal and it's getting worse with each new actor pr created or merged :(
If we are to do this - I will definitely need help. It's not only about actors, it's also about tests, and it's not only dumb substitution, but refactoring as well.

fernflower: I don't want to rush code update to be honest, I'd rather hack our deprecation procedure a little to allow room for not-raising-a-warning-for-this-specific-model.

pirat89: @fernflower that would mean to update each our actor anyway to suppress the warning - which means we can already update actors directly. creating additional mechanism on top of what exists already does not make sense from my pov and more logically could be to drop the code change raising the warning. The code change seems pretty simple in actors so I would go with that. I am going to open the pr

pirat89: @fernflower I have opened the draft PR with the applied changes, so we could check how much stuff it affects in existing tests to have some data what else will be possibly needed on top of that change. See https://github.com/oamg/leapp-repository/pull/932

In case we find there a lot of problems around, I would drop the change in the code regarding the deprecation directives, so customer will not see bunch of deprecations warnings. In case we find cheap to do the change, I prefer to do it. wdyt about that?

pirat89: P.S. I know that in this case we will have to deal with mass-lint-fix as well. But when we are already in state that change majority of actors is inevitable, I think it's good time to deal with that as well.

pirat89:
@fernflower I rebased & squashed commits. I was missing blob to be able to test qith up-to-date master

pirat89:
With the old leapp-repository build, bunch of deprecation reports are generated, which means the deprecation delivered in this commit works as expected:
```
Risk Factor: medium
Title: Usage of deprecated class "Tags" at /etc/leapp/repos.d/system_upgrade/common/actors/unsupportedupgradecheck/actor.py:35
Summary: The primitive is deprecated as Tags and Flags have been joined into the Groups primitive.Please use Groups for report message typing instead.
Since: 2022-09-01
Location: /etc/leapp/repos.d/system_upgrade/common/actors/unsupportedupgradecheck/actor.py:35
Near:                 reporting.Tags([reporting.Tags.UPGRADE_PROCESS, reporting.Tags.SANITY]),

Key: 3a127f10711839a2fae566e69b05ab61c406f379
```

Jut for the completness, with the current upstream leapp-repository builds, no deprecation warning have been raised, so current changes in leapp-repository are correct as well.

pirat89 (approved):
lgtm and works well (tested manually)

pirat89:
Merging!! @fernflower  congratulation!! Happy to have this one long story around Flags/Tags finally finished after almost 2y :fireworks:  :champagne:

PR #789, opened 2022-07-28
Introduce a Blob model field

fernflower:
As models have to be JSON-serializable and byte
fields apparently don't fit in, let's introduce
a special model field type that could be used in
such situations.

OAMG-4306

### DIFF
--- a/leapp/models/fields/__init__.py
+++ b/leapp/models/fields/__init__.py
@@ -1,3 +1,4 @@
+import base64
 import copy
 import datetime
 import json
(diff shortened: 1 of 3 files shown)


vinzenz:
Well I'd have called it Blob or something, but well really I don't mind that name either.

However please actually extend the serialization tests, for this thanks.

fernflower:
@vinzenz Sure, will do, that's just a POC to see how it goes. Thanks for review!

Rezney (approved):

PR #790, opened 2022-08-03
Add a workaround around non-ascii and non-standard FQDNs

matejmatuska:
Add a wrapper around `socket.getfqdn()` which returns `invalid.hostname`
in cases where hostname, or rather FQDN contains non-ascii and
non-standard characters.

This works around the issue that non-ascii characters can't be inserted
into leapp database.

For the standard for hostnames see `hostname(7)`. The current
implementation doesn't check for length of hostname nor it's individual
segments.

Jira ref: OAMG-7102

### DIFF
--- a/leapp/utils/workarounds/fqdn.py
+++ b/leapp/utils/workarounds/fqdn.py
@@ -1,0 +1,18 @@
+import re
+import socket
+
+
+def decorate_getfqdn(fn):
+    def wrap(*args, **kwargs):
+        result = fn(*args, **kwargs)
+        # check whether FQDN conforms to standard, see HOSTNAME(7)
+        pattern = r"^(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])$"
+        if re.match(pattern, result, re.IGNORECASE):
+            return result
+        return 'invalid.hostname'
+
+    return wrap
+
+
+def apply_workaround():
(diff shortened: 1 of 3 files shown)


leapp/utils/workarounds/fqdn.py
```diff
@@ -0,0 +1,18 @@
+import re
+import socket
+
+
+def decorate_getfqdn(fn):
+    def wrap(*args, **kwargs):
+        result = fn(*args, **kwargs)
```
vinzenz, line 7: I think we should be careful here about the string type. In the case of the reported problem there were unicode characters in the hostname, so in that case `socket.getfqdn` might not return an instance of `str` but an instance of `unicode` @ python 2.7 - In that case there might be a problem with this code.

Also if something like this happens on python 3 it actually might instead return bytes, we need to validate that I guess.

WDYT?

vinzenz (changes requested):
We should ensure

 1. that we always get a `str` if we handle it that way with the regex. 
 2. This is quite testable, so please add some test(s) for it.

leapp/utils/workarounds/fqdn.py
matejmatuska: I just tested this and in both cases `str` was returned.

matejmatuska: Added some tests

vinzenz:
Looks good now, please rebase and squash all the commits into one. Then we can merge it.

pirat89 (approved):

PR #791, opened 2022-08-04
Add Makefile target for running unit tests in containers

matejmatuska:
Adds `test_container` Makefile target for running tests locally in
containers. This way tests can be run on all the different versions
locally.

Run `make test_container` to run tests in container. By default `rhel8`
container is used. This can be changed by setting `TEST_CONTAINER`
environment variable.

There is also `test_container_all` target which runs tests in all
available containers.

Available containers are rhel7, rhel8, rhel9.

Run `make clean_containers` to cleanup container images.

Containerfiles in `/res/container-tests/` are reused. This means that
`pip` installs test dependencies on every run. However this doesn't seem
like a significant slowdown. In future caching could be added at the
cost of creating different Containerfiles.

Building new image on every run also doesn't take too long thanks to
`podman`s caching.

Tests in containers use a different virtualenv `testenv`, which is
removed before running test to avoid conflicts between individual
containers.

Jira ref: OAMG-6965

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,9 @@ VERSION=`grep -m1 "^Version:" packaging/$(PKGNAME).spec | grep -om1 "[0-9].[0-9.
 _COPR_REPO=$${COPR_REPO:-leapp}
 _COPR_CONFIG=$${COPR_CONFIG:-~/.config/copr_rh_oamg.conf}
 
+# container used to run unit tests
+_TEST_CONTAINER=$${TEST_CONTAINER:-rhel8}
+
 # just to reduce number of unwanted builds mark as the upstream one when
 # someone will call copr_build without additional parameters
 MASTER_BRANCH=master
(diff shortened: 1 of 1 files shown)


pirat89:
@matejmatuska 
```
make test_container
Error: unknown flag: --env
make: *** [Makefile:153: test_container] Error 125
```

Makefile
```diff
@@ -132,11 +145,44 @@ else
...
+			export _CONT_FILE="res/container-tests/Containerfile.ubi9"; \
+			;; \
+		*) \
+			echo "Available test containers are rhel7, rhel8, rhel9"; \
+			exit 1; \
+			;; \
+	esac; \
+	export TEST_IMAGE="leapp-tests-$(_TEST_CONTAINER)"; \
+	rm -rf testenv/ && \
+	podman build -t $$TEST_IMAGE --env PYTHON_VENV=$$_VENV -f $$_CONT_FILE res/container-tests && \
```
pirat89, line 173: `podman build` does not know the `--env` nor `-e` option. at least the one I am using

```
$ podman version
Version:      3.4.7
API Version:  3.4.7
Go Version:   go1.16.15
Built:        Mon Jul 11 00:35:25 2022
```

matejmatuska: It is indeed a version thing. Seems like the option is not needed at all.

pirat89 (approved):
lgtm and works as expected \o/

PR #792, opened 2022-08-04
Add Makefile target for building RPMs locally and in containers

matejmatuska:
Add `build` target to build locally and `build_container` to build
locally in a container. Building in containers has the benefit of being
able to build on all the different platforms e.g. both on  el7 and el8.

Run `make build` to build locally.

Run `make build_container` to build in a container. Note that
BUILD_CONTAINER environment variable must be set.

Run `clean_containers` to clean up container images used for building.

The containers are built from Containerfiles in `res/container-builds/`.
Since Containerfiles for all the Fedora versions would differ only in the `FROM`
statement, a generic Containerfile `Containerfile.fedora_generic` is
used and the `FROM` statement is inserted in the Makefile.

For el7 builds CentOS 7 container is used, because rpm-build package
is not available in ubi7 image.

Builds cannot be run in parallel because containers operate on the same
files, which would result in them conflicting.

Also update `.gitignore` to exclude build files from git.

Jira ref: OAMG-2251

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,13 @@ help:
 	@echo "                         - available containers are: rhel7, rhel8, rhel9"
 	@echo "  test_container_all     runs linter and tests in all available containers"
 	@echo "  lint                   runs just the linter"
-	@echo "  clean_containers       cleans up all testing containers"
+	@echo "  build                  create the RPM"
+	@echo "  build_container        create the RPM in container"
+	@echo "                         - set BUILD_CONTAINER to select the container"
+	@echo "                         - available containers are: el7, el8, f35, f36, rawhide"
+	@echo "                         - this can't be used to build in parallel,"
+	@echo "                           as build containers operate on the same files"
+	@echo "  clean_containers      clean container images used for building"
 	@echo ""
 	@echo "Possible use:"
 	@echo "  make <target>"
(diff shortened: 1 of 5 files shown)


pirat89 (changes requested):
@matejmatuska `clean_containers` is defined twice, each with different functionality.
```
$ make clean_containers 
Makefile:238: warning: overriding recipe for target 'clean_containers'
Makefile:178: warning: ignoring old recipe for target 'clean_containers'
```

matejmatuska:
@pirat89 I must have merged incorrectly, it's fixed now.

Makefile
```diff
@@ -123,6 +130,50 @@ copr_build: srpm
...
+	*) \
+		echo "Available containers are el7, el8, f35, f36, rawhide"; \
+		exit 1; \
+		;; \
+	esac && \
+	echo "--- Preparing $$BUILD_CONTAINER container for building ---" && \
+	IMAGE_NAME="leapp-build-$${BUILD_CONTAINER}"; \
+	podman build -f res/container-builds/$$_CONT_FILE -t $$IMAGE_NAME res/container-builds/ && \
+	podman run --rm -ti -v $$PWD:/payload:Z --name "$${IMAGE_NAME}-cont" $$IMAGE_NAME && \
+	[ $${_CONT_FILE:0:1} = '.' ] && rm -f res/container-builds/$$_CONT_FILE
```
pirat89, line 175: @matejmatuska  If the condition result to false, the make commands ends with error and non-zero exit code which could be confusing.
```suggestion
	[ '$${_CONT_FILE:0:1}' = '.' ] && rm -f res/container-builds/$$_CONT_FILE || :
```

When fixed, we can merge it. Everything else seems to be working right.

pirat89 (changes requested):

pirat89 (approved):

PR #794, opened 2022-08-17
Be more specific about distro

fernflower:
Instead of vague distro=7 and distro=8 use
7.9 and 8.6 respectedly as our refactored regression
tests now use.
This should fix the "no plans found" issue with 8.6-rhui
upgrade as well as continue running e2e test as part of
8->9 integration tests.

### DIFF
--- a/.github/workflows/reuse-tests-7to8.yml
+++ b/.github/workflows/reuse-tests-7to8.yml
@@ -28,7 +28,7 @@ on:
         default: 'epel-7-x86_64'
       tmt_context:
         type: string
-        default: 'distro=rhel-7'
+        default: 'distro=rhel-7.9'
       compose:
         type: string
         default: 'RHEL-7.9-ZStream'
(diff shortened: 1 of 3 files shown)


fernflower:
Proof it works - https://github.com/fernflower/leapp/actions/runs/2874838042

Please let's wait for test results to make sure all e2e tests are picked up as part of integration tests.

fernflower:
Hmm something fishy going on with 8.6-rhui, please don't merge this

fernflower:
@MichalHe The failure with rhel-8.6-rhui tests timeout was not caused by this patch, but by infra issues (for some reason artemis can't provision vms from RHEL-8.6-rhui compose).
Anyway, imho this should not prevent us merging this PR as we will be just unifying test scheduling parameters with dev TF-* scripts.

pirat89 (approved):

PR #798, opened 2022-09-26
Add possibility to use docker to run tests and build in containers

matejmatuska:
On some systems podman is unavailable. Also, some people prefer docker over podman. Add possibility to use docker to run tests and build packages.

Set the CONTAINER_TOOL environment variable to change the container tool used for running tests and building packages, for example: CONTAINER_TOOL="docker" make test_container.

Unless you are running the Docker daemon in rootless mode, you will have to run as root. In such case files created in `packaging/` using `build_container` and in `testenv/` using `test_container` will be owned by root. To change the ownership back to your user, you can for example run `chown -R $USER packaging/ testenv/`.

The `clean_container` target now checks the existency of image before deleting it, otherwise the command would silently fail when ran without correct permission.

Jira ref.: OAMG-7286


pirat89 (approved):
expecting it works for docker as similar changes have been done in leapp-repository as well. Tested with podman manually:
- building with el8 container
- tests in default container

PR #799, opened 2022-10-06
Document workaround for the broken snactor

Jakuje:
prefer having this documented instead answering everyone questions why it does not work when it is in documentation. At least until it is fixed.

### DIFF
--- a/docs/source/el7toel8/actor-rhel7-to-rhel8.md
+++ b/docs/source/el7toel8/actor-rhel7-to-rhel8.md
@@ -327,6 +327,17 @@ data is stored in the database for the current session:
 ```shell
 snactor run --save-output IPUWorkflowConfig
 ```
+Since the leapp supports several upgrade paths, the `snactor` needs to know target system and upgrade flavor specification for correct execution.
+If you see similar errors when running `snactor`
+```
+leapp.models.fields.ModelViolationError: The value of "target" field is None, but this is not allowed
+```
+please, set the following environment variables (adjust the `LEAPP_UPGRADE_PATH_TARGET_RELEASE` as needed):
+```
+export LEAPP_UPGRADE_PATH_TARGET_RELEASE=8.6
+export LEAPP_UPGRADE_PATH_FLAVOUR=default
+snactor run --save-output IPUWorkflowConfig
+```
 
 ### Executing the whole upgrade workflow with the new actor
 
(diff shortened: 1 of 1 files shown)


docs/source/el7toel8/actor-rhel7-to-rhel8.md
```diff
@@ -328,6 +328,17 @@ data is stored in the database for the current session:
 snactor run --save-output IPUWorkflowConfig
 ```
 
+Because of a bug in leapp, this does not work and gives the following error:
```
fernflower, line 331: Technically it's not a bug or something can be fixed for good, as snactor is a development tool and is not being run on the machine-to-be-upgraded, so it has no knowledge of what system leapp is planned to be run for and thus needs this information to be passed this way ot another. Setting `LEAPP_UPGRADE_PATH_TARGET_RELEASE` and `LEAPP_UPGRADE_PATH_FLAVOR` to the values you mentioned is the way to say "Run leapp actors as you would for a 7to8 upgrade".

Could you please rephrase the description a bit? Maybe something like `After leapp started supporting several upgrade paths snactor requires target system and upgrade flavor specification for a correct execution. To get around the possible leapp ModelViolationError ... please set the following environment variables prior to running snactor`.

And thanks a lot for the patch, that's definitely an awesome addition to the dev docs!

Jakuje: Thanks for the feedback. I reworded the changed based on your suggestion.

Giving this another though, would it make sense to have this information somewhere inside of the `snactor`, when it is executed without these variables, reporting that it might not work for all the workflow related tasks, at least as a warning? Or even make it mandatory for execution of `snactor`? Or provide some user-friendly commandline switches for `snactor` to set this for you?

fernflower: You are definitely right, we should add this information to snactor.

docs/source/el7toel8/actor-rhel7-to-rhel8.md
```diff
@@ -327,6 +327,17 @@ data is stored in the database for the current session:
 ```shell
 snactor run --save-output IPUWorkflowConfig
 ```
+Since the leapp supports several upgrade paths, the `snactor` needs to know target system and upgrade flavor specification for correct execution.
```
pirat89, line 330:
```suggestion
Since the leapp upgrade repositories support several upgrade paths, the `snactor` needs to get additional data for the correct execution, like the release of the target system, the flavor specification, etc.
```

fernflower (approved):
Looks perfect now, thanks!

pirat89 (approved):
Same as Ina. Thanks!

Issue #692, opened 2021-03-02
Leapp doesn't log "real errors" / tracebacks


pirat89:
**Actual behavior**
If any actor crashes completely, leapp is not able to handle the issue nor log it (so without the console log it's not possible to see the error).

**To Reproduce**
Let's make any actor crash compltely and check leapp.db & /var/log/leapp/leapp-upgrade.log.
No info about the error is present inside.

**Expected behavior**
Crash of actor is processed by leapp and information is present in leapp.db & log files. 

Addtitional info:
- leapp-0.12.0-100.202102180840Z.8e3280a.master.el7_9.noarch

PR #802, opened 2022-11-09
Expose tracebacks from actor exceptions

prilr:
Previously when leapp would be running actors, any exceptions raised by
those actors in child processes would only be output to stderr,
while the main process would only have the exit code to examine.

Consequently, there was no real way to log occurring exceptions -
only providing the aforementioned exit code was possible.

This patch adds a pipe between the main and the actor processes,
through which a formatted exception + traceback string is provided to
the main process.
This string is then packaged into the `LeappRuntimeError` thrown from the
process, to be utilized downstream in
workflow/command code.

This should serve as a step to closing #692 - although actually logging
the exception traceback likely should be done on the `leapp-repository` side.

### DIFF
--- a/leapp/exceptions.py
+++ b/leapp/exceptions.py
@@ -109,7 +109,7 @@ class UnknownCommandError(LeappError):
 
 
 class LeappRuntimeError(LeappError):
-    def __init__(self, message, exception_info):
+    def __init__(self, message, exception_info=None):
         super().__init__(message)
         self.exception_info = exception_info
 
--- a/leapp/exceptions.py
+++ b/leapp/exceptions.py
@@ -104,13 +104,13 @@ class CommandDefinitionError(LeappError):
 
 class UnknownCommandError(LeappError):
     def __init__(self, command):
-        super().__init__('Unknown command: {}'.format(command))
+        super(UnknownCommandError, self).__init__('Unknown command: {}'.format(command))
         self.requested = command
 
 
 class LeappRuntimeError(LeappError):
     def __init__(self, message, exception_info=None):
-        super().__init__(message)
+        super(LeappRuntimeError, self).__init__(message)
         self.exception_info = exception_info
 
 
--- a/leapp/exceptions.py
+++ b/leapp/exceptions.py
@@ -109,7 +109,9 @@ class UnknownCommandError(LeappError):
 
 
 class LeappRuntimeError(LeappError):
-    pass
+    def __init__(self, message, exception_info):
+        super().__init__(message)
+        self.exception_info = exception_info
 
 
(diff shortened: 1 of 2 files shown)


pirat89:
@prilr  Thanks for the contribution! @vinzenz do you have a time?

pirat89:
@prilr also please take a look at the failed test
```

repository_dir = local('/tmp/pytest-of-root/pytest-0/repositories2/testing')

    def test_repo(repository_dir):
        setup_repo(repository_dir)
    
        def _run_test(repo_path):
            with repo_path.as_cwd():
                repository = find_and_scan_repositories(repo_path.dirpath().strpath)
                assert repository
                repository.load(resolve=True)
                assert getattr(leapp.tags, 'TestTag')
                assert repository.lookup_actor('TestActor')
                assert repository.lookup_workflow('TestWorkflow')
                assert not repository.lookup_workflow('MissingWorkflow')
                assert not repository.lookup_actor('MissingActor')
                assert repository.repos
                assert len(repository.serialize()) >= 1
                assert repository.actors
                assert not repository.topics
                assert not repository.models
                assert repository.tags
                assert repository.workflows
                assert repository.tools
                assert repository.libraries
                assert repository.files
                with pytest.raises(LeappRuntimeError):
                    repository.lookup_actor('TestActor')().run()
    
        p = Process(target=_run_test, args=(repository_dir,))
        p.start()
        p.join()
>       assert p.exitcode == 0
E       assert 1 == 0
E         -1
E         +0

tests/scripts/test_repository.py:112: AssertionError
```

leapp/exceptions.py
```diff
@@ -109,7 +109,9 @@ def __init__(self, command):
 
 
 class LeappRuntimeError(LeappError):
-    pass
+    def __init__(self, message, exception_info=None):
+        super().__init__(message)
```
vinzenz, line 113: This is not python2.7 compatible, which is required, that's why the test is failing
```suggestion
        super(LeappRuntimeError, self).__init__(message)
```

vinzenz (dismissed):
Please fix the `LeappRuntimeError` constructor

vinzenz:
Other than the constructor the PR LGTM, once it passes the tests I will give it a try myself.

prilr:
> Please fix the `LeappRuntimeError` constructor

Should've checked with Python 2 before submitting, sorry about that.
Went ahead and changed `UnknownCommandError`'s constructor to a compatible form while at it.

fernflower:
Wow, that's a great addition that has somehow slipped under the radar. I will try to take a look at it this week.

fernflower (approved):

PR #807, opened 2022-12-21
change the permissions for /var/lib/leapp directory from '755' to '700'

PeterMocary:
The nature of the content stored in the ``/var/lib/leapp`` directory requires stricter permissions.


fernflower (approved):
lgtm

fernflower (approved):

PR #812, opened 2023-02-16
Use upstream assets for tests

fernflower:
Upstream tests should be testing bleeding edge features using latest bleeding edge leapp data.
So let's pass LEAPPDATA_BRANCH=upstream as test env var.


fernflower:
Proof it works: https://github.com/fernflower/leapp/actions/runs/4195041798

pirat89 (approved):

PR #821, opened 2023-04-05
Improve handling of remediations with non-ascii data

fernflower:
When a remediation command is printed that manipulates non-ascii but totally valid utf-8 filenames, leapp should not give out any tracebacks upon report generation.

Related-to: OAMG-8629

### DIFF
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -238,30 +238,61 @@ class BaseRemediation(BaseListPrimitive):
         return ('detail', 'remediations')
 
 
+def _guarantee_encoded_str(text, encoding='utf-8'):
+    """Guarantees that result is a utf-8 encoded string"""
+    # Apply conversion only in py2 case
+    if six.PY3:
+        return text
+    # Unless data is already encoded -> encode it
+    try:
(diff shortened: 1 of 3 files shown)


leapp/reporting/__init__.py
```diff
@@ -245,6 +245,10 @@ def __init__(self, value=None):
         self._value = {'type': 'command', 'context': value}
 
     def __repr__(self):
+        # NOTE(ivasilev) As filenames can contain non-ascii characters let's deal with it properly.
```
pirat89, line 248: a nitpick, filenames can be confusing here. I would stick with "the string" or something generic. The issue does not stick just with filenames. It's general problem. The same in other notes.
 
```suggestion
        # NOTE(ivasilev) As the string can contain non-ascii characters let's deal with it properly.
```

pirat89 (approved):
@fernflower  I am little bit afraid of japannese systems, but I hope they also use UTF-8 (e.g. `Ja_jp.utf8`) instead of SJIS. I think it's still much better than before, so let's merge it and see if someone report problems in future about that. As UTF has been broken, SJIS has been broken too. So at least UTF should be covered now.

PR #823, opened 2023-04-14
Do not set COPR_CHROOTS for copr-build

fernflower:
Previous patch follow up: in case of leapp this
should not be specified.

OAMG-8876


fernflower:
Proof it works: https://github.com/fernflower/leapp/actions/runs/4699779277

pirat89 (approved):

PR #826, opened 2023-05-01
Dialog creation fails for a Component without choices

ygalblum:
`choices` is not an attribute of `Component` only of some of its subclasses

### DIFF
--- a/leapp/dialogs/dialog.py
+++ b/leapp/dialogs/dialog.py
@@ -54,7 +54,7 @@ class Dialog(object):
 
     @property
     def answerfile_sections(self):
-        return {"{}.{}".format(self.scope, c.key): c.choices for c in self.components}
+        return {"{}.{}".format(self.scope, c.key): getattr(c, "choices", []) for c in self.components}
 
     @property
     def min_label_width(self):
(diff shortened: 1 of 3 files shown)


ygalblum:
review please @oamg/developers

pirat89:
@ygalblum  thank you for the contribution! can you elaborate more on problems that the previous solution created? I mean, regarding the code it seems you hit some tracebacks, but I am curious what was the trigger actually as we haven't seen any so far. 

@fernflower  can you take a look?

ygalblum:
> can you elaborate more on problems that the previous solution created?

You can see the dialog I created here: https://github.com/ygalblum/leapp-supplements/blob/veritas-infoscale/repos/system_upgrade_supplements/el7toel8/actors/veritasinstallerpathfinder/actor.py#L19

Since `TextComponent` does not have a `choices` attribute, the code crashes in the line of my fix.

From a quick search I did, I couldn't find any actors that use the `TextComponent` and I guess it was missed in the tests.

fernflower:
Yeah, this very well might be the case as we are not using TextComponents. Could you please cover the failure with a unit test? It would really give us some peace of mind :)

ygalblum:
> Could you please cover the failure with a unit test?

Please see the second commit

fernflower (approved):
lgtm, thank you for the contribution! 
Scheduling regression tests

PR #827, opened 2023-05-04
Packit config for leapp

fernflower:


pirat89 (approved):
seems good and tests are running as expected!!! great work!!

PR #828, opened 2023-05-05
Generate empty report correctly

pirat89:
On Python2, the json.dumps library can produce either unicode or str output. Nowadays, the unicode is required, however in case the input for json.dumps function does not contain any unicode objects (e.g. when no report entry exists), it produces str type.

Relates: #821


pirat89:
@fernflower aah. copy&paste error /o\ should be fixed now

fernflower (approved):

PR #830, opened 2023-05-17
Make copr-build for /rerun 42 functional again

fernflower:
copr config username for a dependent leapp-repository PR build case wasn't changed in the 5ddce5a. This patch should address that.

OAMG-8876


fernflower:
Let's see if it helps https://github.com/fernflower/leapp/actions/runs/5001864021/jobs/8961105091

pirat89 (approved):

PR #832, opened 2023-06-07
Save actor tracebacks in leapp.db

fernflower:
Now when an exception is thrown during actor
execution it won't be shown only on stdout, but
will appear in leapp.db and leapp report as high
risk report message.

OAMG-4186

### DIFF
--- a/leapp/workflows/__init__.py
+++ b/leapp/workflows/__init__.py
@@ -340,7 +340,11 @@ class Workflow(with_metaclass(WorkflowMeta)):
                         instance.run()
                     except BaseException as exc:
                         self._unhandled_exception = True
-                        messaging.report_stacktrace(message=exc.message, trace=exc.exception_info, actorname=actor.name)
+                        messaging.report_stacktrace(message=exc.message,
+                                                    trace=exc.exception_info,
+                                                    actorname=actor.name)
+                        current_logger.error('Actor {actor} has crashed: {trace}'.format(actor=actor.name,
+                                                                                         trace=exc.exception_info))
                         raise
 
                     self._stop_after_phase_requested = messaging.stop_after_phase or self._stop_after_phase_requested
--- a/leapp/workflows/__init__.py
+++ b/leapp/workflows/__init__.py
@@ -338,8 +338,9 @@ class Workflow(with_metaclass(WorkflowMeta)):
                                      config_model=config_model, skip_dialogs=skip_dialogs)
                     try:
                         instance.run()
-                    except BaseException:
+                    except BaseException as exc:
                         self._unhandled_exception = True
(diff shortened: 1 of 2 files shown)


leapp/workflows/__init__.py
```diff
@@ -338,8 +338,9 @@ def run(self, context=None, until_phase=None, until_actor=None, skip_phases_unti
                                      config_model=config_model, skip_dialogs=skip_dialogs)
                     try:
                         instance.run()
-                    except BaseException:
+                    except BaseException as exc:
                         self._unhandled_exception = True
+                        messaging.report_stacktrace(message=exc.message, trace=exc.exception_info, actorname=actor.name)
```
pirat89, line 343: @fernflower I like it! However I am thinking whether we could even improve it from the point of adding nice error msg like: "The {name} leapp actor failed unexpectedly", putting the traceback in details. And being able to provide normal error output like in case `StopActorExecutionError`. I mean, instead of total hard crash of the whole framework and printing just raw output.

WDYT about that? Not saying to do this in this PR, just curious about your thoughts here. The code looks good to me. I want to try it later yet.

fernflower: It is exactly what will be reported :) The exc.message is holding the 'The {name} leapp actor failed unexpectedly' and exc.exception_info has all the traceback, as has been added by @prilr (many thanks btw!).
Here is the except from the report file:
```
Risk Factor: high
Title: Actor bacula_check unexpectedly terminated with exit code: 1
  File "/usr/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 74, in _do_run
    actor_instance.run(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/leapp/actors/__init__.py", line 289, in run
    self.process(*args)
  File "/etc/leapp/repos.d/system_upgrade/common/actors/baculacheck/actor.py", line 20, in process
    0 / 0
ZeroDivisionError: division by zero

Summary: 
Key: 640a2258f1032d584a2cad87a1a70a57448337c8
```

fernflower:
@PeterMocary @pirat89 Thanks for the review! 
If I understand you guys correctly, you would like to wrap the hard crash into a leapp exception and put the traceback in the details, right?
(I honestly prefer the hard crash way, as any actor crashing is a severe thing that has to be as obvious as possible, but I will implement it the way you suggest and we'll see :)

pirat89:
@fernflower @PeterMocary btw, I've realized that for error models in the report.txt file, we do not print that it's error, inhibitor, ... it looks just a high severity msg. I am thinking we should improve it in future somehow to make it clear that it's error, but I do not want to change the structure/schema because of that.

pirat89 (approved):

pirat89:
@fernflower thanks! lgtm and tests passed. Merging.

PR #833, opened 2023-06-07
Dialog Component - Add TextListComponent to support list of text

ygalblum:
Add a new Component to hold a list of strings with a minimal count
Implement a specific renderer hook
Load and dump the list from and to the answer file as a json
Add test cases for the dialog components
Add test cases for the answerstore

### DIFF
--- a/leapp/dialogs/components.py
+++ b/leapp/dialogs/components.py
@@ -57,6 +57,17 @@ class TextComponent(Component):
         renderer.render_text_component(self, dialog=dialog)
 
 
+class TextListComponent(Component):
+    """
+    TextListComponent is a list of text input component.
+    """
+    value_type = list
+    min_count = 1
+
+    def dispatch(self, renderer, dialog):
+        renderer.render_text_list_component(self, dialog=dialog)
+
+
 class PasswordComponent(TextComponent):
     """
     PasswordComponent is a text input component which will use non echoing input when possible (see getpass).
--- a/leapp/dialogs/renderer.py
+++ b/leapp/dialogs/renderer.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 from getpass import getpass
+import json
 import six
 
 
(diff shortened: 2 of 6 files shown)


ygalblum:
review please @oamg/developers

PeterMocary:
Hi, thank you for your contribution @ygalblum! :rocket: 

Could you please update the description with some additional information about the reason behind the changes?

pirat89:
@ygalblum @PeterMocary  we need to discuss this yet as the use case behind this feature is aside of the original purpose of Dialogs. Please do not do other changes until we discuss that. We are now affected by vacations so waiting for a time most of the team is back to discuss it.

Below sharing the use case behind the PR:
* An actor needs to get multiple URLs from a user. The number of URLs is not known in advance. So the point is to get all URLs that customer needs to be used for something and consumed by an actor.

The possible solution (that is nowadays more expected regarding the dialogs design) is to tell user to create a file under `/etc/leapp/files/` that will address such a problem. In this case actors can simply read the content of such a configuration file and work with it as needed.

ygalblum:
@pirat89 thanks for the detailed comment and yes, you are correct, this was the reason for this PR.

The reason I prefer this solution over your proposal is that I think that users will prefer to provide all the information in one place instead of multiple files. Don't think only about my actor. Once this becomes a known practice, what stops the next actor from getting answers in another file?

As for the way it was implemented, i.e. JSON inside an INI file, I looked around for how arrays are implemented in INI files and this was the most common and elegant way I found.

Having said that, because of your response (in the chat) and knowing that even if this change is merged, it will take time until it gets released, I decided to go with a hybrid solution. That is, the actor uses a JSON file to get the information it needs, but it does not force a location. Instead, it uses a Text dialog to get a path or URL to get the json from.

The advantage I saw in this solution is that it allows the user to maintain this JSON file next to where the files, to which the URLs are needed to being with, are located.

pirat89:
@ygalblum We have discussed this in the team and we agreed this is not way we want to go in future. We want to keep dialogs for purposes they were designed and do not want to extend them in this way. Also we do not want to mix multiple formats in one file - but this was just a minor fact for the decision as that part could be done differently. In this case we suggest to go the way I've proposed above - e.g.
* ask for a path to the file in the dialog
* just inhibit the upgrade and inform user to create the file on a specified path using the specified format

Closing, we are sorry about that but thanks for your time and the contribution.

PR #834, opened 2023-06-12
Fix invalid type check

dkubek:
The `isinstance` type check in `report_info` checks the type of an incorrect variable. This fixes the issue.


fernflower (approved):

MichalHe (approved):
Nice catch!

PR #836, opened 2023-07-17
stdlib.run: Fix missing executable

pirat89:
Currently, if the executable required in `run` does not exist (or is not executable), the child process's code is not replaced by `os.execvpe` function and it raises the OSError instead. However, the parent process does not get this OSError. It consumes exit code, stderr, ... of the child process. So in case the code does something like this:

    ```
    try:
        result = run(['non-executable'])
    except OSError:
        pass
    except CalledProcessError:
        # do something..
    ```

the child process passes and do whatever.. In case it ends with zero exit code, the obtained result is usually something totally different than expected in actors. Also there could be problems with non-idempotent code, when some actions could be done twice (once executed by the child, send time executed by the parent [current] process).

We have realized that number of existing leapp actors for in-place upgrades already count with the raise of OSError when executable cannot be used. So we choosed for now to check whether executables are present and raise OSError if not, so we are sure that only one process leave the function really.

Also applied another seatbelt into the child process - if the OSError is raised anyway despite our checks (e.g. SELinux prevents the execution) let's just kill the process instead giving it a possibility to continue. In such a case, always print a msg to stderr of the child process and exit with ecode 1.

**Note**: To check an executable we use `distutils.spawn.find_executable` which is deprecated in Python 3 and will be dropped in Python 3.12. However it exists now for Python 2 & 3, so we use this one for now and will replace it in future by `shutils.which` when the time comes.

## Additional changes:
* Make pylint happy (set noqa for E721 for the check the Fields is not
created directly). The "isinstance" function has a different behaviour
in this case than we want.

* Fix imports to make pylint happy

* Set `result` to empty dict instead of None: in case a ValueError
  or TypeError is raised inside the _call function, it's expected
  to copy the `result` content inside the final block, however in
  case the results is None, it fails and raise additional exception
  that covers the original one.

* Update unit tests to cover issues with missing executable

## Missing:
- [x] update unit-tests

### DIFF
--- a/leapp/libraries/stdlib/__init__.py
+++ b/leapp/libraries/stdlib/__init__.py
@@ -3,11 +3,11 @@
 represents a location for functions that otherwise would be defined multiple times across leapp actors
 and at the same time, they are really useful for other actors.
 """
+import base64
 import logging
 import os
 import sys
 import uuid
-import base64
 
 from leapp.exceptions import LeappError
 from leapp.utils.audit import create_audit_entry
@@ -167,6 +167,10 @@ def run(args, split=False, callback_raw=_console_logging_handler, callback_lineb
     :type stdin: int, str
     :return: {'stdout' : stdout, 'stderr': stderr, 'signal': signal, 'exit_code': exit_code, 'pid': pid}
     :rtype: dict
+    :raises: OSError if an executable is missing or has wrong permissions
+    :raises: CalledProcessError if the cmd has non-zero exit code and `checked` is False
+    :raises: TypeError if any input parameters have an invalid type
+    :raises: valueError if any of input parameters have an invalid value
     """
     if not args:
         message = 'Command to call is missing.'
@@ -174,7 +178,7 @@ def run(args, split=False, callback_raw=_console_logging_handler, callback_lineb
         raise ValueError(message)
     api.current_logger().debug('External command has started: {0}'.format(str(args)))
     _id = str(uuid.uuid4())
-    result = None
+    result = {}
     try:
         create_audit_entry('process-start', {'id': _id, 'parameters': args, 'env': env})
         result = _call(args, callback_raw=callback_raw, callback_linebuffered=callback_linebuffered,
@@ -191,6 +195,12 @@ def run(args, split=False, callback_raw=_console_logging_handler, callback_lineb
             result.update({
                 'stdout': result['stdout'].splitlines()
             })
+    except OSError:
+        # NOTE: currently we expect the result to be always set
+        # let's copy bash a little bit and set ecode 127
+        result = {'exit_code': '127', 'stdout': '', 'signal': 0, 'pid': 0}
+        result['stderr'] = 'File not found or permission denied: {}'.format(args[0])
+        raise
     finally:
         audit_result = result
         if not encoding:
--- a/tests/scripts/test_stdlib_call_audit.py
+++ b/tests/scripts/test_stdlib_call_audit.py
@@ -1,7 +1,18 @@
+import os
 import pytest
 import six
 
 from leapp.libraries.stdlib import run
+
+CUR_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+@pytest.fixture
+def adjust_cwd():
+    previous_cwd = os.getcwd()
+    os.chdir(os.path.join(CUR_DIR, "../"))
+    yield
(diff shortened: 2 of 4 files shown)


fernflower:
Can you add a unit test please? Currently the description reads pretty scary and not that easy to grasp, a repro can help explain what's it actually about imho

leapp/libraries/stdlib/__init__.py
```diff
@@ -175,8 +177,13 @@ def run(args, split=False, callback_raw=_console_logging_handler, callback_lineb
     api.current_logger().debug('External command has started: {0}'.format(str(args)))
     _id = str(uuid.uuid4())
     result = None
+    _path = env.get('PATH', None) if env else None
```
fernflower, line 180: Nit: `_path = (env or {}).get('PATH')` should give you the same result

leapp/libraries/stdlib/__init__.py
```diff
@@ -175,8 +177,13 @@ def run(args, split=False, callback_raw=_console_logging_handler, callback_lineb
...
     try:
         create_audit_entry('process-start', {'id': _id, 'parameters': args, 'env': env})
+        if not find_executable(args[0], _path):
+            #... just give something for audit - we count now to have always something inside...
+            result = {'exit_code': '1', 'stderr': 'File not found: {}'.format(args[0]), 'stdout': '', 'signal': 0, 'pid': 0}
```
fernflower, line 185: Hmm so you are setting result to this, but when is it going to be used?

pirat89: it's used in the final: statement. I was thinking originally about skipping it completely, but we already get used to having the result statement always available, so I decided to go with the way to keep it in this state. but it's discussable - cabal topic I (I will add it to the agenda in the morning). this is one of main reasons why postponed other tests and kept it in the draft state for now.

leapp/libraries/stdlib/__init__.py
pirat89: updated

pirat89:
@fernflower  seems the current impementatino of the fix is wrong as bunch of checks (expected by tests..) are executed inside the _call function

leapp/models/fields/__init__.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, default=None, help=None):  # noqa; pylint: disable=redefined-
         self._nullable = False
         self._default = default
 
-        if type(self) == Field:
+        if type(self) == Field:  # noqa;
```
fernflower, line 49: Hmm maybe `is` comparison could work here?

pirat89: let me try

pirat89: @fernflower  you were right!! thanks! updated

pirat89:
@fernflower btw, just realized the unit-tests is invalid. updating it now

tests/scripts/test_stdlib_call_audit.py
```diff
@@ -35,3 +46,24 @@ def test_no_encoding():
     cmd = ['echo', '-n', '-e', '\\xeb']
     result = run(cmd, encoding=None)
     assert isinstance(result['stdout'], six.binary_type)
+
+
+def test_missing_executable(monkeypatch, adjust_cwd):  # no-qa: W0613; pylint: disable=unused-argument
```
fernflower, line 51: Hmm are you actually using adjust_cwd? Can't follow the magic

pirat89: it's magic for me too. the same is used in unit-tests for leapp actors in this way. And I remember it replaced the original functionality. so it's either nonsense everywhere or it works somehow magically. I have to admit that dropping it I do not see any difference in test results here. I remember it was explained to me that just having this in the parameters is automagically used already by the pytest because it's decorated by `pytest.fixture`.

pirat89: ah, ok, so it really works. it's pytest magic around fixtures: https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization

fernflower (dismissed):
lgtm, thanks @pirat89

pirat89:
@fernflower rebased to fix the conflict. can you reapprove/rereview it and merge?

michalreznik (approved):

Rezney (approved):

PR #837, opened 2023-08-08
Run only destructive tier1 tests

fernflower:
Let's try to make the default test set smaller by running only destructive tests (no max_sst)


fernflower:
This will be covered by https://github.com/oamg/leapp/pull/827 , closing

PR #839, opened 2023-08-10
Make linter happy

matejmatuska:
Linter started giving new warnings on Python 3.9, let's make it happy
again.


leapp/models/fields/__init__.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, default=None, help=None):  # noqa; pylint: disable=redefined-
         self._nullable = False
         self._default = default
 
-        if type(self) == Field:
+        if type(self) == Field: # noqa: type-comparison - don't want to check for subclasses, can't use isinstance()
```
fernflower, line 49: Hmmm can't you use `is` for comparison here?

fernflower:
Thank you so much for finally resolving it!

leapp/models/fields/__init__.py
matejmatuska: You are right, you can and it is more appropriate, but both should work. I will change it.

fernflower: @pirat89 has already hit that and we briefly discussed this could be a solution https://github.com/oamg/leapp/pull/836/files#diff-c514bd4e364527bfcb2ee2e9cedcf24aa7e48de28e87df89057efa7c40cc3c66R49

fernflower (approved):
lgtm!

leapp/models/fields/__init__.py
matejmatuska: Looks like now the `noqa` isn't required anymore, way better.

PR #842, opened 2023-09-06
add external links to the leapp-report.txt

PeterMocary:
The external links were only included in the `leapp-report.json` and its `.txt` version was lacking this essential information. I didn't find any other problems regarding the information that is not included in one or the other format. Although the OAMG-9665 seems to be related to this. Do you think that it should be fixed in this PR as well?

Jira: OAMG-8553

### DIFF
--- a/leapp/utils/report.py
+++ b/leapp/utils/report.py
@@ -8,6 +8,7 @@ import six
 from leapp.reporting import (
     _DEPRECATION_FLAGS,
     _UPCOMING_DEPRECATION_FLAGS,
+    ExternalLink,
     Groups,
     Remediation,
     Severity,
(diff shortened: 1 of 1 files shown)


pirat89 (approved):
Works as expected. tested on several scenarios, IPU 7 -> 8 & 8 -> 9, including multiple external links in a report:
```
Risk Factor: high (inhibitor)
Title: Cannot perform the VDO check of block devices
Summary: [snip]
Related links:
    - Importing existing VDO volumes to LVM: https://red.ht/import-existing-vdo-volumes-to-lvm
    - yahuoo: https://red.ht/import-existing-vdo-volumes-to-lvm
    - bungabungbainkaaa: https://red.ht/import-existing-vdo-volumes-to-lvmKey: 429a99e13b19a7eebadbb8cb35233d8119bcf255
```

Also tried behaviour with non-ascii characters, which in such a case it's broken on RHEL 7, but we do not have handled non-ascii characters correctly anyway in other fields (e.g. title). The only handled part of the report for non-ascii characters is now the remediation text. regarding that I am merging this anyway. As this is going to be the last RHEL 7 release, it is not worthy to spent time on that and the problem is not present on RHEL 8+.

leapp/utils/report.py
```diff
@@ -115,6 +116,31 @@ def importance(message):
...
+    if remediation:
+        # NOTE(ivasilev) Decoding is necessary in case of python2 as remediation is an encoded string,
+        # while io.open expects "true text" input. For python3 repr will return proper py3 str, no
+        # decoding will be needed.
+        # This hassle and clumsiness makes me sad, so suggestions are welcome.
+        remediation_text = '\nRemediation: {}\n'.format(remediation)
+        if isinstance(remediation_text, six.binary_type):
+            # This will be true for py2 where repr returns an encoded string
+            remediation_text = remediation_text.decode('utf-8')
+            detail += remediation_text
```
pirat89, line 139: @PeterMocary  this is wrong. shame on me I missed that

pirat89: Should be fixed by #844

PR #845, opened 2023-10-17
Python-3.12 compat and refactor of path determination.

abadger:
In `stdlib._call()`, we lookup whether the command that we are going to call is present on the user's PATH.  Two changes are made to the function:

* Simplify the lookup of PATH now that we are making a copy of the environment's PATH.
* Add compat code to use `shutil.which()` on Python3 (while still using `distutils.spawn.find_executable()` on Python2).  As noted in the comment, `which()` will be needed when the code runs on Python-3.12.

### DIFF
--- a/leapp/libraries/stdlib/call.py
+++ b/leapp/libraries/stdlib/call.py
@@ -1,10 +1,17 @@
 from __future__ import print_function
 
-from distutils.spawn import find_executable
 import codecs
 import errno
 import os
 import sys
+try:
+    # shutil.which is available in Python 3.3 and above.
+    from shutil import which
+except ImportError:
+    # (pstodulk): find_executable() is from the distutils module which was
+    # removed in Python 3.12. We can get rid of this fallback when we drop
+    # support for Python 2. https://peps.python.org/pep-0632/
+    from distutils.spawn import find_executable as which
 
 from leapp.compat import string_types
 from leapp.libraries.stdlib.eventloop import POLL_HUP, POLL_IN, POLL_OUT, POLL_PRI, EventLoop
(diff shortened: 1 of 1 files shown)


leapp/libraries/stdlib/call.py
```diff
@@ -134,14 +141,9 @@ def _call(command, callback_raw=lambda fd, value: None, callback_linebuffered=la
...
 
-    _path = (env or {}).get('PATH', None)
-    # NOTE(pstodulk): the find_executable function is from the distutils
-    # module which is deprecated and it is going to be removed in Python 3.12.
-    # In future, we should use the shutil.which function, however that one is
-    # not available for Python2. We are going to address the problem in future
-    # (e.g. when we drop support for Python 2).
-    # https://peps.python.org/pep-0632/
-    if not find_executable(command[0], _path):
+    _path = environ.get('PATH', None)
```
pirat89, line 144: @abadger  cannot use environ here. it needs to respect the `env` input argument, which modifies set of envars. in this case keep `_path = (env or {}).get('PATH', None)`.

pirat89 (changes requested):
it makes completely sense. no idea why it didn't come to my mind originally. thanks :) just one liner change is needed

leapp/libraries/stdlib/call.py
abadger: <nod> I think that this is taken care of by Line 142:  https://github.com/oamg/leapp/pull/845/files#diff-e9be02d61bf13555898f98489e3edadaf8124a0c3eeb44a70813de2ed652ba03R142

If I'm reading the code right, the current logic of line 137 is:
* Look in the input `env` argument for `PATH`.
* If it exists, use that.
* Otherwise, get `PATH` from `os.environ` (inside of `find_exectuable()`).

The new logic is:
* Set `environ` to be a copy of `os.environ`.
* Replace any values in it with the values from the input `env` argument (including `PATH`).
* Then use `PATH` from the **modified** `environ`.

I think that makes them equivalent.  But I could be missing something?

pirat89: aah. you are right! I am sorry, I overlooked that completely that this is not `os.environ` but already updated `environ` variable specified in the function. yes, it make sense than and it should be in such a case equivalent. thanks :)

pirat89 (approved):

leapp/libraries/stdlib/call.py
abadger: No problem at all!  I'm always pleased when someone is double checking that my logic is correct :-)

PR #846, opened 2023-10-17
Recommend descriptive variable name for discarded values.

abadger:
* We tell people not to use `_`.  Now we advise them on what to use instead.
* Also fix link syntax in the contribution guide.


pirat89 (approved):
Thanks! lgtm

PR #847, opened 2023-11-21
Extend information from leapp saved to leappdb

dkubek:
Extend leappdb for leapp-inspector

This commit combines multiple additions to the leapp database.

First addition is tracking of entity metadata. The `Metadata` model stores the metadata of entities such as `Actor` or `Workflow`. This data is stored in a new table `metadata` of the `leapp.db` file.

  1. metadata of *discovered* actors.

     For an actor, the metadata stored contain:

        `class_name`    - the name of the actor class
        `name`          - the name given to the actor
        `description`   - the actor's description
        `phase`         - phase of execution of the actor
        `tags`          - names of any tags associated with an actor
        `consumes`      - list of all messages the actor consumes
        `produces`      - list of all messages the actor produces
        `path`          - the path to the actor source file

  2. workflow metadata.

     For a workflow, the metadata stored contain:

       `name`           - name of the workflow
       `short_name`     - short name of the workflow
       `tag`            - workflow tag
       `description`    - workflow description
       `phases`         - all phases associated with the workflow

Next addition is tracking of dialog question. Previously leapp was not able to detect the actual question asked from the user as it could be generated dynamically when actor is called and depend on the configuration of the user's system.

Last addition includes storing the actor exit status. Exit status is now saved as an audit event `actor-exit-status`. Exit status 0 represents successful execution or `StopActorExecution`/`StopActorExecutionError`, while 1 indicates an unexpected and unhandled exception.

These changes collectively improve the metadata handling capabilities of, ensuring accurate storage and retrieval of essential information for various entities.


leapp/workflows/__init__.py
```diff
@@ -332,6 +337,19 @@ def run(self, context=None, until_phase=None, until_actor=None, skip_phases_unti
...
+                        'consumes': [model.__name__ for model in metadata.get('consumes', ())],
+                        'produces': [model.__name__ for model in metadata.get('produces', ())],
+                        'tags': [tag.__name__ for tag in metadata.get('tags', ())],
+                        'dialogs': [dialog.serialize() for dialog in metadata.get('dialogs', ())],
+                        'phase': phase[0].name,
+                    })
+                    md = Metadata(kind='actor', name=actor.name, context=context,
+                                  hostname=os.environ['LEAPP_HOSTNAME'], metadata=metadata)
+                    md.store()
+
```
pirat89, line 352: @dkubek IIUC, this would be problematic. It stores the metadata about actors in time they are are going to be executed, instead of time when they are discovered. With that, we could just say what actors have been executed, but cannot say for sure whether all actors have been already executed.

I am also thinking yet about dialogs, as now they are hidden between all metadata. However, we would like to be able to work with dialogs in future in reports (or something like taht), so they could be shared with webUI applications (RH Insights, cockpit, foreman/satellite). I need to check what exact information is stored in this case now, but I can imagine we would like to have stored dialoges also somewhere else. My idea is to have a possibility to get list of produced dialogs. In this case, if I get it right, we have information about all dialogs defined in actors, but we will need information about actually asked dialogs.

I need to play around a little bit yet. But sharing my current ideas already.

res/schema/audit-layout.sql
```diff
@@ -43,6 +43,16 @@ CREATE TABLE IF NOT EXISTS message (
...
 
 
+CREATE TABLE IF NOT EXISTS metadata (
+  id                INTEGER PRIMARY KEY NOT NULL,
+  context           VARCHAR(36)         NOT NULL REFERENCES execution (context),
+  kind              VARCHAR(256)        NOT NULL DEFAULT '',
+  name              VARCHAR(1024)       NOT NULL DEFAULT '',
+  metadata          TEXT                         DEFAULT NULL,
+  UNIQUE (context, kind, name)
+);
```
pirat89, line 53: @dkubek  just thinking, could we obtain easily information about the file in which the actor is defined? adding the `path` column (or having it part of metadata?) would be potentially beneficial also in some cases.

pirat89: @fernflower not sure whether I understand it right. what do you mean by function of interest? do you mean the actor definition? that should be present always in the actor.py file. But I dunno whether the framework has a high requirement for that. Speaking about other functions and libraries, we are not interested about them here. We want to know only where the actor is defined.

pirat89: just to add a reason why I am thinking about having the path:
* actor name and directory names differs (and can differ completely)
* we do not have safe namespaces between repositories. it's possible to have 2 actors of the same name on 2 different paths and execute them. in this case, having information about that could be valuable as we will know that there is something fishy

for the experiment, it's possible to `cp -a <existing-actor> <my-existing-actor>`. now, let's imagine someone creates a generic actor name like "read_configuration". Then we are getting to dark waters where we have ambiguous actor names.

pirat89: not saying it's critical to have it, just that it could help us to detect such problems if happens.

fernflower: You are right, I thought it's impossible to have 2 actors with the same name. Then let's add path to actor file.

dkubek: @pirat89 @fernflower Actually, the ``path`` is already present in the metadata field. Sorry, I should have been more explicit in what the metadata looks like. Here is a sample from a random actor:

```json
{
  "class_name": "ScanSystemdSource",
  "path": "/usr/share/leapp-repository/repositories/system_upgrade/common/actors/systemd/scansystemdsource",
  "name": "scan_systemd_source",
  "tags": [
    "IPUWorkflowTag",
    "FactsPhaseTag"
  ],
  "consumes": [],
  "produces": [
    "SystemdServicesInfoSource",
    "SystemdServicesPresetInfoSource",
    "SystemdBrokenSymlinksSource"
  ],
  "dialogs": [],
  "description": "\n    Provides info about systemd on the source system\n\n    The provided info includes information about:\n    - vendor presets of services\n    - systemd service files, including their state\n    - broken systemd symlinks\n\n    There is an analogous actor :class:`ScanSystemdTarget` for target system.\n    ",
  "apis": [],
  "phase": "FactsCollection"
}
```

I don't think having a separate column for ``path`` is necessary:

1. Currently, the ``metadata`` table stores information about different kinds of objects (differentiated by the ``kind`` column). Concretely, I am saving also workflow metadata, which does not have any path associated with it. So the ``path`` column would be only meaningful for some entries, which is something I am not that big of a fan of doing if I can avoid it
2. I don't see a use case right now where I would need to query metadata given an actor path, so it would be redundant in this way. But definitely correct me if I'm wrong and miss something.

However, any application needing the path of an actor (given actor name) would have to parse it from the json, which _is_ an inconvenience, however, we have to parse the metadata as json later anyway, so I am not sure whether it has too much value to have this one specific information separate.

One thing however, I would like your opinion on is whether to make the data that is stored more explicit in the source code. Seeing that you both missed that the ``path`` is already being stored, means I have done a very poor job of telegraphing what data is actually being saved. Do you think it would be better to manually list all the fields I am storing in the metadata column?

leapp/workflows/__init__.py
dkubek: @pirat89 I was thinking about the point where should I save the actor. It seemed kind of unnecessary to store the information about actors which are not being executed and just have the information about what _has_ been executed. However, I think I now see that having even the information about discovered actors is valuable for debugging. And the information about executed actors can be always found in the audit. Will rework it.

As for the dialogs, I am aware that there is information missing and I have actually not come tome implementing that yet. So far I've been just puzzling over how they work so it's still WIP

res/schema/audit-layout.sql
pirat89: @dkubek  thank you for detailed explanation. I agree with you that the separate column is not needed in such a case. Having it as a part of metdata is completely ok, especially in case we have already the data there :)

pirat89: Can you put info about what is now stored inside the metadata as part of the commit msg? I think that listing the keys should be ok.

res/schema/audit-layout.sql
```diff
@@ -43,6 +43,24 @@ CREATE TABLE IF NOT EXISTS message (
 );
 
 
+CREATE TABLE IF NOT EXISTS metadata (
+  id                INTEGER PRIMARY KEY NOT NULL,
+  context           VARCHAR(36)         NOT NULL REFERENCES execution (context),
+  kind              VARCHAR(256)        NOT NULL DEFAULT '',
+  name              VARCHAR(1024)       NOT NULL DEFAULT '',
+  metadata          TEXT                         DEFAULT NULL,
```
vinzenz, line 51: Seeing this and also the dialog data, I think you might want to take an inspiration here on the message_data and deduplicate it.
You would just store a hash of the data here, as this will be stored multiple times and every execution adds the same data here every time.

res/schema/migrations/0002-add-metadata-dialog-tables.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE TABLE IF NOT EXISTS entity (
+  id                INTEGER PRIMARY KEY NOT NULL,
+  context           VARCHAR(36)         NOT NULL REFERENCES execution (context),
+  kind              VARCHAR(256)        NOT NULL DEFAULT '',
+  name              VARCHAR(1024)       NOT NULL DEFAULT '',
+  metadata_hash     VARCHAR(64)         NOT NULL REFERENCES metadata (hash),
+  UNIQUE (context, kind, name)
+);
+
+CREATE TABLE dialog (
```
vinzenz, line 17: IF NOT EXISTS as well, to ensure avoiding problems, when it already exists

pirat89:
rebased

res/schema/migrations/0002-add-metadata-dialog-tables.sql
pirat89: good catch. updated

leapp/workflows/__init__.py
pirat89: resolved! \o/

pirat89:
@dkubek seems good to me :100:   I will do some testing yet on rhel8 to be sure and I think we can squash&merge it.

pirat89 (approved):
seems to be working like a charm. Manually tested:
* (pre)upgrade 7 -> 8 & 8 -> 9
* update of existing leapp.db to the new format (7, 8)
  * run older leapp, install 847 PR build, run it again, check leapp.db
* run downgraded leapp (released one) after new leapp.db has been created already

In all cases it worked without problems.

I've been thinking whether the `StopActorExecutionError` should be handled as success or fail in the leapp.db as this one is honestly trickey. People who works with leapp.db directly, could be confused by this, but I understand reasoning for this logic. I think it's ok to go this way and in case we realize later we would like to change the behaviour, it should not be problematic to do it any time. Especially before the next upstream release. So merging to enable works on other features that are waiting for this PR.

PR #849, opened 2024-01-26
Add RHUI-8.8 and RHUI-8.9 tests

fernflower:
Having extended coverage for leapp-repository tests we just can't leave leapp behind.


pirat89 (approved):
@fernflower  seems everything works as expected. LGTM, but I am want to postpone the merge until we sync to be sure we will not have crashing tests in couple of weeks due to some upcoming changes - or to sync how we will deal with it.

fernflower:
If we have crashing tests in 2 weeks - let's deal with it in 2 weeks imho and disable in the worst case. But ok, let's sync.

PR #851, opened 2024-02-22
Add process locking

dkubek:
This change addresses the potential risk of running multiple instances of Leapp simultaneously on a single system. 

Considerations:
- Insights Tasks (preupgrade and upgrade) currently allow multiple executions on the same system
- The absence of a filter/tooltip in the Insights user interface makes it challenging to determine if a task is already running on a system.

A simple lock mechanism to prevent concurrent executions using a BSD lock has been implemented (see [1](http://gavv.net/articles/file-locks/#bsd-locks-flock)).

JIRA: https://issues.redhat.com/browse/OAMG-9827

### DIFF
--- a/leapp/config.py
+++ b/leapp/config.py
@@ -40,6 +40,9 @@ _CONFIG_DEFAULTS = {
         'dir': '/var/log/leapp/',
         'files': ','.join(_FILES_TO_ARCHIVE),
     },
+    'lock': {
+        'path': '/var/run/leapp.pid'
+    },
     'logs': {
         'dir': '/var/log/leapp/',
         'files': ','.join(_LOGS),
--- a/leapp/exceptions.py
+++ b/leapp/exceptions.py
@@ -148,3 +148,7 @@ class RequestStopAfterPhase(LeappError):
 
     def __init__(self):
         super(RequestStopAfterPhase, self).__init__('Stop after phase has been requested.')
+
+
+class ProcessLockError(LeappError):
+    """ This exception is used to represent an error within the process locking mechanism. """
--- a/leapp/utils/lock.py
+++ b/leapp/utils/lock.py
@@ -1,0 +1,83 @@
+import os
+import fcntl
+import logging
+
+from leapp.config import get_config
+from leapp.exceptions import ProcessLockError
+
+
+def leapp_lock(lockfile=None):
+    return ProcessLock(lockfile=lockfile)
+
+
+def _acquire_lock(fd):
+    try:
+        fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        return True
+    except OSError:
+        return False
+
+
+def _clear_lock(fd):
+    os.lseek(fd, 0, os.SEEK_SET)
+    os.ftruncate(fd, 0)
+
+
+def _read_pid(fd):
+    return os.read(fd, 20)
+
+
+def _write_pid(fd, pid):
+    _clear_lock(fd)
+    os.write(fd, str(pid).encode('utf-8'))
+
+
+class ProcessLock(object):
+
+    def __init__(self, lockfile=None):
+        self.log = logging.getLogger('leapp.utils.lock')
+        self.lockfile = lockfile if lockfile else get_config().get('lock', 'path')
+
+        self.fd = None
+
+    def _get_pid_from_lockfile(self):
+        running_pid = _read_pid(self.fd)
+        self.log.debug("_get_pid_from_lockfile: running_pid=%s", running_pid)
+        running_pid = int(running_pid)
+
+        return running_pid
+
+    def _try_lock(self, pid):
+        if not _acquire_lock(self.fd):
+            try:
+                running_pid = self._get_pid_from_lockfile()
+            except ValueError:
+                process_msg = ''
+            else:
+                process_msg = ' by process with PID {}'.format(running_pid)
+
+            msg = (
+                'Leapp is currently locked{} and cannot be started.\n'
+                'Please ensure no other instance of leapp is running and then delete the lockfile at {} and try again.'
+            ).format(process_msg, self.lockfile)
+            raise ProcessLockError(msg)
+
+        try:
+            _write_pid(self.fd, pid)
+        except OSError:
+            raise ProcessLockError('Could not write PID to lockfile.')
+
+    def __enter__(self):
+        my_pid = os.getpid()
+
+        self.fd = os.open(self.lockfile, os.O_CREAT | os.O_RDWR, 0o600)
+        try:
+            self._try_lock(my_pid)
+        except ProcessLockError:
+            os.close(self.fd)
+            raise
+
+    def __exit__(self, *exc_args):
+        _clear_lock(self.fd)
+        os.close(self.fd)
+        os.unlink(self.lockfile)
--- a/leapp/cli/__init__.py
+++ b/leapp/cli/__init__.py
@@ -2,12 +2,12 @@ import os
 import pkgutil
 import socket
 import sys
-import textwrap
 
 from leapp import VERSION
 from leapp.cli import commands
-from leapp.exceptions import UnknownCommandError
+from leapp.exceptions import UnknownCommandError, ProcessLockError
 from leapp.utils.clicmd import command
+from leapp.utils.lock import leapp_lock
 
 
 @command('')
@@ -42,7 +42,8 @@ def main():
     os.environ['LEAPP_HOSTNAME'] = socket.getfqdn()
     _load_commands(cli.command)
     try:
-        cli.command.execute('leapp version {}'.format(VERSION))
+        with leapp_lock():
+            cli.command.execute('leapp version {}'.format(VERSION))
     except UnknownCommandError as e:
         bad_cmd = (
             "Command \"{CMD}\" is unknown.\nMost likely there is a typo in the command or particular "
@@ -53,4 +54,7 @@ def main():
             # A quick ack not to confuse users with install a leapp-command(--some-wrong-argument) suggestion
             bad_cmd = "No such argument {CMD}"
         print(bad_cmd.format(CMD=e.requested))
+        sys.exit(1)
+    except ProcessLockError as e:
+        sys.stderr.write('{}\nAborting.\n'.format(e.message))
         sys.exit(1)


fernflower:
(not a real review yet) Could you please rebase? The tests should be fixed by this

leapp/utils/lock.py
```diff
@@ -0,0 +1,98 @@
...
+                       'Please ensure all previous instances of the application are closed and try again.')
+                raise ProcessLockError(msg)
+
+            old_pid = _read_pid(fd)
+            if len(old_pid) == 0:
+                # No pid in lockfile
+                _write_pid(fd, pid)
+                return pid
+
+            old_pid = _validate_pid(old_pid)
```
fernflower, line 68: I was checking some corner cases: currently if you modify the PID stored in leapp file a parallel leapp run execution would be allowed, and in the end both runs will be terminated with the following trace
```
  File "/usr/bin/leapp", line 11, in <module>
    load_entry_point('leapp==0.17.0', 'console_scripts', 'leapp')()
  File "/usr/lib/python3.6/site-packages/leapp/cli/__init__.py", line 46, in main
  File "/usr/lib/python3.6/site-packages/leapp/utils/lock.py", line 98, in __exit__
    os.unlink(self.lockfile)
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/leapp/leapp_lock.pid'
```

Maybe we could change that? To me it feels like if there is already a lock file present regardless of contents - no other leapp run should be allowed at least. WDYT?

dkubek: I see your point. The reason the PID is stored within the lock file is that leapp could have been terminated with a SIGKILL previously, which would not remove the lock file, but the idea here was that leapp is not in fact running and should start normally.

What you are suggesting is, I am guessing, that leapp would not run in this case and we would prompt the user to delete the lock file manually.

The main disadvantage I see that in case someone creates the lock file after running ``leapp upgrade`` but before reboot (suppose we take this into account as we are already considering someone externally modifying the lock file during execution; the current mechanism should not corrupt the file), leapp would fail in initram beause it detects an existing lock file. The current approach would at least check whether a process with the PID in it is running and proceed.

Overall I think there is not much we can guarantee if someone externally modifies the lock, but the way it is there is at least less of a chance we fail in initram because of it, so I would propose to keep it as is.

leapp/config.py
```diff
@@ -40,6 +40,9 @@
         'dir': '/var/log/leapp/',
         'files': ','.join(_FILES_TO_ARCHIVE),
     },
+    'lock': {
+        'path': '/var/lib/leapp/leapp_lock.pid'
+    },
```
vinzenz, line 45: This will cause the lock file to moved on the first execution right away...

https://github.com/oamg/leapp-repository/blob/214ed9b57c5e291cda5ff6baf7c7a790038fef34/commands/upgrade/__init__.py#L81

vinzenz (commented):
You have to overthink the lock cleanup design here

vinzenz:
I am not really sure, right now if it would work for leapp but did you consider open the SQLite db exclusively?

vinzenz:
> I am not really sure, right now if it would work for leapp but did you consider opening the SQLite db exclusively?

Or rather a SQLite db, that will tell you that you can't use it 😁

leapp/utils/lock.py
MichalHe: If someone touches leapp internals (e.g. modifies the lock file, or creates a lock file after execution) then it is their problem what happens (e.g. leapp fails in initramfs). Although these concerns can be hypothetically materialized, I do not think that we should spend time discussing them.

As for leaving out a PID file after a SIGKILL -- same applies. The user has opted in to killing leapp in a way that it cannot clean after itself. So using SIGKILL semantically involves acknowledging the fact that there will be mess on the system. As long as the lock file can be removed manually afterwards, everything is fine.

leapp/utils/lock.py
```diff
@@ -0,0 +1,98 @@
...
+        self.log = logging.getLogger('leapp.utils.lock')
+        self.lockfile = lockfile if lockfile else get_config().get('lock', 'path')
+
+    def _try_lock(self, pid):
+        fd = os.open(self.lockfile, os.O_CREAT | os.O_RDWR, 0o644)
+
+        try:
+            if not _acquire_lock(fd):
+                msg = ('Leapp is currently locked by another process and cannot be started.\n'
+                       'Please ensure all previous instances of the application are closed and try again.')
```
abadger, line 59: I think this should instruct the user to delete the lockfile once they've verified that no other leapp instances are running?

leapp/config.py
abadger: Combining this with the idea that the lockfile should not survive reboot, perhaps adding it to `/run` and letting it be cleared on reboot would be appropriate?  (I believe that /run is usually a tmpfs nowadays so that would protect it.  I don't recall if systemd also has a way to clear out stale lockfiles on boot if /run happens to not be a tmpfs.)

leapp/utils/lock.py
abadger: Although I agree with modification of leapp internals pushing responsibility onto the user, I think dealing with a stale lockfile in the middle of an upgrade, is important.  Bugs, things outside of the user's conscious control on the system, etc can lead to stale lockfiles.  While it's somewhat standard for users to have to clean up those lockfiles if necessary when invoking a program manually from the command line, forcing the user to clean up a stale lockfile themselves while leapp is supposed to be handling the next steps of the upgrade automatically feels like it will cause the user quite a bit of anxiety that it would be best for us to avoid.

leapp/config.py
vinzenz: /run is actually the place it should be at

leapp/utils/lock.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class ProcessLock(object):
+
+    def __init__(self, lockfile=None):
+        self.log = logging.getLogger('leapp.utils.lock')
+        self.lockfile = lockfile if lockfile else get_config().get('lock', 'path')
+
+    def _try_lock(self, pid):
+        fd = os.open(self.lockfile, os.O_CREAT | os.O_RDWR, 0o644)
```
abadger, line 54: I think there's a bug here.  `os.O_RDWR` will truncate the file similar to how `open(self.lockfile, 'w')` works. 

A fix might be to use `os.O_APPEND` as well but I would have to test whether that works and does not break anything else.

abadger, line 54: EDIT: Sorry, I reversed my filename and the content of my file when I tested whether these flags would truncate the file on open or not.  This set of flags is fine for this purpose (it does **not** truncate the contents of the file)

abadger:
I've looked over the PR now and have a few higher level questions... In the past, I've implemented locks using filesystem atomicity (the key filesystem operation here is `link()` which will fail if a file already exists).  That is robust and relatively (compared to fcntl and flock) intuitive but has one problem: there is an unavoidable race condition dealing with stale lockfiles.  I have usually forced the user to clean up stale lockfiles for this reason.

I see in this PR that although you are creating a lockfile and adding the PID to it (things which are needed for a filesystem-driven lockfile approach), you're using `fcnt()l` (one of the POSIX APIs) for locking.  However, you aren't using that to determine whether another instance of LEAPP is already running directly, you're just using that to protect the lockfile while determining whether the lockfile is currently allocated by another process, to record our ownership of it by writing our PID to it (these two needs are what `link()` addresses in the filesystem atomicity case), and to ensure that no one else takes the lock while we're evaluating if the lockfile is stale.

What I'm wondering is what problems we are trying to avoid by doing that instead of relying on the lock for the length of the process?  The traditional problem with `fcntl()` is that the lock is associated with the process rather than the specific file descriptor.  So you can't synchronize resources across threads of a single process (the lock is process-wide) and when the process closes any open filedescriptor for that inode, the lock is closed (even if it is a different filedescriptor than the lock was opened against).  Since we're only using the lock at the toplevel of the LEAPP cli, I don't think either of these apply?  We can simply open the filedescriptor in the context manager, successfully take the lock on it, then hold onto that file descriptor until the context manager exits.  That should eliminate the problem of having a stale lockfile as the contents of the file won't matter any longer.

Let me know if I'm missing something.

[*] The other problem with `fcntl()` locking is that it may or may not work on NFS depending on how all the nfs clients are configured (`flock()` is even worse in this regard).  My kind of think this is not a problem, though.  Is the lockfile ever going to be present on an NFS share or will that cause other problems to the upgrade?

leapp/utils/lock.py
abadger, line 54:
```suggestion
        fd = os.open(self.lockfile, os.O_CREAT | os.O_RDWR, 0o600)
```

Security thing I just learned: This needs to be `0o600`.  The way `fcntl()` and `flock()` work, a read lock can be taken out on a file if the user can read the file.  A read lock will block a write lock.  So, if the file is readable by other users, the other users could create a denial of service against LEAPP by asking for a read lock on the file.

leapp/config.py
dkubek: Yeah I definitely agree. This solves a lot of issues. Thanks!!

dkubek:
@abadger Thanks for awesome comment!

_DISCLAIMER: I am very new to this topic and it might soon become obvious_

> I see in this PR that although you are creating a lockfile and adding the PID to it (things which are needed for a filesystem-driven lockfile approach), you're using `fcnt()l` (one of the POSIX APIs) for locking. However, you aren't using that to determine whether another instance of LEAPP is already running directly, you're just using that to protect the lockfile while determining whether the lockfile is currently allocated by another process, to record our ownership of it by writing our PID to it (these two needs are what `link()` addresses in the filesystem atomicity case), and to ensure that no one else takes the lock while we're evaluating if the lockfile is stale.

My original idea I was aiming for was exactly something like you are describing using the ``link()`` call. Basically, just create a file, write PID inside and check if it exists. Instead of just using a plain python ``open()``, I have tried to introduce atomicity. A quick google search suggested using filesystem locks and then I have basically emulated the approach of using ``link()`` (which I was not aware of) using ``flock()``. Which resulted in the current implementation.

So correct me if I'm wrong. What you are suggesting to be a better approach is to acquire a filesystem lock, keep it for the whole duration of the process and release it when leapp ends. Am I correct? This would solve all the problems with stale lockfiles

abadger:
Correct. Avoiding the stale lockfile problem is what we'd be able to
solve.  I think that the known problems with `fcntl()` and `flock()` won't
be triggered by the way we need to use the lock in LEAPP so we can use them
as the primary locking mechanism (correct me if I'm wrong about that,
though).

dkubek:
Thanks everybody for awesome feedback! This last changes should fix all of the problems noted above.

Firstly, as per @abadger 's awesome suggestion, we will not rely on the PID in the lockfile, but instead on the BSD lock itself which we will keep for the duration of the execution. The lockfile has been moved to ``/var/run/leapp.pid`` (thanks @vinzenz !) and will store the PID purely for informational purposes (so we can inform user which process is blocking). This solves the issues around stale lockfiles. The issue raised by @fernflower should also be obsoleted by this.

leapp/utils/lock.py
```diff
@@ -0,0 +1,96 @@
...
+    except OSError:
+        return False
+
+
+def _read_pid(fd):
+    return os.read(fd, 20)
+
+
+def _write_pid(fd, pid):
+    os.write(fd, str(pid).encode('utf-8'))
```
abadger, line 26: `_write_pid()` will corrupt the file if there is already data in it and `_clear_lock()` isn't called first so we should enforce doing the right thing by calling `_clear_lock()` here,  inside of `_write_pid()`.  
```suggestion
    _clear_lock(fd)
    os.write(fd, str(pid).encode('utf-8'))
```

leapp/utils/lock.py
```diff
@@ -0,0 +1,96 @@
...
+            process_msg = ''
+            if running_pid:
+                process_msg = ' by process with PID {}'.format(running_pid)
+
+            msg = ('Leapp is currently locked{} and cannot be started.\n'
+                   'Please ensure no other instance of leapp is running and try again.').format(process_msg)
+            raise ProcessLockError(msg)
+
+        try:
+            _clear_lock(self.fd)
```
abadger, line 78: To go along with the suggestion to move the call to `_clear_lock()` to `_write_pid()`:
```suggestion
```

leapp/utils/lock.py
```diff
@@ -0,0 +1,96 @@
...
+
+        if len(running_pid) == 0:
+            # No pid in lockfile
+            return
+
+        running_pid = _validate_pid(running_pid)
+        if not running_pid:
+            return
+
+        return running_pid
```
abadger, line 63:
```suggestion
    def _get_pid_from_lockfile(self):
        running_pid = _read_pid(self.fd)
        self.log.debug("_get_pid_from_lockfile: running_pid=%s", running_pid)
        running_pid = int(running_pid)
        
        return running_pid
```

It feels like the error checking and exception handling should be done by the caller of this function.  The reason is that `_get_pid_from_lockfile()` doesn't know what to do if the lockfile does not contain a pid.  Is it an error?  Should a warning be emitted?  The caller knows these things but not `_get_pid_from_lockfile()` so the errors should raise an exception that the caller can handle.

In this case, `int("")` and `int("not a pid")` will raise a `ValueError` so we can just let that propogate up.

leapp/utils/lock.py
```diff
@@ -0,0 +1,96 @@
...
+    def _get_pid_from_lockfile(self):
+        running_pid = _read_pid(self.fd)
+        self.log.debug("_get_pid_from_lockfile: running_pid=%s", running_pid)
+
+        if len(running_pid) == 0:
+            # No pid in lockfile
+            return
+
+        running_pid = _validate_pid(running_pid)
+        if not running_pid:
```
abadger, line 60: Note: a pid could be zero so technically this should have been checking if `running_pid is None`.  (In practice, pid 0 is reserved for the kernel so this wouldn't generate a noticable bug.  But it is good to keep in mind for other times you are writing conditions with variables that are numbers.)

leapp/utils/lock.py
```diff
@@ -0,0 +1,96 @@
...
+
+        return running_pid
+
+    def _try_lock(self, pid):
+        if not _acquire_lock(self.fd):
+            running_pid = self._get_pid_from_lockfile()
+
+            process_msg = ''
+            if running_pid:
+                process_msg = ' by process with PID {}'.format(running_pid)
```
abadger, line 71:
```suggestion
            try:
                running_pid = self._get_pid_from_lockfile()
            except ValueError:
                process_msg = ''
            else:
                process_msg = ' by process with PID {}'.format(running_pid)
```
If you accept the change to have`_get_pid_from_lockfile()` propogate the errors up to the caller, then this is where we catch the error and turn it into the proper message string.

leapp/utils/lock.py
```diff
@@ -0,0 +1,96 @@
...
+
+            msg = ('Leapp is currently locked{} and cannot be started.\n'
+                   'Please ensure no other instance of leapp is running and try again.').format(process_msg)
+            raise ProcessLockError(msg)
+
+        try:
+            _clear_lock(self.fd)
+            _write_pid(self.fd, pid)
+        except OSError:
+            raise ProcessLockError('Could not write PID to lockfile.')
```
abadger, line 81: Do we want this to stop execution or to log and continue?  Now that the PID in the lockfile is only informational, I could see it either way.

leapp/utils/lock.py
```diff
@@ -0,0 +1,96 @@
...
+    def _try_lock(self, pid):
+        if not _acquire_lock(self.fd):
+            running_pid = self._get_pid_from_lockfile()
+
+            process_msg = ''
+            if running_pid:
+                process_msg = ' by process with PID {}'.format(running_pid)
+
+            msg = ('Leapp is currently locked{} and cannot be started.\n'
+                   'Please ensure no other instance of leapp is running and try again.').format(process_msg)
```
abadger, line 74:
```suggestion
                   'Please ensure no other instance of leapp is running and then delete the lockfile'
                   ' at {} and try again.').format(process_msg, self.lockfile)
```

abadger (commented):
Code reviewed.  I didn't find anything major, just some adjustments that I thing would make the code cleaner.

leapp/exceptions.py
```diff
@@ -148,3 +148,7 @@ class RequestStopAfterPhase(LeappError):
 
     def __init__(self):
         super(RequestStopAfterPhase, self).__init__('Stop after phase has been requested.')
+
+
+class ProcessLockError(LeappError):
```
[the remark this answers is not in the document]
pirat89: Good idea. I think we should be creating docstrings more to have well documented code for readers outside of the team

leapp/utils/lock.py
pirat89: Hmm. Good question. thinking about it, in case it's hyphotethical situation, most likely doesn't matter whether we raise it or whether we continue. But if this happens, it's question why it should happen and whether it could have another impact on the execution - not just speaking about having or not the content in the file, but speaking about the root cause, whether the same issue could have another impact on the rest of the leapp execution that could lead to another set of errors later.

leapp/utils/lock.py
```diff
@@ -0,0 +1,96 @@
...
+    os.write(fd, str(pid).encode('utf-8'))
+
+
+def _validate_pid(old_pid):
+    try:
+        old_pid = int(old_pid)
+    except ValueError:
+        return False
+
+    return old_pid
```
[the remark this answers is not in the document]
dkubek: This function existed purely for readability purposes. But you are correct that the API creates more confusion then necessary and being such a simple wrapper there is no harm in removing it. Thanks! Fixed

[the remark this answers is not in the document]
dkubek: And I appreciate the insight on bool values!!

leapp/utils/lock.py
dkubek: Very true! Now that I removed the validation function I actually can do away with the whole check altogether. Thanks! Fixed

leapp/exceptions.py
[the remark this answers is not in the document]
dkubek: Definitely! Changed

abadger (approved):
I think all the issues have been resolved so I'm approving this.  There is the open question of whether an incorrectly formatted lockfile should warn or error (currently errors) but I don't feel strongly one way or the other and it seems like Petr can see both sides of it as well.  I'm okay to leave that as it is and revisit in the future if it is a problem.

PR #853, opened 2024-03-01
Update packit with currently supported upgrade paths

fernflower:
That' an adoptation of leapp-repository's packit update and refactoring introduced in PR1176.

OAMG-10855


pirat89 (approved):
lgtm

PR #857, opened 2024-04-25
Update packit config after tier redefinition

fernflower:
Now for basic sanity test verification in upstream tests tagged by 'tier0' will be used instead of 'sanity'.

RHELMISC-3211


fernflower:
Please don't merge until QE give their blessings.

PR #870, opened 2024-07-02
Extend LEAPP with actor configuration

abadger:
Provide a new unified solution to enable User-Configurable actors.

The point is to introduce mechanism how various parts of workflow could be configured, on level actors which should be able to easily access configuration related to them (e.g. via a framework API if designed in such a way). So users will not have to create their own custom actors for all possibilities we provide right now, but they could simply create/update configuration files to achieve what they want.

Issues:
- [x] Reading configs on RHEL7 (Python 2.7) fails, crashing the entire framework because `glob.glob` has no kw argument `recursive=True`

JIRA: [OAMG-8803](https://issues.redhat.com/browse/OAMG-8803)


dkubek:
Adding quick note here so we won't forget. We will want to include the config information also in ``leapp.db``. If we make the ``Config`` serializable we could simply edit ``store_actor_metadata`` in ``leapp/utils/audit/__init__.py`` and it should work well enough.

leapp/actors/__init__.py
```diff
@@ -470,6 +482,8 @@ def get_actor_metadata(actor):
         _get_attribute(actor, 'dialogs', _is_dialog_tuple, required=False, default_value=()),
         _get_attribute(actor, 'description', _is_type(string_types), required=False,
                        default_value=actor.__doc__ or 'There has been no description provided for this actor.'),
+        _get_attribute(actor, 'config_schemas', _is_type(Config), required=False,
+                       default_value=actor.__doc__ or 'Description of the configuration used by this actor.'),
```
dkubek, line 486: We will want here something similar like ``_is_config_tuple`` similar to the functions above. That is if I understand it correctly and we are allowing multiple config schema definitions per actor. Am I right @abadger ?

leapp/actors/__init__.py
```diff
@@ -86,6 +92,7 @@ def serialize(self):
             'path': os.path.dirname(sys.modules[type(self).__module__].__file__),
             'class_name': type(self).__name__,
             'description': self.description or type(self).__doc__,
+            'config_schemas': [c.__name__ for c in self.config_schemas],
```
dkubek, line 95: I am thinking here, whether we only add the name of the config or a whole serialized representation as we do with dialogs. In my head ``c.serialize()`` makes more sense, however it is true that the schema is an entity that can be shared across actors opposed to dialog which is always specific to the given actor, so doing it this way has merit also. What do you think? @abadger @pirat89

dkubek: If we keep it like this, this also brings to my attention the issue where we store config information in leappdb. Is config part of actor metadata? I argue that the config scheme (the name) is actor metadata, however the schema itself is not. Maybe we should create another table and store the schema information there instead of in metadata.

leapp/repository/__init__.py
```diff
@@ -189,6 +190,7 @@ def mapped_actor_data(data):
             })
             return data
 
+        # TODO: Do we need to add configs here as well?
```
dkubek, line 193: This is a good question, I think we need to decide and have a discussion on the position of configs in the "hierarchy" of leapp (how tightly is a config definition coupled to given actor, what is its scope...). Currently I am thinking about Configs having similar status as a Model. Meaning it is an entity that has the whole repository as a scope and is not bound to a specific actor. From that point of view I think the information would fit here.

Could you maybe explain what is your view and thought process?

leapp/actors/__init__.py
abadger: Nice catch!  I'ce implemented such a function.  The one change I made from how _is_model_tuple() and etc operate is to check for a Sequence instead of a Tuple.  This allows Tuples, Lists, NamedTuples, to all satisfy the requirement rather than only tuples.  If that shouldn't be the case, we can change it back.

leapp/actors/__init__.py
abadger: I'm not sure whether name only or a serialization of the schema class is appropriate here.  @pirat89 ?

I'm not sure what leappdb is used for either:  Would we want to store both the config_schema in it and the actual actor config settings?  Would we want to store just the config settings from config files or would we want to store the config dictionary that has been expanded to include default values as well?

leapp/repository/__init__.py
abadger: I may be dragging in an unrelated discussion: I remember that we were figuring out how we could make sure that changes to a config schema that was shared amongst several different Actors would be caught and error before an upgrade started.  IIRC, the strategy we decided upon was for the config schema to be copied into every Actor that wanted to use it.  Then the framework could detect if these opies got out of sync and error out (forcing the user to fix the code before continuing).

Would that mean we do not want configs added here?

leapp/actors/__init__.py
dkubek: The reason we would like to have this information in the database is for debugging and troubleshooting purposes (using for example leapp-inspector). My idea is that we would like to have a snapshot of the leapp configuration just as the user sees it. This means that in case something goes wrong for the user we can see what is the state of leapp when he provides the leapp.db file.

So I personally would store both the entire schema, so we can see the setup of the configuration as well as the actual values provided by the user. The format in which this is stored does not matters from my point of view only in terms of how easily we can work with it later when this information is retrieved from processing by other tools (such as the leapp-inspector I mentioned)

leapp/repository/__init__.py
dkubek: I'll be honest I am a bit lost about what we agreed on at this point :joy: However, in such a case I agree that it should _not_ be added here and the information should be present only in conjunction with actors just as dialogs are

abadger:
Hey @vinzenz , I don't know if you have time but I was wondering if you recall where in the code I need to add something to recognize directories inside of the actor as being a plugin?  I'm trying to add a `configs/` directory that is pretty much the same as `libraries/`.  I found the places in the code I need to modify to make a configs directory at the repository level work but I'm stll not recognizing a configs directory inside of each individual actor.  I know I'm probably just missing a few line of code somewhere that sets it up but I can't seem to find it.

leapp/repository/actor_definition.py
```diff
@@ -175,6 +175,7 @@ def serialize(self):
             'class_name': self.class_name,
             'description': self.description,
             'tags': self.tags,
+            'config_schemas': self.config_schemas,
```
vinzenz, line 178: isn't  ```'configs': self.configs,``` missing here?

abadger: I thought that serialize would need to be used for things that are class attributes rather than the directories but I'll give that a try.

vinzenz: I was just going through the PR and tried to see if I can find what's missing and just stumbled upon this.
From what I can tell just by looking at the PR, it does seem like you did it right. I will get a snapshot of your PR and try it and have a look if I can find the issue.

Btw this is feature that asks for adding snactor support.

pirat89: > Btw this is a feature, that asks for adding snactor support.

You are right, however we decided to have updated snactor as a stretch goad for the next (spring) release and possibly finish it in the other one.

leapp/configs/actor/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""
+:py:mod:`leapp.configs.actor` represents the import location for private actor config schema that
+are placed in the actor's configs folder.
+
+
+Example:
+      If your actor has a configs folder with a schemas.py python module, import it
+      from the actor like this::
+
+      from leapp.libraries.actor import schemas
```
matejmatuska, line 10: I believe it should be `leapp.libraries.actor`:
```suggestion
      from leapp.configs.actor import schemas
```

matejmatuska: Sorry I know this is still a draft just noticed it while checking out the code

MichalHe (changes requested):
I went through the code. It looks OK, but I cannot claim I am sure since the codebase is a bit hidden in "fog of war". @dkubek I will take care of the my own review suggestions. Can you comment on the comment where I tagged you explicitly?

leapp/actors/__init__.py
```diff
@@ -390,17 +421,23 @@ def _lint_warn(actor, name, type_name):
...
-    if isinstance(value, type) and issubclass(value, Model):
-        _lint_warn(actor, name, "Models")
+def _is_foo_sequence(cls, cls_name, actor, name, value):
+    if isinstance(value, type) and issubclass(value, cls):
+        _lint_warn(actor, name, cls_name)
         value = (value,)
-    _is_type(tuple)(actor, name, value)
-    if not all([True] + [isinstance(item, type) and issubclass(item, Model) for item in value]):
+    _is_type(Sequence)(actor, name, value)
+    if not all([True] + [isinstance(item, type) and issubclass(item, cls) for item in value]):
```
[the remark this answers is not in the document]
MichalHe: I have changed this part to
```
all(isinstance(item, type) and issubclass(item, cls) for item in value)
```
removing the needless construction of list.

leapp/actors/config.py
```diff
@@ -0,0 +1,349 @@
...
+
+try:
+    # Compiled versions if available, for speed
+    from yaml import CSafeLoader as SafeLoader, CSafeDumper as SafeDumper
+except ImportError:
+    from yaml import SafeLoader, SafeDumper
+
+
+_ACTOR_CONFIG = None
+_ACTOR_CONFIG_VALIDATED = False
```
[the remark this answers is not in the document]
MichalHe: Removed.

leapp/actors/config.py
```diff
@@ -0,0 +1,349 @@
...
+    except ModelViolationError as e:  # pylint: disable=broad-exception-caught,broad-except
+        # Any problems mean that the field did not validate.
+        log.info("Configuration value failed to validate with: %s", e)
+        return False
+    return True
+
+
+def _normalize_config(actor_config, schema):
+    # Validate that the config values read from the config files obey the known
+    # structure.
```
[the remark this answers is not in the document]
MichalHe: Suggestion applied.

leapp/actors/config.py
```diff
@@ -0,0 +1,349 @@
...
+            to_keep:
+                - leapp
+                - python2-leapp
+                - python3-leapp
+                - leapp-repository
+                - snactor
+    """
+    # TODO: This is just a placeholder.  We need to do some additional
+    # formatting that includes the documentation, not just return it as is.
+    return yaml.dump(_ACTOR_CONFIG, SafeDumper)
```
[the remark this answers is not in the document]
MichalHe: Suggestion applied.

leapp/configs/common/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
+:py:mod:`leapp.configs.common` represents an import location for shared libraries that
+are placed in the repository's configs folder.
+"""
+
+LEAPP_BUILTIN_COMMON_CONFIGS_INITIALIZED = False
+""" Framework internal - were common configs initialized? """
```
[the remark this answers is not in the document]
MichalHe: Suggestion applied.

leapp/models/fields/__init__.py
```diff
@@ -457,6 +465,70 @@ def serialize(self):
...
+    Map from strings to instances of a given value type.
+    """
+    def __init__(self, value_type, **kwargs):
+        super(StringMap, self).__init__(**kwargs)
+
+        if self._default is not None:
+            self._default = copy.copy(self._default)
+
+        if not isinstance(value_type, Field):
+            raise ModelMisuseError("elem_field must be an instance of a type derived from Field")
```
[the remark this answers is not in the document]
MichalHe: Suggestion applied.

leapp/actors/__init__.py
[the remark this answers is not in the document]
MichalHe: @dkubek could you, please, double-check this whether it looks OK and if so close this conversation?

leapp/repository/actor_definition.py
```diff
@@ -212,7 +214,10 @@ def discover(self):
             if p.exitcode != 0:
                 self.log.error("Process inspecting actor in %s failed with %d", self.directory, p.exitcode)
                 raise ActorInspectionFailedError('Inspection of actor in {path} failed'.format(path=self.directory))
-            result = q.get()
+
+            with self.injected_context():
+                result = q.get()
```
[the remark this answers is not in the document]
dkubek: Very good idea. Addded! Can you check it if it makes sense, please?

leapp/utils/audit/__init__.py
```diff
@@ -651,10 +651,15 @@ def store_actor_metadata(actor_definition, phase):
         'consumes': sorted(model.__name__ for model in _metadata.get('consumes', ())),
         'produces': sorted(model.__name__ for model in _metadata.get('produces', ())),
         'tags': sorted(tag.__name__ for tag in _metadata.get('tags', ())),
+        'config_schemas': [field.serialize() for field in _metadata.get('config_schemas', ())],
     })
     _metadata['phase'] = phase
 
-    actor_metadata_fields = ('class_name', 'name', 'description', 'phase', 'tags', 'consumes', 'produces', 'path')
+    # TODO(dkubek): Is configuration metadata? Does it make sense to include it here?
+    _metadata['config'] = config
```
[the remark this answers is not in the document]
dkubek: I have removed it from the metadata. I have decided to store it in a separate table, so that it is linked only to 'execution'. I was considering multiple possibilities, but in the end this was one that made the most sense to me.

leapp/actors/__init__.py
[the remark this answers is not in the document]
dkubek: Yes, I agree. It is superfluous. Nice catch!!!

leapp/repository/actor_definition.py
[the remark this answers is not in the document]
MichalHe: Nice! Thank you. Closing this conversation.

packaging/leapp.spec
```diff
@@ -13,11 +13,11 @@
 # This is kind of help for more flexible development of leapp repository,
 # so people do not have to wait for new official release of leapp to ensure
 # it is installed/used the compatible one.
-%global framework_version 5.0
+%global framework_version 6.0
```
pirat89, line 16: :100:

packaging/leapp.spec
```diff
@@ -13,11 +13,11 @@
...
 
 # IMPORTANT: everytime the requirements are changed, increment number by one
 # - same for Provides in deps subpackage
-%global framework_dependencies 5
+%global framework_dependencies 6
```
pirat89, line 20: :100:

leapp/utils/audit/__init__.py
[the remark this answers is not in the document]
MichalHe: Good job! Closing the conversation to keep track of what still needs work/discussion.

pirat89 (commented):
In general I think it's ok. I haven't found any serious problems in the code that would not be already covered and some stuff that I think we should still deliver is covered by TODOs which can definitely wait for a follow up PR. However, when I've been testing configs with the RHUI configuration in leapp-repository, I found several things that seems suspicious to me, but it's possibly unrelated to the framework and it's problem in the leapp-repository PR. I will continue in deeper investigation tomorrow with fresh mind, but I am considering it actually already good enough.

Before the merge:
* polish commits (squash lint fixes, etc.)
* fix the docstring "misinformation" noted by @matejmatuska

leapp/configs/actor/__init__.py
MichalHe: Thanks, resolved.

pirat89:
ignore failed tests on RHEL 7.9. there are some hicupps due to yum and some special requirement to have a build from leapp-repository PR. it's ok

MichalHe (approved):

pirat89:
manually tested on AWS for IPU 7 -> 8 and 8 -> 9 as well. work as expected.

PR #871, opened 2024-07-03
Packit: drop epel-7-ppc64le chroot

pirat89:
CentOS 7 is EOM and based on that additional changes have to be made in COPR. Use x86_64 arch for epel7 now only.


fernflower (approved):
lgtm!

PR #875, opened 2024-07-16
Packit: Bring all changes from leapp-repository + update GH actions

pirat89:
- replace the sanity tag by 'tier0' (RHELMISC-3211)
- Update ENV variables for RHUI tests
- Enable upgrade path to 8.10 on AWS
- adjust post_install_script for RHEL 7.9 (moved to TF)
- Use frozen branch for rhel7 downstream testing as RHEL 7 is EOM


pirat89:
it seems as infra issue, as majority tests have passed before the rerun without any issue - but 2.

PR #876, opened 2024-07-17
Add .readthedocs.yaml fixing builds on RTD

matejmatuska:
The builds in Read the Docs have been broken for a few months now, because the .readthedocs.yaml is required.


pirat89 (approved):
lgtm. merging to see how it's going

PR #877, opened 2024-07-17
docs: Pin sphinx-rtd-theme version fixing build error

matejmatuska:
Previously the 0.4.3 version was correctly used. There are still some warnings during build, but those were also present previously (verified by checking old RTD build).

Follows up on #876 - fixes required to make build on RTD working again.


pirat89 (approved):
lgtm again :D

PR #880, opened 2024-08-12
Fix leaking FDs in the run (_call) function 

t184256:
updated by @pirat89 

Original implementation of the `_call` function (used by `run` from actors) has been leaking file descriptors per each invocation. So in case an actor executed too many of them during an execution it could reach the limit of allowed opened FDs and the execution failed with `OSError: Too many opened files` error.

There have been actually 2 issues:
* `EventLoop` (alias `select.epoll`) have been called before the fork so the child process inherit the created FD too
  * that has not been so big problem actually as the child process always ended before the parent finished the function so this has been closed always safely by OS anyway. But updating it anyway to make it cleaner.
* `stdout` and `stderr` pipes (to consumes content from child process) have been created before the fork but have not been closed in the parent
  * The parent process hasn't closed related FDs so this led to the situation where 2 FDs leaked per the code execution as these have been opened until the end of the actor execution. Closing these after the `EventLoop` is closed (and death of the child process) resolved this issue.

Warning: totally untested (@pirat89 manually tested on rhel7 & rhel9)

### DIFF
--- a/leapp/libraries/stdlib/call.py
+++ b/leapp/libraries/stdlib/call.py
@@ -202,6 +202,8 @@ def _call(command, callback_raw=lambda fd, value: None, callback_linebuffered=la
         # Wait for the child to finish
         pid, status = os.wait()
         ep.close()
+        os.close(stdout)
+        os.close(stderr)
 
         # The status variable is a 16 bit value, where the lower octet describes
         # the signal which killed the process, and the upper octet is the exit code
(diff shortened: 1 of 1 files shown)


pirat89:
tested and it's still broken with the same error for IPU 9 -> 10. Let's see whether we will be able to fix it before the release, otherwise let's keep it for the next release. Currently we have just ~4d for changes in this release and the issue is discovered only for IPU 9 -> 10.

pirat89 (approved):
seems it works as expected now. tested manually (rhel7 & rhel9), the stdout/stderr FDs have been really  leaking too. Let's wait for the test results yet.

pirat89 (approved):
lgtm and works as expected. double checked with @abadger that it should be safe enough even so close to the release. merging

PR #889, opened 2025-02-12
ci(fix): change e2e source and target to supported versions

MichalHe:
The current source version used in e2e tests is 8.6 and the target version is 9.0, which are unsupported. This patch bumps the source/target versions used in the CI to source=8.10 and target=9.4. There is also a job with target=9.6, which is expected to be failing until 9.6 is released.


tomasfratrik (commented):
lgtm, can be squashed

pirat89 (approved):
lgtm.

PR #893, opened 2025-04-03
Handle case when db directory is missing

karolinku:
When directory is missing, this change will introduce creation of missing path and allow proceeding with execution.

JIRA: RHEL-54382


pirat89 (approved):
lgtm!!

PR #894, opened 2025-04-06
Fix CLI: allow to set falsy values as default

pirat89:
When specifying an option for a leapp CLI command using @add_option decorator, it has not been possible to specify a "falsy" value (or let's say empty) as `default` e.g. empty string, empty list, etc. Start to check NoneType for the `default` parameter properly to allow it to set this.

**Bump leapp-framework to 6.1.**

Note:
The similar problem is for `choices` and `metavar` parameters. In case of `metavar` it's however ok to ignore empty string. In case of empty `choices` it would lead to the same evaluation as in case of NoneType (any input is ok). So I that the `default` is the only parameter that needs to be checked properly.

jira: RHEL-86206

### DIFF
--- a/leapp/utils/clicmd.py
+++ b/leapp/utils/clicmd.py
@@ -228,8 +228,8 @@ class Command(object):
         :type metavar: str
         :param choices: range of values that the argument is allowed to take
         :type choices: list
-        :param choices: default value of the argument if nothing is specified
-        :type choices: str
+        :param default: default value of the argument if nothing is specified
+        :type default: any
         :return: self
         """
         name = name.lstrip('-')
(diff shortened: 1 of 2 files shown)


pirat89:
@oamg/developers  I consider it kind of done. I want to however discuss it first to deal with some internals.

MichalHe (approved):
I have tested this PR by adding the following lines into `commands/upgrade/__init__.py`:
```python3
...
@command_opt('myflag', default=[])
def upgrade(args, breadcrumbs):
     print(args.myflag)
     import pdb; pdb.set_trace()
```

Without having the changes introduced in this PR, I observe `args.myflag` having the value `None`. With this PR, I have correct value `[]`. LGTM

MichalHe:
The failing aws test is not related to the changes introduced by this PR, hence merging.

PR #897, opened 2025-04-23
Snactor: add command to show dependencies between actors

MichalHe:
Adds a new command `show-dependencies <workflow>` that displays all producer/consumer dependencies in the workflow. The output of the command is in the [DOT](https://graphviz.org/doc/info/lang.html) language that allows easy visualization of graphs.

Example use:
```
# snactor show-dependencies WORKFLOW > deps.dot
# xdot deps.dot
```

### DIFF
--- a/leapp/snactor/commands/show_dependencies.py
+++ b/leapp/snactor/commands/show_dependencies.py
@@ -42,7 +42,7 @@ def cli(params):
     if params.only_influencing_actor:
         actor_of_interest = repository.lookup_actor(params.only_influencing_actor)
         if not actor_of_interest:
-            raise CommandError('Could not find any workflow named "{}"'.format(params.name))
+            raise CommandError('Could not find given actor named "{}"'.format(params.only_influencing_actor))
 
     workflow_instance = workflow()
 
--- a/leapp/snactor/commands/show_dependencies.py
+++ b/leapp/snactor/commands/show_dependencies.py
@@ -36,7 +36,7 @@ def cli(params):
 
     workflow = repository.lookup_workflow(params.workflow_name)
     if not workflow:
-        raise CommandError('Could not find any workflow named "{}"'.format(params.name))
+        raise CommandError('Could not find any workflow named "{}"'.format(params.workflow_name))
 
     actor_of_interest = None
     if params.only_influencing_actor:
--- a/leapp/snactor/commands/show_dependencies.py
+++ b/leapp/snactor/commands/show_dependencies.py
@@ -1,0 +1,64 @@
+import itertools
+from collections import defaultdict
+import sys
+
+from leapp.exceptions import CommandError, LeappError
+from leapp.logger import configure_logger
+from leapp.repository.scan import find_and_scan_repositories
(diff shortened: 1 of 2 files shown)


leapp/snactor/commands/show_dependencies.py
```diff
@@ -0,0 +1,64 @@
...
+    try:
+        repository.load()
+    except LeappError as exc:
+        sys.stderr.write(exc.message)
+        sys.stderr.write('\n')
+        sys.exit(1)
+
+    workflow = repository.lookup_workflow(params.workflow_name)
+    if not workflow:
+        raise CommandError('Could not find any workflow named "{}"'.format(params.name))
```
pirat89, line 36:
```
# snactor show-dependencies IPUConfig
  File "/usr/bin/snactor", line 11, in <module>
    load_entry_point('leapp==0.18.0', 'console_scripts', 'snactor')()
  File "/usr/lib/python3.6/site-packages/leapp/snactor/__init__.py", line 79, in main
  File "/usr/lib/python3.6/site-packages/leapp/utils/clicmd.py", line 111, in execute
    args.func(args)
  File "/usr/lib/python3.6/site-packages/leapp/utils/clicmd.py", line 133, in called
    self.target(args)
  File "/usr/lib/python3.6/site-packages/leapp/utils/repository.py", line 20, in checker
    return f(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/leapp/snactor/context.py", line 34, in wrapper
    return f(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/leapp/snactor/commands/show_dependencies.py", line 36, in cli
    raise CommandError('Could not find any workflow named "{}"'.format(params.name))
AttributeError: 'Namespace' object has no attribute 'name'
```

pirat89:
@MichalHe thinking about the current implementation, it seems to me that it would fit more as sub-command for the `workflow` command, like:

```
$ snactor workflow show-dependencies
```

as snactor (and leapp) can works also without workflows at all, but currently it's connected strictly to workflow. If we want to have it as the main command, I would expect maybe something like this:

```
snactor show-dependencies  # nothing else - print dependencies for all actors discovered in the leapp repository
snactor show-dependencies --all # cover also actors in referenced repositories
snactor show-dependencies --workflow <WORKFLOW>  # like --all but filter everything based on the specified workflow
```

wdyt?

leapp/snactor/commands/show_dependencies.py
MichalHe: Thank you, should be fixed (untested, will test in a moment).

MichalHe:
@pirat89 You are right with having a subcommand of `workflow` would be better.

What I also wish to do is to collect actors from multiple repositories, not just the repository defined at `.`, because this feature is a bit useless if the actor I am developing lives in `el8to9`, but all of its dependencies are in `common`. Perhaps, is there a part of snactor's codebase that I could use as an example?

pirat89:
@MichalHe can you rebase it? it should be working again now. and we could finally finish it to the merge :)

PR #878, opened 2024-07-30
Initial documentation refactor

matejmatuska:
The documentation is reorganized to adhere to the new hierarchy design and also drops the documentation related to oamg/leapp-repository and related direct references. 

Note that some pages/sections might feel a little out of place or miss some content (most notably building-blocks-and-architecture and libraries-and-api), that's expected and these places will be filled in in future patches.

A few typo corrections are also included.

Depends on #854

PR #902, opened 2025-07-10
docs: Add reporting tutorial

matejmatuska:
This was never properly covered by docs on the framework level and has been completely missing since the removal of oamg/leapp-repository related docs removal in #878.


pirat89:
@matejmatuska  when you have free cycles, we could finish that one :)

PR #905, opened 2025-08-04
Drop support for el7toel8

tomasfratrik:
Removing support for el7toel8 since since the upgrade path is no longer supported.

Changes:
* Makefile adjusted
* Reflect changes in documentation
* Remove rhel7 build and test containers

Jira: RHELMISC-11005


pirat89 (approved):
lgtm! good catch in the documentation!

PR #921, opened 2026-04-13
fix(reporting) use single quoting when converting remediation commands

PeterMocary:
The remediation commands are provided as a list of arguments that is used in leapp-report.json. The same command is also converted to a string that is used in leapp-report.txt and can be copy-pasted by the user into shell.

Originally, the command conversion tried to make the resulting commands prettier by enabling double quoting whenever there was a single quote in an argument, while still trying to make the double quoted argument a literal. This proved to be problematic approach due to multiple cases that resulted with incorrect commands. Namely, when the command called `bash -c` with a subcommand that would be evaluated by another non-interactive shell, or when there was an exclamation mark that would have different meaning in non-interactive and interactive shell and its escaping was not possible unless single-quotes are used.

This patch makes the command conversion produce pretty commands **in the most cases** (none of the current commands are affected as far as i could tell). Moreover, using shlex exclusively makes sure the command is always correct (if the list version is correct to begin with). This approach leaves possibility to encounter unintuitive escaping in the resulting converted command, although this can be prevented by carefully crafting the command when creating the report.

Note that if there is a subcommand called in the converted command, the report creator is responsible for its form and for the best resulting command it should favor the double quoting of its arguments when possible.

Bumps specifle version to 6.5
Jira: RHEL-156521

Related leapp-repository PR: [PR#1520](https://github.com/oamg/leapp-repository/pull/1520)

### DIFF
--- a/leapp/reporting/__init__.py
+++ b/leapp/reporting/__init__.py
@@ -275,32 +275,6 @@ def _guarantee_decoded_str(text, encoding='utf-8'):
         return text
 
 
-def _quote_for_shell(s):
-    """
-    Quote a string for shell usage as a literal.
-
-    This function quotes the given string in single-quotes using the
-    shlex.quote. However, when the single-quote is present in the string it
-    instead uses double-quotes, resulting in cleaner output (e.g. "don't"
-    instead of 'don'"'"'t' that would be produced by shlex.quote).
-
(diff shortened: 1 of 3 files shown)


pirat89 (approved):
discussed and tested on mtg. we agree with the change

PR #925, opened 2026-05-20
Fix exception in normalize_schemas for conflicting actor config schemas

matejmatuska:
The normalize_schemas function incorrectly raises an exception when a config schema (subclass of Config) is contained more than once in it's input, regardless of whether it's a conflicting (same section and name, but a difference in other members) or the same (same section and name).

Traceback example:
```
  File "/usr/bin/leapp", line 33, in <module>
    sys.exit(load_entry_point('leapp==0.21.0', 'console_scripts', 'leapp')())
  File "/usr/lib/python3.9/site-packages/leapp/cli/__init__.py", line 51, in main
  File "/usr/lib/python3.9/site-packages/leapp/utils/clicmd.py", line 111, in execute
    args.func(args)
  File "/usr/lib/python3.9/site-packages/leapp/utils/clicmd.py", line 133, in called
    self.target(args)
  File "/usr/lib/python3.9/site-packages/leapp/cli/commands/upgrade/breadcrumbs.py", line 169, in wrapper
    return f(*args, breadcrumbs=breadcrumbs, **kwargs)
  File "/usr/lib/python3.9/site-packages/leapp/cli/commands/preupgrade/__init__.py", line 86, in preupgrade
    command_utils.load_actor_configs_and_store_it_in_db(context, repositories, cfg)
  File "/usr/lib/python3.9/site-packages/leapp/cli/commands/command_utils.py", line 291, in load_actor_configs_and_store_it_in_db
    actor_config_schemas = actor_config.normalize_schemas(actor_config_schemas)
  File "/usr/lib/python3.9/site-packages/leapp/actors/config.py", line 180, in normalize_schemas
    if unique_name in added_fields and added_fields[unique_name] != field:
TypeError: 'set' object is not subscriptable
```

The added_fields variable is a set. The condition for checking conflicts attempts to do indexing on the set, which is not possible in Python sets and results in a TypeError.

This patch removes the set altogether since it doesn't bring any advantage over just using the normalized_schemas dict.

### DIFF
--- a/leapp/actors/config.py
+++ b/leapp/actors/config.py
@@ -169,31 +169,35 @@ def _get_config(config_dir='/etc/leapp/actor_conf.d'):
 def normalize_schemas(schemas):
     """
     Merge all schemas into a single dictionary and validate them for errors we can detect.
+
+    :param schemas: List of schemas to normalize
+    :type schemas: Iterable[Iterable[Type[Config]]]
+    :return A dictionary with structure: section -> name -> Config
+    :rtype: DefaultDict[str, Dict[str, Type[Config]]]
     """
-    added_fields = set()
     normalized_schema = defaultdict(dict)
     for schema in schemas:
         for field in schema:
-            unique_name = (field.section, field.name)
-
+            # section and name are the unique key for a config
+            existing_field = normalized_schema[field.section].get(field.name, None)
             # Error if the field has been added by another schema
-            if unique_name in added_fields and added_fields[unique_name] != field:
-                # TODO: Also include information on what Actor contains the
-                # conflicting fields but that information isn't passed into
-                # this function right now.
-                message = ('Two actors added incompatible configuration items'
-                           ' with the same name for Section: {section},'
-                           ' Field: {field}'.format(section=field.section,
-                                                    field=field.name))
-                log.error(message)
-                raise SchemaError(message)
+            if existing_field:
+                if existing_field != field:
+                    # TODO: Also include information on what Actor contains the
+                    # conflicting fields but that information isn't passed into
+                    # this function right now.
+                    message = (
+                        'Two actors added incompatible configuration items with'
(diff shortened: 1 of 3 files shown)


leapp/actors/config.py
```diff
@@ -169,31 +169,35 @@ def _get_config(config_dir='/etc/leapp/actor_conf.d'):
...
+    :rtype: DefaultDict[str, Dict[str, Type[Config]]]
     """
-    added_fields = set()
     normalized_schema = defaultdict(dict)
     for schema in schemas:
         for field in schema:
-            unique_name = (field.section, field.name)
-
+            # section and name are the unique key for a config
+            existing = normalized_schema[field.section].get(field.name, None)
```
MichalHe: I don't like the variable name. Existing what?

matejmatuska: I think it's obvious from the initialization, but I will update it.

MichalHe: Yeah, sorry. I am driven by the PTSD gained while investigating context management in conftest.py. Many things there might have been once obvious, but I could not see them :D

matejmatuska: Done

tests/scripts/test_actor_config.py
```diff
@@ -0,0 +1,87 @@
...
+    default = "different than in A"
+
+
+# conflicts in default
+class _TestConfigConflictsInType(Config):
+    section = "test"
+    name = "a_setting"
+    type_ = fields.String()
+    description = 'Description here'
+    default = "different than in A"
```
pirat89: @matejmatuska copy paste error?

matejmatuska: it is, I just found it independently :D

pirat89: fixed

pirat89 (approved):
lgtm now. we discussed the problem and changes via call and all requested changes have been delivered.

PR #926, opened 2026-05-20
Improve Config definition API

matejmatuska:
Found some annoyances when working with the Config API, the commit msgs should have all the info. 

I think it should be essentially backwards compatible with the previous version + the added checks which when violated would either way result in an error elsewhere.

There is also a fix for AttributeError in retrieve_config.

TODO tests

### DIFF
--- a/leapp/actors/__init__.py
+++ b/leapp/actors/__init__.py
@@ -389,7 +389,7 @@ class Actor(object):
         :return: Dictionary containing requested configuration.
         :rtype: dict
         """
-        return retrieve_config(self.config_schema)
+        return retrieve_config(self.config_schemas)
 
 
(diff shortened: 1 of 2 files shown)


matejmatuska:
I improved the commit message for the first commit, hopefully it's easier to follow now, lmk if not.

leapp/actors/__init__.py
MichalHe: Nice catch 👍 
```suggestion
        Retrieve the configuration described by self.config_schemas.
```

MichalHe (changes requested):
I approve, except for the small requested change.

Thank you for improving the things that cause friction instead of just working around the issues. Your work is for me a reminder that I should also step up and improve things that I do not like.

